PR #5, opened 2017-12-03
Update components-and-apis.md

damonbauer:
Links to specific components [on this page](https://facebook.github.io/react-native/docs/components-and-apis.html) are broken because they link to `.md`. Changing the links to `.html` fixes the issue.


hramos:
Thanks for catching this. Links to md are only converted to html when used within markdown links. Plain old html links won't get converted. I'm sure we have a bunch of these left over to fix!

PR #27, opened 2017-12-05
Fix a typo in image docs

haikyuu:
Split two words in image docs


hramos:
Can you apply the fix to the version in `docs/` as well?

haikyuu:
@hramos Done.

PR #28, opened 2017-12-05
Fix Animated.timing() link in Easing

charpeni:
The link to `Animated.timing()` was wrong. 

Also applied to versioned docs.


hramos:
Excellent

Issue #9, opened 2017-12-04
Edit button is linking to versioned_docs instead of docs


liamsheppard:
I was confused when clicking the Edit button that I was being taken into the `versioned_docs/version-X.XX` folder.

I'm assuming this is incorrect, based on this quote from `README.md`:

> Do not edit the auto-generated files within 'versioned_docs/' or 'versioned_sidebars/' unless you are sure it is necessary. Edits made to older versions will not be propagated to newer versions of the docs.

gaearon:
This also confused me. It directed me to the versioned page for a very old version of the docs.

hramos:
For now, you may have noticed I've been merging changes made against `docs/` as well as `versioned_docs/`. These are mostly changes that fix broken links and other issues related to the transition, as we do want those changes to be applied to past versions.

PR #32, opened 2017-12-05
Update edit links

hramos:
This will require an update to our static site generator. Once the newer version of the generator is on npm, we'll bump the version in package.json and land this.

The result will be that Edit this Doc links will always point to the version in 'docs/' instead of 'website/versioned_docs/'.

Fixes https://github.com/facebook/react-native-website/issues/9.


brunolemos:
The links seems to be still pointing to an invalid page, e.g.:

https://github.com/facebook/react-native-website/blob/master/website/versioned_docs/virtualizedlist.md

PR #23, opened 2017-12-05
Better component examples for DatePickerIOS and KeyboardAvoidingView

liamsheppard:
## Motivation
I had trouble when navigating the Docs working out what components were or how they worked. Sometimes the names can be a bit vague, for example, after building `<Picker>` into my project on iOS, the result was unexpected, I was actually wanting `ActionSheetIOS()`.

## Solution
Putting screenshots/gifs into the docs for each component to show how it works and, if applicable, how props effect the component. I am going to split this across a few PRs so that it doesn't become overwhelming or get overly stalled, and also to hopefully set a precedent for others to do the same.

## List of Fixes
- Added example code for `DatePickerIOS`
- Added example images/gifs for `DatePickerIOS` and props to demonstrate their behavior
- Added example gif for `KeyboardAvoidingView`
- Added a caveat to the `KeyboardAvoidingView` that I have noticed while developing personal projects
- Fixed a few broken links

PR #35, opened 2017-12-06
Better component examples for Alert, MaskedViewIOS and SegmentedControlIOS

liamsheppard:
Following on from #23

I'm currently mainly targeting components that have vague names, or I struggled to understand by reading the docs.

Site has been tested locally and confirmed that all images work.

## List of Fixes
- Added working example code for `MaskedViewIOS`
- Added example images for `Alert`, `MaskedViewIOS` and `SegmentedControlIOS`
- Fixed the size of `DatePickerIOS` images from my previous PR


docs/alert.md
```diff
@@ -13,6 +13,23 @@ This is an API that works both on iOS and Android and can show static
 alerts. To show an alert that prompts the user to enter some information,
 see `AlertIOS`; entering text in an alert is common on iOS only.
 
+## Example
+
+<table>
```
liamsheppard: I've used a table here instead of markup as the markup table styles the first column as 60% wide, instead of 50/50

PR #36, opened 2017-12-07
Document `tintColor` option for ActionSheetIOS

gianpaj:
also:
- fixed the `showActionSheetWithOptions` example

This option was added in https://github.com/facebook/react-native/pull/4590

FYI, this is not even documented in the comments
https://github.com/facebook/react-native/blob/master/Libraries/ActionSheetIOS/ActionSheetIOS.js#L28

Are the docs generated from the code? Should I update that instead?


hramos:
The docs were initially generated from code, but going forward they will not be synced. Feel free to add any necessary comments back into the JS file.

The idea here is to have the full API reference documented on the website, while only having minimal comments for each method in the source code. It will take a while to update every source file to match this style but you can see an example of what we're going for here:
https://github.com/facebook/react-native/blob/master/Libraries/AppState/AppState.js

hramos:
Can you make this change to the file in `docs/` as well? Normally we would only update the `docs/` version whenever an api is added/changed, but fixes like these are worth adding both to `docs/` as well as `versioned_docs/`.

gianpaj:
Thanks for fixing the conflicts. I didn't have time and now I know how I should have done it as well :)
👍 to RN

PR #37, opened 2017-12-07
Replace deprecated View.propTypes by ViewPropTypes

charpeni:
`View.propTypes` are deprecated and not available since RN 0.49.

Replaced with `ViewPropTypes`.


hramos:
Thanks for catching this!

PR #41, opened 2017-12-07
Add missing props documentation in FlatList

AudyOdi:
based on issue https://github.com/facebook/react-native/issues/16501 about missing documentation in some RN components, in this PR I added back the props documentation of `FlatList` component. All the documentation and description were taken from https://facebook.github.io/react-native/docs/0.48/flatlist.html with some adjustment to match the current doc style.

Before:
image
image

With this PR:
image
image
image

Issue #40, opened 2017-12-07
Docs doesn't show props descriptions


brunolemos:
<img width="706" alt="image" src="https://user-images.githubusercontent.com/619186/33725040-26a7a77c-db58-11e7-981c-3479a45e3b4a.png">

hramos:
Can you link to the doc that contains these props?

AudyOdi:
@hramos I think this one is from `VirtualizedList` https://facebook.github.io/react-native/docs/virtualizedlist.html

brunolemos:
^ yes, `VirtualizedList`

PR #53, opened 2017-12-10
Add missing doc virtualizedlist

AudyOdi:
Continuation from PR #41 and following up issue #40, in this PR I added back the props documentation of `VirtualizedList` component. All the documentation and description were taken from https://facebook.github.io/react-native/docs/0.48/virtualizedlist.html with some adjustment to match the current doc style.
Notice that I removed `scrollEventThrottle` from props documentation as it has been covered in `ScrollView`. I also remove `first` and `last` props from the doc as they are actually the `State` of the component (https://github.com/facebook/react-native/blob/ed2bfcb35a2756eb700882ab8e21b6b273efa80a/Libraries/Lists/VirtualizedList.js#L209).

Before:
image
image

Current:
image
image
image

### DIFF
--- a/docs/virtualizedlist.md
+++ b/docs/virtualizedlist.md
@@ -1,7 +1,6 @@
 ---
-id: version-0.50-virtualizedlist
+id: virtualizedlist
 title: VirtualizedList
-original_id: virtualizedlist
 ---
 
 Base implementation for the more convenient [`<FlatList>`](flatlist.md) and [`<SectionList>`](sectionlist.md) components, which are also better documented. In general, this should only really be used if you need more flexibility than `FlatList` provides, e.g. for use with immutable data instead of plain arrays.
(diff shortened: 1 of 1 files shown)


AudyOdi:
Thanks @hramos! As you have covered `SectionList` missing docs in your commit https://github.com/facebook/react-native-website/commit/834370e64f7921f694a3990b4aeac872fb5de0d3, all the missing docs that I found in https://github.com/facebook/react-native/issues/16501 are covered. Please let me know if there is any missing docs left. Happy to help!

devlinb:
Details for functions like scrollToIndex are still missing. It'd be nice to know what the members of the object are there for, but just knowing what they are would be great. :)

hramos:
You can take a look at the source code in the facebook/react-native repo, and go from there. We need help getting this documented :)

PR #54, opened 2017-12-10
[Performance] Add info about debuggerWorker.js thread

brunolemos:
Select debuggerWorker.js thread under Source panel


brunolemos:
Thanks 👍

PR #75, opened 2017-12-15
Updated tutorial.md to fix export error

LuisRizo:
As shown in [#17208](https://github.com/facebook/react-native/issues/17208) in facebook/react-native, "`App.js` should not include `AppRegistry.registerComponent('SampleApp', () => App)`;. That is only needed in `index.js.`"

Replace `AppRegistry` with `export default App;`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -13,7 +13,7 @@ In accordance with the ancient traditions of our people, we must first build an 
 
 ```ReactNativeWebPlayer
 import React, { Component } from 'react';
-import { AppRegistry, Text } from 'react-native';
+import { Text } from 'react-native';
 
 export default class HelloWorldApp extends Component {
   render() {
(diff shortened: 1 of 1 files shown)


hramos:
Thanks for flagging this! We really need to rework the tutorial so that it builds upon prior steps.

I'm not sure this is the proper fix. That line can be removed entirely since "HelloWorldApp" is already exported, though you would need to update your index.js to ensure it refers to HelloWorldApp and not just "App".

The proper long term fix would be to either use "App" as the name of the app on all the sample code in the tutorial, or to include explicit instructions that guide the user through creating new files as well as updating index.js.

LuisRizo:
@hramos You're totally right. I missed that. However, after running some tests, I realized that the `index.js` file just imports `App.js` using `AppRegistry` in the case of `react-native init` apps.

For that reason, deleting `AppRegistry` completely from `App.js` is a better option, as it is not required in any case.

PR #76, opened 2017-12-18
Release 0.52-RC

hramos:
Cut a new version of the docs for RC 52


docs/building-for-apple-tv.md
```diff
@@ -92,7 +92,7 @@ The RNTester app supports Apple TV; use the `RNTester-tvOS` build target to buil
 * _JavaScript layer_: Support for Apple TV has been added to `Platform.ios.js`. You can check whether code is running on AppleTV by doing
 
 ```javascript
-var Platform = require('Platform');
+var Platform = require("Platform");
```
charpeni: Hmm, it shouldn't be a single quote here?

hramos: You're right. Looks like I have a difference in configuration in Prettier on my IDE vs the one we run in this repo. After running `yarn prettier`, this was fixed.

Issue #71, opened 2017-12-14
Can't click on link/button that appears next to each prop title


rigdern:
Each prop title has a link/button to the left of it:

![image](https://user-images.githubusercontent.com/199935/34018662-be3a100c-e0df-11e7-9067-678e120d4d81.png)

This link/button only appears when you hover over the text representing the prop name. Consequently, you cannot click on the link/button because it disappears when you attempt to hover over it in order to click it.

Adam Comella
Microsoft Corp.

Tonours:
Can i take this one ?

hramos:
Thanks for the report. Looks like an issue with Docusaurus, the site generator. I've duplicated this issue in their repo. I'll leave this open until it's fixed in Docusaurus.

hramos:
@Tonours's comment hadn't loaded when I posted above (I had an old tab open). Thanks for looking into it, interesting to see it might be an issue with our CSS and not on Docusaurus's side. Let me take a look and make sure this is not something we need to upstream to Docusaurus.

hramos:
Fix looks great, merged.

Tonours:
@hramos Hi ! Yes at the beginning i thought it was in Docusaurus but this is an override in react-native.css :-)

PR #79, opened 2017-12-19
Fix hover anchor click.

Tonours:
Due to `margin-right: 0;` of `.docsNavContainer` `.hash-link` became 'un-clickable' when title parent missed hover.
So `margin-right: 0;` has been removed in order to fix this bug.

**Actual behavior:**
capture d ecran 2017-12-18 a 21 57 52

**Fixed behavior:**
capture d ecran 2017-12-18 a 21 58 03

#71

### DIFF
--- a/website/static/css/react-native.css
+++ b/website/static/css/react-native.css
@@ -188,12 +188,6 @@ p {
   }
 }
 
-@media only screen and (min-width: 1024px) {
-  .docsNavContainer {
-    margin-right: 0;
-  }
-}
-
 .prose h1 {
   padding: 0 0;
 }


hramos:
Thank you!

charpeni:
Hey @Tonours, thanks. 

Unfortunately, I don't think that's the right fix we want here.

Removing `margin-right: 0` add exaggerated margin to the container see details below.

<details>

#### Before
image

#### After
image

</details>

____

`.docsNavContainer` have a width of `300px` but `.toc` have a width of `240px`. 

I'd say the real fix could be a `.docsNavContainer` with a width of `240px` to match with width of the inner nav and apply the difference (60px) to `.mainContainer` with `margin-left: 60px`.

cc @hramos.

Tonours:
@charpeni Hi ! Thanks for your feedback, if you want i can do this fix later this day, is it good for you ?

charpeni:
@Tonours That would be appreciated if you have time. Otherwise, I'll do it in the next few days.

Merci!

PR #88, opened 2017-12-20
Update appstate.md with note on Android activities

Minishlink:
Hello,

On Android, it appears that the AppState becomes [`background`](https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/ReactAndroid/src/main/java/com/facebook/react/modules/appstate/AppStateModule.java#L59) when you start a new intent from your app. It is worth mentioning because when you use third-party SDKs that rely on other activities, you'll enter this state. On iOS, the behavior is not the same (AppState stays active).


hramos:
Thanks for the PR!

PR #106, opened 2017-12-31
Update Image resizeMode=repeat docs

motiz88:
1. Android implementation pending review: https://github.com/facebook/react-native/pull/17404
2. Current behaviour on iOS (as well as on Android per https://github.com/facebook/react-native/pull/17404) differs from what's documented for an image that's larger than the view.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -249,7 +249,7 @@ Determines how to resize the image when the frame doesn't match the raw image di
 
 * `stretch`: Scale width and height independently, This may change the aspect ratio of the src.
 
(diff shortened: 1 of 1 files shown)


charpeni:
We'll wait for facebook/react-native#17404 to be reviewed and merged before.

motiz88:
I've corrected the documentation in light of the actually-implemented scaling behaviour (see item 2 in my edited comment above). This could in principle be broken out into its own PR as it needn't be blocked on facebook/react-native#17404. Should I go ahead and do that?

PR #120, opened 2018-01-08
Added `spaceEvenly` value to `justifyContent`

turnrye:
In 0.52.0-RC, value `justifyContent` was added to `spaceEvenly` (facebook/react-native@1050e0b). Update the docs to reflect this. Note that this is already explained in the existing link to Mozilla, so I don't believe we need to add any more detail of what `spaceEvenly` is.


charpeni:
Can you also apply your changes to the versioned docs in `website/versioned_docs/version-0.52-RC` ?

https://github.com/facebook/react-native-website/blob/master/website/versioned_docs/version-0.52-RC/layout-props.md

turnrye:
@charpeni done.

PR #122, opened 2018-01-08
Update images.md

MoonBall:
File structure differs from the code `<Image source={require('./assets/check.png')} />`.


hramos:
Thanks for your PR! I believe we just addressed this in a separate PR that just landed.

PR #125, opened 2018-01-08
Remove out of date reference to ResponderEventPlugin.js

rick-schubert:
The file ResponderEventPlugin.js is nowhere to be found in the source code of react-native although it is referenced by name 33 times in 6 files. ResponderEventPlugin can be found inside of Libraries\Renderer\ReactNativeRenderer-prod.js though.

Having had a look at ReactNativeRenderer-prod.js I realised that there are neither details nor documentation inside of this file. Thus, I would propose to completely remove this sentence.


hramos:
Thanks, looks like this file is not present in open source.

PR #127, opened 2018-01-10
Update images.md

eXhausted:
Folder is named as img, but path contains assets folder


hramos:
I believe I just merged a PR that fixes this.

PR #131, opened 2018-01-11
Added BatchedBridge to Podfile in Integration-Doc as it is required for #15049

itinance:
// If your PR references an existing issue, please add the issue number below:

#15049

In order to get the integration-guide done with already existing iOS-Apps, a specific Pod is required called BatchedBridge. 

See https://github.com/facebook/react-native/issues/15049#issuecomment-316115595

### DIFF
--- a/docs/integration-with-existing-apps.md
+++ b/docs/integration-with-existing-apps.md
@@ -213,6 +213,7 @@ target 'NumberTileGame' do
   pod 'React', :path => '../node_modules/react-native', :subspecs => [
     'Core',
     'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
+    'BatchedBridge', # Fix for [#15049](https://github.com/facebook/react-native/issues/15049)
     'RCTText',
     'RCTNetwork',
     'RCTWebSocket', # needed for debugging


charpeni:
@itinance What version of react-native are you using? 

I thought that `BatchedBridge` is only relevant for react-native < 0.47.

itinance:
@charpeni latest release, tested today :) (RN 0.52)

charpeni:
Since RN 0.47 you should be able to use the cxx bridge instead of batched bridge.

I think it should be something like this:
```pod
   pod 'React', :path => '../node_modules/react-native', :subspecs => [
      'Core',
      'CxxBridge', # Include this for RN >= 0.47
      'BatchedBridge', # Include this for RN <= 0.46
      'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
      'RCTText',	
      'RCTNetwork',
```

@itinance are you up to test swaping `BatchedBridge` by `CxxBridge` in your RN 0.52 project?

But anyway you're right, the bridge subspec is missing in the documentation. We'll get it sorted out.

charpeni:
Thanks @itinance, `CxxBridge` have been fixed in the latest releases and I've fixed remaining ones in #268.

PR #134, opened 2018-01-12
Add configuration for deploy preview

charpeni:
I was tired to checkout every branch to review visual changes and I think that deploying each pull requests could be useful and a time saver for maintainers as well as contributors.

We could easily achieve this with Netlify and their deploy preview service.

Here's deploy preview in action on `react-native-website`: https://github.com/charpeni/react-native-website/pull/3.

image

Note that [`siteConfig.js`](https://github.com/facebook/react-native-website/blob/master/website/siteConfig.js) will be altered in Netlify infra because we need `baseUrl` to be `/`.

For the additional configuration on GitHub & Netlify let's hook up on Slack @hramos.

🦖 🤖 :tada:

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,4 +1,4 @@
 [build]
   base    = "website"
-  publish = "website/build"
-  command = "yarn install && yarn build"
+  publish = "website/build/react-native"
+  command = "sed -i -e \"s|const baseUrl .*|const baseUrl = '/';|g\" siteConfig.js && yarn install && yarn build"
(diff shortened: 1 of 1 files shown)


Daniel15:
It would be good to actually move the site onto Netlify one day (especially since the React site is on Netlify now).

hramos:
@Daniel15 I think the biggest blocker right now is the lack of a vanity URL. React can do it because they have reactjs.org. We do not own reactnative.com, sadly.

We'd like to at some point just have the site be part of ReactJS.org so we might just get to this eventually :)

Daniel15:
@hramos  Given we already own `reactjs.org`, what about `native.reactjs.org`? 😃

PR #136, opened 2018-01-13
Documentation for Yarn

vivekkiran:
Updated documentation for using Yarn


hramos:
Hi! Thanks for the PR. Up until now we've explicitly avoided mentioning Yarn in the Getting Started, in order to keep it simple, as many people will have access to npm once they install Node, therefore reducing the number of steps. We want to make sure it's clear when something is optional as opposed to required.

I think it's time we do add Yarn instructions to the Guide. I'd probably go with a different approach, where if the reader already has Yarn installed, they can use it in place of npm. As it is, I am not sure we should prescribe everyone to install Yarn unless it's required.

PR #137, opened 2018-01-13
fixes type for ItemSeparatorComponent prop

nathmack:
The `ItemSeparatorComponent` prop for `FlatList` seems to take a function, not a component.


hramos:
It's likely a function that is expected to return a component. How is this handled in other areas of the docs?

nathmack:
@hramos yeah, it is. It behaves the same way `renderItem` does, and that describes the prop type as a function

PR #138, opened 2018-01-14
Update asyncstorage.md

hovi:
Hello, just very simple change. That part of sentence "AsyncStorage instead of AsyncStorage". Doesn't make sense to me and I guess this is what was meant instead.


hramos:
Hmm, I'm not familiar with this API. This could just be poorly written. It can be read as a suggestion to write an abstraction over AsyncStorage, as opposed to simply using AsyncStorage directly.

charpeni:
I agree with @hramos.

PR #144, opened 2018-01-17
fix: move features into correct headings

craigbilner:
This PR moves each feature into a correct/updated heading from when it was first written.

Admittedly it's quite an odd page with a sprinkling of some features but not all, and potentially should use the ES20XX naming convention instead?

### DIFF
--- a/docs/javascript-environment.md
+++ b/docs/javascript-environment.md
@@ -40,11 +40,14 @@ ES6
 * [Rest Params](https://github.com/sebmarkbage/ecmascript-rest-spread): `function(type, ...args) { }`
 * [Template Literals](http://babeljs.io/docs/learn-es2015/#template-strings): `` var who = 'world'; var str = `Hello ${who}`; ``
 
-ES7
+ES8
 
-* [Object Spread](https://github.com/sebmarkbage/ecmascript-rest-spread): `var extended = { ...obj, a: 10 };`
 * [Function Trailing Comma](https://github.com/jeffmo/es-trailing-function-commas): `function f(a, b, c,) { }`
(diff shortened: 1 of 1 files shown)


docs/javascript-environment.md
```diff
@@ -66,12 +69,16 @@ Browser
...
 * [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)
-* Array.prototype.{[find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex), [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)}
+* Array.prototype.{[find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)}
 
 ES7
 
+* Array.prototype.{[includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)}
+
+ES8
```
hramos: Shouldn't we move this to the ES8 heading above then?

hramos:
This page has not been touched in a while, thanks for looking at it! Feel free to update the layout if you think it would make it clearer.

docs/javascript-environment.md
craigbilner: There's the polyfill/transformation split?

craigbilner:
@hramos I would just remove the majority of the page and put links to the babel config/docs which are pretty thorough and then it won't ever be incorrect/out of date/out of sync?

PR #153, opened 2018-01-25
Update sectionlist.md

mafelix:
<h4> Work performed: </h4>

Update SectionList documentation to list React Native version dependencies

<h4> Issue number: </h4>

[#17731](https://github.com/facebook/react-native/issues/17731)


hramos:
Thanks for the PR! The docs are already versioned. If you visit older version pages, you'll notice this API is not listed there. 

I'm closing as I think this won't be required.

PR #154, opened 2018-01-25
No need to install / link any library

marcoliceti:
CameraRoll API doesn't require any library install / link.

### DIFF
--- a/docs/cameraroll.md
+++ b/docs/cameraroll.md
@@ -3,7 +3,7 @@ id: cameraroll
 title: CameraRoll
 ---
 
-`CameraRoll` provides access to the local camera roll / gallery. Before using this you must link the `RCTCameraRoll` library. You can refer to [Linking](linking-libraries-ios.md) for help.
+`CameraRoll` provides access to the local camera roll / gallery.
 
 ### Permissions
 


hramos:
Did this change recently? I'm not familiar with this, just want to understand how you came to this conclusion. Happy to merge if this is now the case.

marcoliceti:
Hi don't know when it changed (I'm new to React Native) but [the docs](https://facebook.github.io/react-native/docs/cameraroll.html) say that the RCTCameraRoll library is needed and provide a link to the [linking guide](https://facebook.github.io/react-native/docs/linking-libraries-ios.html).

...but I succesfully ran the example in the docs on my Android device without installing / linking anything. My guess is that:

1. The install / link step is not needed anymore.
2. It's only needed for iOS.

I can't verify option 2 since I don't have a Mac, but it should be easy:

1. `react-native init TestCameraRoll`
2. Copy the [example code from here](https://facebook.github.io/react-native/docs/cameraroll.html).
3. Run it on an iOS device.

hramos:
The current docs state this is applicable on iOS, not Android. I'll close this for now.

marcoliceti:
I'm sorry, I reached the ["Linking Libraries"](https://facebook.github.io/react-native/docs/linking-libraries-ios.html) page directly from the [CameraRoll](https://facebook.github.io/react-native/docs/cameraroll.html) docs, so it wasn't clear that it applies only to iOS.

Since `CameraRoll` works on both platform, maybe it would be helpful to change from:

> Before using this you must link the RCTCameraRoll library. You can refer to Linking for help.

to:

> **On iOs** you must link the RCTCameraRoll library. You can refer to Linking for help.

hramos:
Fixed in 4838dac91c6e6bd0045db5bc527437238203ff98

PR #156, opened 2018-01-26
Update docs for https://github.com/facebook/react-native/pull/17747

msand:
Update docs for https://github.com/facebook/react-native/pull/17747

### DIFF
--- a/docs/android-building-from-source.md
+++ b/docs/android-building-from-source.md
@@ -11,9 +11,9 @@ Assuming you have the Android SDK installed, run `android` to open the Android S
 
 Make sure you have the following installed:
 
-1. Android SDK version 23 (compileSdkVersion in [`build.gradle`](https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle))
-2. SDK build tools version 23.0.1 (buildToolsVersion in [`build.gradle`](https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle))
-3. Android Support Repository >= 17 (for Android Support Library)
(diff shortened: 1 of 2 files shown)


hramos:
Thanks, let's come back to this when the PR is merged.

hey99xx:
Wait how does this commit makes sense without updating build.gradle of react-native simultaneously? 

This is changing

> Android SDK version ~~23~~ 27 (compileSdkVersion in [`build.gradle`](https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle))

and 

> SDK build tools version ~23.0.1~~ 26.0.2 (buildToolsVersion in [`build.gradle`](https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle))

But if you look at the [relevant part of build.gradle](https://github.com/facebook/react-native/blob/0.54-stable/ReactAndroid/build.gradle#L234) the numbers there are still are

```
android {
    compileSdkVersion 23
    buildToolsVersion "23.0.1"
```

So I think merging this PR as it is will only create confusion, because the docs will suggest Android tools that have a mismatch with what RN actually uses.

msand:
@hey99xx I had updated and force-push to the branch in the PR in RN, and am waiting for more feedback https://github.com/facebook/react-native/pull/17747#issuecomment-363969697 
Haven't updated this PR to reflect the changes in versions there yet.

msand:
But, the instructions about build tool versions could probably be removed completely, as the new gradle version will pick one itself depending on compatibility.

cpojer:
It seems like this is outdated now and we'll need a new PR with the latest versions for everything. Sorry that it took so long to get back to this.

PR #170, opened 2018-01-31
Add missing method on PushNotificationIOS

lwinkyawmyat:
I have been changed for missing method called `presentLocalNotification` on `PushNotificationIOS`.

Fixed broken link on `JavaScript Environment`.


hramos:
Awesome, thanks for fixing master as well as past versions!

PR #169, opened 2018-01-30
Include a native module preparation guide …

jugutier:
with code scaffolding tool so that readers can jump straight into the code

Thank you for the PR! Contributors like you keep React Native awesome!

Please see the Contribution Guide for guidelines:

https://github.com/facebook/react-native-website/blob/master/CONTRIBUTING.md

If your PR references an existing issue, please add the issue number below:

#163

PR #178, opened 2018-02-03
Update README for contributing

isaacchien:
Just started contributing to this repo today and had trouble viewing changes I made to some markdown files. Thanks to @charpeni's comment on #169, I realized you have to switch to the MASTER version. This update clarifies that in the README. I also figured the README should remind people to fork the repo when cloning like the README in [reactjs](https://github.com/reactjs/reactjs.org).

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,10 @@ React Native showcase.
 ### Make the change
 
 1. Follow the "Running locally" instructions 
-1. `open http://localhost:3000/react-native/versions.html` to see other versions.
 1. Save the files and check in the browser. Some changes may require a server restart.
+1. Changes to /docs will only be visible in the latest version of the documentation (master). 
+
+`open http://localhost:3000/react-native/versions.html`to see other versions.
 
 ### Test the change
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,9 +138,9 @@ React Native showcase.
 
 ### Make the change
 
-1. Follow the "Running locally" instructions
-1. Save the files and check in the browser. Some changes may require a server
-   restart.
+1. Follow the "Running locally" instructions 
+1. `open http://localhost:3000/react-native/versions.html` to see other versions.
```
charpeni: I'd say something like:

>1. Follow the "Running locally" instructions 
>1. Save the files and check in the browser. Some changes may require a server restart.
>1. Changes to `/docs` will only be visible in the latest version of the documentation (master). 
>`open http://localhost:3000/react-native/versions.html` to see other versions.

The files should be saved before checking versions and this add clarification on why we should see other versions.

What do you think?

isaacchien: Yup makes sense. I added an update with your suggestion.

PR #180, opened 2018-02-03
Fix `start/end_cursor` type (`CameraRoll`)

mobily:
This PR fixes the wrong type of `start_cursor` and `end_cursor` properties.

Screenshot:

react native debugger 2018-02-03 16-45-23


mobily:
thanks! 🎉

PR #184, opened 2018-02-06
Update broken links.

JChord:


charpeni:
The link should be the updated guide: https://reactnavigation.org/docs/redux-integration.html

PR #197, opened 2018-02-11
Add link to install instructions for emulators

alexkrolick:
The Quick Start doesn't mention where the instructions for running on emulators actually are in the docs; it's not obvious this means the tab at the top of the page.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -137,7 +137,7 @@ If you're curious to learn more about React Native, continue on to the [Tutorial
 
 ### Running your app on a simulator or virtual device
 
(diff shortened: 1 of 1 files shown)


alexkrolick:
Ok, it's not showing in the deploy preview. Why not though...?

charpeni:
Your changes are only applied to master (next version).

image

https://deploy-preview-197--react-native.netlify.com/docs/next/getting-started.html#now-what

alexkrolick:
👍 thanks for that

Looks like it may require some JS to switch the tabs when switching within the page. I'll check it out later.

cpojer:
Seems like this PR is stale so I'm going to close it.

PR #206, opened 2018-02-19
Preserve quotes in udev config (they are required)

rxaviers:
Problem: the `echo` command (as is in the docs) discards the double quotes (they are expanded by the shell). The double quotes must be preserved for proper udev config loading.


rxaviers:
For reviewers: https://deploy-preview-206--react-native.netlify.com/docs/running-on-device.html (Android > Linux), though I don't see my update in this preview.

PR #212, opened 2018-02-22
Simplify

olakaldestad:
Simplified setState


hramos:
This change is technically correct, but I am hesitant to accept it. When we write sample code, we tend to err on the side of clarity. Not all RN developers may be familiar with modern JavaScript and may miss the fact that the change you're proposing is equivalent.

PR #231, opened 2018-03-02
add filenames to example (Toast Example)

hhg2288:
I'm not familiar with Android development, but since I've got a client that needs a native Android module I looked at the docs to create one.

Because I don't know much about Android development, I lost a lot of time figuring out where & how I needed to add the files needed to create the Toast example.

Maybe being more specific regardless what you need to create & how it would help others like me to save time :)

hopefully this little contribution helps!

### DIFF
--- a/docs/native-modules-android.md
+++ b/docs/native-modules-android.md
@@ -135,7 +135,7 @@ The package needs to be provided in the `getPackages` method of the `MainApplica
 protected List<ReactPackage> getPackages() {
     return Arrays.<ReactPackage>asList(
             new MainReactPackage(),
-            new AnExampleReactPackage()); // <-- Add this line with your package name.
+            new CustomToastPackage()); // <-- Add this line with your package name.
 }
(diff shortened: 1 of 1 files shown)


docs/native-modules-android.md
```diff
@@ -105,7 +109,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-public class AnExampleReactPackage implements ReactPackage {
+public class CustomToastPackage implements ReactPackage {
```
charpeni: The class name should match with the package name below.

```diff
protected List<ReactPackage> getPackages() {
    return Arrays.<ReactPackage>asList(
            new MainReactPackage(),
-           new AnExampleReactPackage()); // <-- Add this line with your package name.
+           new CustomToastPackage()); // <-- Add this line with your package name.
}
```

hhg2288:
@charpeni done!

PR #234, opened 2018-03-05
docs: getting-started: change npm to yarn in some commands

xdevs23:
After creating a new app using create-react-native-app it suggests to use yarn instead of npm.
Since yarn is more secure than npm and effectively has the same functionality, swap out the npm commands for the yarn commands as described after creating a new app:

```
Success! Created AwesomeReactApp at /home/simao/workspace/AwesomeReactApp
Inside that directory, you can run several commands:

  yarn start
    Starts the development server so you can open your app in the Expo
    app on your phone.

  yarn run ios
    (Mac only, requires Xcode)
    Starts the development server and loads your app in an iOS simulator.

  yarn run android
    (Requires Android build tools)
    Starts the development server and loads your app on a connected Android
    device or emulator.

  yarn test
    Starts the test runner.

  yarn run eject
    Removes this tool and copies build dependencies, configuration files
    and scripts into the app directory. If you do this, you can’t go back!

We suggest that you begin by typing:

  cd AwesomeReactApp
  yarn start

Happy hacking!

```

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -108,7 +108,7 @@ Then run the following commands to create a new React Native project called "Awe
 create-react-native-app AwesomeProject
 
 cd AwesomeProject
-npm start
+yarn start
 ```
 
 This will start a development server for you, and print a QR code in your terminal.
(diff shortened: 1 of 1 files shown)


hramos:
There's been PRs in the past that attempt a similar change. If we are going to do this, I'd like to see a PR that changes substantially more files than just the Getting Started page. We need to be consistent and if people are instructed to install yarn, then it should be used everywhere npm is used.

Note that yarn is only a recommendation when using CRNA. You can still use npm. In any case, you'll probably want to update the native code instructions as well.

Ideally, we'd provide instructions for using both npm and yarn.

I apologize for asking for so many changes, but I feel strongly about making a significant change like this. Changing these instructions have consequences for every new user.

cpojer:
This is a stale PR so I'm going to close it. If you plan to work on this again, please submit a new PR :)

PR #235, opened 2018-03-06
Update the AsyncStorage documentation about its problem on Android unsolved since May

yairopro:
Update the AsyncStorage documentation about its problem on Android unsolved since May 2017.
See [issue here](https://github.com/facebook/react-native/issues/14101#issuecomment-345563563).
#14101


hramos:
Thanks for the PR. We usually don't link to issues. Rather, can you explain what workaround people might need?

PR #247, opened 2018-03-11
Update improvingux.md

joegoodall1:
Add backticks to indicate inline code blocks and change from smart quotes to dumb quotes (smart quotes will not work in AndroidManifest.xml).


joegoodall1:
Hi @charpeni, just wondering when the website will update with this commit?

charpeni:
It looks like your changes got overwritten, I'll check this.

PR #251, opened 2018-03-13
add permission write access to node_module global

Arshiamidos:
flags added --unsafe-perm=true --allow-root

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -42,7 +42,7 @@ The `react-native-git-upgrade` module provides a CLI and must be installed globa
 ```sh
 $ npm install -g react-native-git-upgrade
 ```
-
+if you have issue like write access just put `--unsafe-perm=true --allow-root` and run with `sudo` command. 
 #### 3. Run the command
 
 Run the following command to start the process of upgrading to the latest version:


Arshiamidos:
pls add this PR to hint other how install fast 
i searched many site to know this :)

hramos:
Can you elaborate on how this helps others?

Arshiamidos:
yes  
by default it can't install on mac straight forward .and when i searched for error i faced with many errors again :|
only by adding this flags i could install package wth admin previlage
i hope explain well

charpeni:
Hmm, this looks more like an issue with your node installation and we should be fine without allowing root and running this with `sudo`.

Have you installed `node` with [brew](https://brew.sh/) or [nvm](https://github.com/creationix/nvm)?

Arshiamidos:
if manual consist to install with **npm**,it must solve with **npm** .
but maybe we face another problem with __brew__ or __nvm__?
i think just hinting npm may be good solution
BTW it's not issue it's global hint with node.😉

cpojer:
Thank you for your PR. I don't think we want to include this in the documentation at this time so I'm going to close this.

PR #259, opened 2018-03-18
Update keyboard example.md

moskh:
According to the official React documentation on ComponentWillMount method: 

> Avoid introducing any side-effects or subscriptions in this method. For those use cases, use componentDidMount() instead.

### DIFF
--- a/docs/keyboard.md
+++ b/docs/keyboard.md
@@ -14,12 +14,12 @@ import React, { Component } from 'react';
 import { Keyboard, TextInput } from 'react-native';
 
 class Example extends Component {
-  componentWillMount () {
+  componentDidMount () {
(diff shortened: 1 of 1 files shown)


docs/keyboard.md
```diff
@@ -19,7 +19,7 @@ class Example extends Component {
     this.keyboardDidHideListener = Keyboard.addListener('keyboardDidHide', this._keyboardDidHide);
   }
 
-  componentWillUnmount () {
```
charpeni: You're right about `componentWillMount` documentation, but here we have `componentWillUnmount`, which is the right place to remove listeners.

moskh:
Yes you are right, I just corrected the componentWillMount method.

PR #264, opened 2018-03-22
Update Fresco version in the docs

AndrewJack:
Following the Fresco update in the upcoming version, I have updated the version of fresco in the docs. 

Related issue: https://github.com/facebook/react-native/pull/18496


charpeni:
Oops, Fresco update is not included in RN 0.55. I'll move this to the next version.

PR #286, opened 2018-04-07
Fix SectionList example

mcordini:
SectionList Example 1 is displayed incorrectly.
image


charpeni:
Oops, good catch! 

Looks like the tabulations have been overridden, I've changed the tab code blocks to proper markdown and also apply this to versioned docs of 0.55.

PR #289, opened 2018-04-09
Add prop enabled for KeyboardAvoidView document follow RN 0.54.

TronNatthakorn:
KeyboardAvoidingView have prop enabled and mark as required.


hramos:
Can you make the changes to the 0.54 version as well?

charpeni:
I've done it.

PR #300, opened 2018-04-13
GrabPay Merchant App added in the showcase

grabvinod:
Added GrabPay Merchant App as showcase in React Native website.


charpeni:
Hi! We're not yet ready to accept new submissions. For reference, in the past we only added submissions that contained a deep technical article on how the app was built with React Native.

grabvinod:
@charpeni Can yo confirm by when you guys will be ready to accept new submission. This is quite wierd to know you don't accept new submission without technical article while I could see few apps there without article. if that is the case please let me know we can get that done.

hramos:
Hi @grabvinod, I made the decision not to accept any PRs that modify the showcase. The criteria is no longer based on whether a technical article is submitted or not. This is something that I now curate by hand.

PR #303, opened 2018-04-13
Added import to documentation

danielbenedykt:
Thank you for the PR! Contributors like you keep React Native awesome!

Please see the Contribution Guide for guidelines:

https://github.com/facebook/react-native-website/blob/master/CONTRIBUTING.md

If your PR references an existing issue, please add the issue number below:

#<Issue>

### DIFF
--- a/docs/linking.md
+++ b/docs/linking.md
@@ -23,6 +23,8 @@ title: Linking
 If your app was launched from an external url registered to your app you can access and handle it from any component you want with
 
 ```
+import {Linking} from "react-native";
+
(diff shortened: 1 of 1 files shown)


charpeni:
In fact, if we add the import in the example, we should wrap the lifecycle inside a proper React class.

cpojer:
Thank you for your PR. I think this example isn't great overall but just adding the import isn't really solving that. We are planning on rewriting most of this in the near future so I'm going to close this PR.

PR #312, opened 2018-04-17
Update style.md

DaCurse:
Being more informative about CSS' casing


charpeni:
What's your motivation for this change?

I think it can lead to confusion if we quote another case.

charpeni:
I'll close this PR for now, let me know your motivation to reopen it.

PR #326, opened 2018-04-25
Update network.md

manoelangelo:


docs/network.md
```diff
@@ -23,7 +23,7 @@ Fetch also takes an optional second argument that allows you to customize the HT
 fetch('https://mywebsite.com/endpoint/', {
   method: 'POST',
   headers: {
-    Accept: 'application/json',
+    'Accept': 'application/json',
```
charpeni: I'm not sure what to do here.

What's the purpose of this? Object keys shouldn't needed to be surround by quotes, except if they're having a special character.

PR #343, opened 2018-05-07
iOS supports needsOffscreenAlphaCompositing

hfossli:
https://github.com/facebook/react-native/pull/19052


hramos:
If the PR lands, we can simply remove the platform column.

Issue #345, opened 2018-05-07
Lack of documentation for `.native.js` extensions


drochetti:
Although supported, `.native.js` extension docs can't be found anywhere. There are GitHub issues on related projects talking about it, but nothing official that projects can refer to. Since this is a common feature used by projects that share code between React Native and ReactJS there should have official documentation for it.

PR #346, opened 2018-05-07
add documentation about `.native.js` extensions

drochetti:
Lots of projects use the `.native.js` to share code between ReactJS and React Native (including core projects, like Jest) but there's no official documentation about it. I'm adding some info so projects can easily discover and refer to this feature.

Related issue: #345


docs/platform-specific-code.md
```diff
@@ -102,3 +102,22 @@ const BigButton = require('./BigButton');
...
+
+```sh
+Container.js # picked up by Webpack, Rollup or any other Web bundler
+Container.native.js # picked up by the React Native bundler for both Android and iOS (Metro)
+```
+
+You can still require it without the `.native` extension, as follows:
+
+```javascript
+const Container = require('./Container');
```
charpeni: Does the `require` matter here?

I'd say we should prefer `import` here:

```javascript
import Container from './Container';
```

drochetti: Thanks for looking into this. I agree that require may not be required here. It was just to make sure the user knows the effect of `.native.js` is the same of `.ios.js`. I can remove it if you think it's unnecessary.

Regarding the `import` vs `require`, I'm just following the rest of the site. I'm happy to work on another PR that updates the whole site to use `import` instead of `require`.

charpeni: Sorry, I was not clear in my previous message.

I think you're right, the example is important here to make sure the user knowns the effect of the extension omission. But, I'd prefer `import` instead of `require` syntax.

For what I've seen, `require` are only used with images and all Javascript imports are done with ES6 `import` syntax.

PR #350, opened 2018-05-08
Added Alt attribute

saigowthamr:
By adding the attribute to the image tag  accessibility score increased  up to 95% 
in chrome audits tab.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -196,7 +196,7 @@ class AwkwardScrollingImageWithText extends Component {
                 </MarkdownBlock>
               </div>
             </div>
-            <img src="https://media.giphy.com/media/13WZniThXy0hSE/giphy.gif" />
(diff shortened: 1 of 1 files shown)


website/pages/en/index.js
```diff
@@ -196,7 +196,7 @@ class AwkwardScrollingImageWithText extends Component {
                 </MarkdownBlock>
               </div>
             </div>
-            <img src="https://media.giphy.com/media/13WZniThXy0hSE/giphy.gif" />
+            <img src="https://media.giphy.com/media/13WZniThXy0hSE/giphy.gif" alt="hot-loader" />
```
yangshun: `alt` attributes should be natural language descriptions in proper casing (they may be used my screen readers). So it would be more appropriate to call it "Hot reloading demo".

saigowthamr: Sure I will update  it soon..

Issue #316, opened 2018-04-18
Fix author images in blog


amyrlam:
While investigating https://github.com/facebook/react-native-website/pull/314, I noticed that the URL to some author's profile images was broken:

(1)

![image](https://user-images.githubusercontent.com/1372946/38950317-e782c2f2-42f9-11e8-8687-43326491dc53.png)
Image from http://facebook.github.io/react-native/blog/2017/08/30/react-native-monthly-3.html

It seems like the Twitter URL could be used in a more persistent way (right now the URL changes when the profile pic chanages?) re: [this StackOverflow](https://stackoverflow.com/questions/18381710/building-twitter-profile-image-url-with-twitter-user-id?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) which suggests `https://twitter.com/TwitterEng/profile_image?size=original `. This worked for me.

(2)
Some Facebook URLs also seem broken. 

![image](https://user-images.githubusercontent.com/1372946/38950380-2e0a7918-42fa-11e8-9901-977fa69390fb.png)
Image from http://facebook.github.io/react-native/blog/2015/09/14/react-native-for-android.html

To fix these, could just use `authorFBID` and delete `authorURL` re: https://docusaurus.io/docs/en/blog.html. 

//

I can make these changes just wanted to file an issue first. I'm working on Docusaurus as part of the Facebook open source mentorship program. cc @yangshun

amyrlam:
It turns out that using the URL https://twitter.com/grabbou/profile_image?size=original doesn't work with Docusaurus. That URL redirects you to a link like https://pbs.twimg.com/profile_images/988860423897313281/L9ErG_lr.jpg. 

The pbs.twimg.com link changes when you upload a new profile picture, see [this Twitter documentation](https://developer.twitter.com/en/docs/accounts-and-users/user-profile-images-and-banners) - "The URL may have changed, which happens for instance when the user updates their profile image."

We can update the broken images in react-native-website for now, but we should come up with a better long term solution.

//

It also seems like the `authorFBID` isn't a foolproof solution. I attempted to change the file `/react-native/blog/2015/11/23/making-react-native-apps-accessible.html`

```
authorURL: https://www.facebook.com/georgiy.kassabli
authorFBID: 795592927136196
```

A call to the URL that Docusaurus is building gives https://graph.facebook.com/795592927136196/picture/?height=200&width=200 which has a permissions error?

amyrlam:
[...]

PR #356, opened 2018-05-11
Update broken profile images

amyrlam:
Resolves part of: https://github.com/facebook/react-native-website/issues/316.

WAS:
image

IS:
image

WAS:
image

IS:
image

WAS:
image

IS:
image

WAS:
image

IS:
image

***

Note: `2015-11-23-making-react-native-apps-accessible` still has a broken image. The `authorFBID` does not work here. Noted on https://github.com/facebook/react-native-website/issues/316.

The Twitter URLs are not a permanent fix as they change when a new picture is uploaded, also noted on https://github.com/facebook/react-native-website/issues/316.

cc @yangshun


amyrlam:
It seems like a better long-term solution may be to just have `authorImageURL` link to Gravatar or a similar service for all the blog post images.

PR #363, opened 2018-05-14
Adds missing documentation for WebView methods

kristinns:
These methods can be seen in the source files for both Android and iOS
- https://github.com/facebook/react-native/blob/master/Libraries/Components/WebView/WebView.ios.js
- https://github.com/facebook/react-native/blob/master/Libraries/Components/WebView/WebView.android.js

and in the example
https://github.com/facebook/react-native/blob/master/RNTester/js/WebViewExample.js#L114

### DIFF
--- a/docs/webview.md
+++ b/docs/webview.md
@@ -68,6 +68,10 @@ You can use this component to navigate back and forth in the web view's history 
 ### Methods
 
 * [`extraNativeComponentConfig`](webview.md#extranativecomponentconfig)
+* [`goForward`](webview.md#goforward)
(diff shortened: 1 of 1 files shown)


charpeni:
Thank you for noticing this.

Can you also add each method description from [Webview.ios.js](https://github.com/facebook/react-native/blob/8c74864d7bff16ccfd5179ea399b0bc4adaef69b/Libraries/Components/WebView/WebView.ios.js#L525-L527)?

kristinns:
Yes of course, fixed in 44199ef @charpeni

Issue #362, opened 2018-05-12
How to Contribute Guide: Missing information on how to test local changes


MarkOSullivan94:
Spent a bit of time reading the [How to Contribute guide ](https://facebook.github.io/react-native/docs/contributing.html)and I couldn't find information on how to run local changes to the React Native project on my existing React Native projects.

Would love it if someone could do a detailed explanation within the guide along with how to run the tests on Circle CI / any other CI services which is used by the React Native project.

charpeni:
Thanks for bringing this up.

Is this what you're looking for? https://facebook.github.io/react-native/docs/building-from-source.html

MarkOSullivan94:
Yep! That's perfect.

Would it be possible for that link to be included in the How to Contribute Guide @charpeni?

charpeni:
Sure, are you up to do the PR?

charpeni:
Thank you, your contributions are really appreciated.

PR #366, opened 2018-05-16
Added further guidance for testing within contributing.md

MarkOSullivan94:
#362 

Made changes to try and contain all the required guidance for testing at the top of the document within the Get Involved section


MarkOSullivan94:
@charpeni any idea why I can't see these changes live on the website? 

Also noticed I can see my changes from #189 either.

charpeni:
This PR is not versioned yet, but it can be seen in master (next version). It will be available by default once the next version (0.56) will be released.

https://facebook.github.io/react-native/docs/next/contributing.html

Issue #233, opened 2018-03-05
Refactor PanResponder example to avoid componentWillMount()


benadamstyles:
React Native v0.54 uses React v16.3, which [deprecates various lifecycle methods](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md) including `componentWillMount()`. In the [`PanResponder` docs](https://facebook.github.io/react-native/docs/panresponder.html), the recommended approach is to set up `PanResponder` methods in `componentWillMount()`. Could the docs now be updated with the recommended replacement approach?

charpeni:
Thank you for taking this @andrwsalcdo.

> would it be best to add a description or a link describing the various lifecycle changes

Where would you add this? If it's inside the PanResponder docs, I don't think it worth mentioning as after the refactoring, beginners won't notice the changes from the previous version.

andrwsalcdo:
ah fair, but for those beginners that were learning from outdated resources? eh, it's probably a bit of a stretch, but I thought I ask.

PR #373, opened 2018-05-24
Refactor: PanResponder example to avoid componentWillMount()

andrwsalcdo:
#233 
Puts the panResponder example in the constructor instead of componentWillMount().


wcandillon:
Thank you for that PR 👍🏻

PR #374, opened 2018-05-25
Fix issue the podspec 'glog' mismatched

onyoon7:
An error occurs when running `pod install` as described in the [official documentation](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#configuring-cocoapods-dependencies)

Fix `Podfile` file


charpeni:
Thanks. It has been fixed in 526ab295bfa1091ea125c50579f10ccc9fdcdf6d.

PR #376, opened 2018-05-25
Added KSGR app to the showcase

emin93:
Hi,

First of all: Thank you a lot for the great work. We at Edge5 have a great time with React Native and enjoy using it for customer projects.

We'd like to add the KSGR app to the showcase. We checked with the customer and it's fine for them as well.

Thanks in common.

Cheers,
Emin


charpeni:
Thank you for your contribution, and we appreciate that you enjoy React Native.

PR #386, opened 2018-05-29
Remove 'website' from table of contents

mstrum:
It's currently a link to /dev/null


charpeni:
Good catch!

In fact, Website is now refering to [Updating the documentation](https://facebook.github.io/react-native/docs/testing.html#updating-the-documentation). Can you update it?

mstrum:
Ah, I see, done~

PR #391, opened 2018-05-31
Fix a warning in the Networking Fetch Example

ymainier:
While running the fetch example, I saw this warning :
```ExceptionsManager.js:71 Warning: Failed child context type: Invalid child context `virtualizedCell.cellKey` of type `number` supplied to `CellRenderer`, expected `string`.```

I verified in the [doc](https://facebook.github.io/react-native/docs/flatlist.html#keyextractor), `keyExtractor` should return a `string`.

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -118,7 +118,7 @@ export default class FetchExample extends React.Component {
         <FlatList
           data={this.state.dataSource}
           renderItem={({item}) => <Text>{item.title}, {item.releaseYear}</Text>}
-          keyExtractor={(item, index) => index}
+          keyExtractor={(item, index) => index.toString()}
         />
       </View>
     );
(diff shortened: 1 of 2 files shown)


charpeni:
Thanks for bringing this up.

In fact, it's probably a bad idea to put an anti-pattern like index as key in an example. We don't want people to replicate this into their applications.

I'd say we should add an ID into [`movies.json`](https://github.com/facebook/react-native-website/blob/1e1065c4f07f11b0ebd9f5509106d75591b92c27/website/static/movies.json) and use this id as key. 

Are you up to do this?

ymainier:
Of course :-)

ymainier:
What do you think of the updated pull request ?
Do you prefer to have a new one, containing only one commit ?

PR #394, opened 2018-06-04
Update Docs - Android build from source instructions

marafat:
Update the NDK version required from r10e to r16

## Related PRs
This is to support facebook/react-native#19536


cpojer:
Sorry we missed this one, it seems like this was already updated since.

marafat:
I, perhaps, should have closed this one after I closed #19536 😅 - Thanks!

PR #395, opened 2018-06-05
Update signed-apk-android.md

vijayanbu001:
Removed ./ in the gradlew command


Noitidart:
Why this? In Windows its required I do this otherwise it doesnt work it seems.

vijayanbu001:
If this not works other than windows then need to provide a correct code based on the OS

hramos:
We don't have an easy way of distinguishing between platforms when displaying code examples at this time. I'll separate the commands so people can debug their way out more easily.

PR #402, opened 2018-06-10
Grammar and sentence flow in FlatList docs

BlenderDude:
Modified the getItemLayout section such that it is easier to read through. Even though "a priori" technically a correct term, I feel like changing it will save some people (including me) a Google search.


hramos:
Thank you!

PR #404, opened 2018-06-13
RN expects a string, not a number for keyExtractor

vpicone:
Same issue as here: https://github.com/wonday/react-native-pdf/issues/125


Noitidart:
I think this is a bug in FlatList. It's not something that should be documented.

Noitidart:
Thanks @hramos for verification and thumbs up! :)

PR #410, opened 2018-06-16
Instructions on how to add splash screen for Android

Noitidart:
This is a mix of a couple articles to tell users how to create a splash screen on android. This is the simplest version I could find and it works across a bunch of configurations. Like it works with react-native-fcm as it doesn't use two acivites.

It is a mix of the following articles:

* https://stackoverflow.com/a/44380138/1828637
  * which is based on https://medium.com/react-native-development/change-splash-screen-in-react-native-android-app-74e6622d699
     * which is based on https://www.bignerdranch.com/blog/splash-screens-the-right-way/
* https://medium.com/handlebar-labs/how-to-add-a-splash-screen-to-a-react-native-app-ios-and-android-30a3cec835ae

### DIFF
--- a/docs/splash-screen-android.md
+++ b/docs/splash-screen-android.md
@@ -76,3 +76,11 @@ There are a few ways to do splash screens. This shows one of the simpler and mor
             android:name=".MainActivity"
             android:label="@string/app_name"
     ```
+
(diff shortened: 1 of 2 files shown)


cpojer:
It looks like this PR is stale and the review comments were never addressed so I'm gonna go ahead and close this PR. Please let me know if you wish to take this over the finish line.

PR #418, opened 2018-06-21
Remove random characters from Apple TV example

niksudan:


niksudan:
From a little more digging into this, this looks to be an ellipsis that's been formatted incorrectly.

I can add this back in again and format it correctly if you would prefer.

PR #419, opened 2018-06-21
Add more linking documentation

Tvrqvoise:
**Goals**

 - Add documentation for rnpm 
 - Move `react-native link`'s documentation out of the "Platform Specific (iOS)" section since it applies to both platforms

**Concerns**

 - I don't document `rnpm's` "commands" or "plugins" fields.  Namely because I don't use them and don't feel sufficiently knowledgeable about them to write technical documentation.  That being said, `rnpm` is totally undocumented today, so this still seems like an improvement to me.


hramos:
Thanks for this! I think there's an opportunity to move this to a single guide that is not specific to iOS or Android. As it stands now, the PR results in two similarly named guides: "Linking Libraries", and "Linking Libraries (iOS)". The "(iOS)" portion of the title of this second guide is redundant, as the guide already appears under the "Guides (iOS)" section.

This can be worked around by having a singular "Linking Libraries" document that covers both iOS and Android (it's OK to link to Native Modules for more information for Android developers), or alternatively, using different titles for each guide. I know this is not as straightforward - we have "Native Modules" and "Linking Libraries" guides that each cover related topics - so if re-working all of these docs is too much trouble, you can simply add the new documentation to the existing guide without creating a brand new one.

Tvrqvoise:
Okay, so just to clarify before I make the change: we should have a single page under "Guides" which looks like the following:

```md
## Automatic Linking
[ Describe react-native link ]

## Packaging Assets
[ Describe the package.json rnpm.assets field ]

## Manual Linking

### iOS
[ The content which is currently under the iOS Linking Libraries document ]

### Android
[ Just a link to "Guides (Android): Native Modules" ]
```

Also, how would you feel about renaming this guide to "Importing Native Code" or something similar to avoid conflicts with the "Linking" API guide, which is about an unrelated topic?

hramos:
I like that approach. Perhaps "Working with Native Code" or something along those lines for the title?

cpojer:
We are currently rethinking how we want native modules to be used so I'm going to go ahead and close this PR for now. Please do get involved in the discussion though: https://github.com/react-native-community/discussions-and-proposals/issues/96

PR #430, opened 2018-06-27
Fixes Hello World sample code

MattCheetham:
`View` was not imported so the Hello World sample would not compile.


simonmitchell:
@MattCheetham might need to change this in the versioned file too, although not entirely sure how the site works! https://github.com/facebook/react-native-website/blob/34de6d5fc0bbdd5451d8d443ffcd057fd9ffe757/website/versioned_docs/version-0.5/tutorial.md

PR #431, opened 2018-06-27
fix ReferenceError: View is not defined

VictorWinberg:
fix `ReferenceError: View is not defined` in versioned docs, similar to PR #430


charpeni:
Fixed by #430.

PR #434, opened 2018-06-27
put View on import line

alxaln:
the view was missing here import { Text, ... } from 'react-native';

The **Hello World** example was not working 'cause the **View** was not imported !!!


charpeni:
Fixed by #430.

PR #436, opened 2018-06-27
Pr/broken link fix

ighotouch:
The link NativeMethodsMixin redirects user to a 404 page on github https://github.com/facebook/react/blob/master/src/renderers/native/NativeMethodsMixin.js

### How i fixed this
Updated the link `NativeMethodsMixin` in `direct-manipulation.md` file to reflect the new link

### How i know its working
Tested locally and added screenshots

screen shot 2018-06-27 at 7 01 32 pm

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,5 +7,5 @@ https://github.com/facebook/react-native-website/blob/master/CONTRIBUTING.md
 
 If your PR references an existing issue, please add the issue number below:
 
-#421
+#<Issue>
 -->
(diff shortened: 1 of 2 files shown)


ighotouch:
@charpeni made a few updates in line with the suggested implementation, but also noticed the same issue occurs with `receiveComponent in ReactNativeBaseComponent.js` on this same page `https://facebook.github.io/react-native/docs/direct-manipulation`

hramos:
Maybe link to the file on master instead? https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-prod.js#L5793-L5813

charpeni:
Actually, if we link to master directly, the link won't be permanent, and lines will change on each sync.

As Hector suggested, we could link directly to master instead but without mentioning any lines, a search in the file should do the job. https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-prod.js

Thoughts?

charpeni:
@ighotouch Are you up to do the requested changes?

ighotouch:
@charpeni Added a fix for this as requested. Hope this works?

PR #439, opened 2018-06-29
Added Mobx to the documentation

sarath263:
Since Mobx is also very Popular and gaining popularity. It's Reactive way is something which is nice to have, Requesting Mobx also needed to be added to the React native docs.


sarathxm:
Thank you very much @charpeni

PR #442, opened 2018-07-02
Update navigation.md

leethree:
React Navigation no longer encourage support for `redux` and they will drop support very soon. See https://reactnavigation.org/docs/en/redux-integration.html

This PR is to update the documentation to reflect that.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -5,7 +5,7 @@ title: Navigating Between Screens
 
 Mobile apps are rarely made up of a single screen. Managing the presentation of, and transition between, multiple screens is typically handled by what is known as a navigator.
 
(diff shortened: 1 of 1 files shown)


W7PEA:
Missed a line "React Navigation routers make it easy to override navigation logic **_or integrate it into redux_**. Because routers can be nested inside each other, developers can override navigation logic for one area of the app without making widespread changes."

should be 

"React Navigation routers make it easy to override navigation logic. Because routers can be nested inside each other, developers can override navigation logic for one area of the app without making widespread changes."

charpeni:
@W7PEA Are you up to send a PR?

W7PEA:
https://github.com/facebook/react-native-website/pull/536

PR #458, opened 2018-07-14
Clarify how to include static html

jamesporter:
As of React Native 0.56 to include static html you seem to need to set `originWhitelist` prop otherwise it will fail with an unhelpful error.


charpeni:
Thank you, good catch!

Can you also add this to the versioned docs of 0.56? [`website/versioned_docs/version-0.56/webview.md`](https://github.com/facebook/react-native-website/blob/master/website/versioned_docs/version-0.56/webview.md)

PR #462, opened 2018-07-18
Adds some notes about the CLI config file

orta:
Hello there 👋 

It was noted in https://github.com/facebook/react-native/issues/18967 that maybe these could be some docs about how a `rn-cli.config.js` works. This gives a rough overview and include the one used in TypeScript projects, as it's quite likely that's what a lot of people are using this for.

### DIFF
--- a/docs/understanding-cli.md
+++ b/docs/understanding-cli.md
@@ -24,3 +24,52 @@ module.exports = {
 ### Parameters
 
 The command name identifies the parameters that a command would expect. When the command parameter is surrounded by greater-than, less-than symbols `< >`, this indicates that the parameter is expected. When a parameter is surrounded by brackets `[ ]`, this indicates that the parameter is optional.
+
+### Getting a list of all CLI commands
+
+Running `react-native --help` from inside a React Native project will list all of your current commands. Here is an example from version `0.56`:
+
+```sh
+  Usage: react-native [options] [command]
+
+  Options:
+
+    -V, --version                      output the version number
+    -h, --help                         output usage information
+
+  Commands:
+
+    start [options]                    starts the webserver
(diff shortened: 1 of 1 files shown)


pronebird:
The link to `https://github.com/facebook/metro/blob/master/packages/metro/src/Config.js` is broken :/

orta:
Looks like it got pulled out into it's [own repo](_https://github.com/facebook/metro/blob/master/packages/metro-config/) - the right link is probably https://github.com/facebook/metro/blob/master/packages/metro-config/src/configTypes.flow.js - want to update that?

teameh:
I fixed the broken link in #648. But @orta wouldn't it be better to link to https://facebook.github.io/metro/docs/en/configuration instead? The contents in configTypes.flow.js are not very self explaining and if I get this right, it would now be possible to use a regular metro config for react native instead of a config object with `getBlacklistRE` etc right?

orta:
Pretty sure that didn't exist back then, so yeah - seems like a great idea 👍

teameh:
Check. 

And is my assumption about the config change right? So users should now use: 

```js
module.exports = {
  transformModulePath: require.resolve('react-native-typescript-transformer'),
  sourceExts: ['ts', 'tsx'],
};
 ```

instead of: 

```js
module.exports = {
   getTransformModulePath() {
     return require.resolve('react-native-typescript-transformer');
   },
   getSourceExts() {
     return ['ts', 'tsx'];
   },
 };
 ```

If so that needs to be reflected in the docs as well. Who can verify this?

Javier-Rotelli:
Hi @teameh this has changed, docs should be updated. it took me a while to find this comment :/

PR #474, opened 2018-07-27
Added missing code

Stophface:
The example was not working at all. There was a lot of code missing in order to make it work. Especially the import statements as well as methods in the `java` classes. You further used `RCTImageView` which is areserved keyword from react-native. The example was very hard to follow. It shoud be easier now.


hramos:
Had to revert this, in favor of https://github.com/facebook/react-native-website/pull/464

PR #477, opened 2018-07-30
Rename 'unbundle' to 'Random Access Modules bundle'

jeanlauliac:
We renamed the command in this revision: https://github.com/facebook/react-native/commit/ebf5aeab280f2ebc439ec39d25c48fdf1980cf73

So we should now change docs to reflect that. Is the doc versioned, though? How do people access docs for older versions of React Native, if at all?

Thanks!


hramos:
The docs are versioned. This PR makes the change in the right spot (docs/). The change will only be visible when people switch to "master" in the docs. Once we cut a new version of the framework, we'll do the same for the docs.

PR #479, opened 2018-07-30
Kotlin content for "Integration w/ Existing Apps"

CodingItWrong:
This adds Kotlin instructions to the "Integration w/ Existing Apps" page, to complement the existing Java instructions. @dgreenhalgh and I were able to get a Kotlin app working during a hackathon following these steps, so we can confirm it works. This will lower the barrier to entry of Kotlin devs who aren't as familiar with Java, or don't want to go through the syntax translation.

### DIFF
--- a/docs/integration-with-existing-apps.md
+++ b/docs/integration-with-existing-apps.md
@@ -55,14 +55,14 @@ title: Integration with Existing Apps
   }
   .display-language-objc .toggler .button-objc,
   .display-language-swift .toggler .button-swift,
-  .display-language-android .toggler .button-android {
+  .display-language-java .toggler .button-java {
     background-color: #05A5D1;
     color: white;
   }
   block { display: none; }
   .display-language-objc .objc,
   .display-language-swift .swift,
-  .display-language-android .android {
+  .display-language-java .java {
     display: block;
   }
 </style>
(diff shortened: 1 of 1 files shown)


endiliey:
@kelset docs changes are reflected in the `next`/ master version. 
https://deploy-preview-479--react-native.netlify.com/docs/next/integration-with-existing-apps

When RN website cut new version (e.g: 0.57), those all changed docs will become 0.57

kelset:
🤦‍♂️

Thanks for pointing it out, I completely missed that.

kelset:
Hey @CodingItWrong following our conversation, I've tried to run a local test - creating a Kotlin project and then following the steps in the guide to do the integration.
But, basically, I got stuck quite early and couldn't find a way to proceed; I feel like this overall section of Android integration needs a better wording and updated sections (ex. removing the `compile` in favour of `implementation` etc).

I also tried to run [your test project](https://github.com/CodingItWrong/RNKotlinIntegration) but couldn't get past this [very common StackOverflow issue](https://stackoverflow.com/questions/34353220/android-studio-please-select-android-sdk) - and since you also mentioned that
> They’re set up identically, and Java works, but Kotlin errors out on “Unable to load script from assets 'http://index.android .bundle'. Make sure your bundle is packaged correctly or you're running a packager server.”

Can I propose that we revisit this PR after making making this section up-to-date and more beginner-friendly? (I know there are already a few PRs in this repo that work in this direction)

cpojer:
Closing based on @kelset's comment. We should absolutely have some form of this documentation, though.

PR #480, opened 2018-07-30
Make the comment easier to read

tumanov-alex:
By using the comment symbol you can not only separate the value pairs remarkably, but also simplify them by removing extra comas and using well known comment symbols which can't be used in any way other than to separate stuff.

UX texts, baby!

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -158,7 +158,7 @@ For example, you may want to think about your `Animated.Value` as going from 0 t
     transform: [{
       translateY: this.state.fadeAnim.interpolate({
         inputRange: [0, 1],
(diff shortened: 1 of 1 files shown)


kelset:
I think that in this scenario adding those extra `//` is actually incorrect, since it (to me) would mean `//comment A // comment B // comment C` while the comment is instead just trying to say how the output value will change based on the entry value.

charpeni:
I agree with kelset, the comment represents the interpolation result.

Would you have a better representation? I'd think of an object, but this could be biased 🤔.

cpojer:
Seems like we don't agree that this is the right change to make so I'm going to close this PR.

PR #488, opened 2018-08-01
Update gradle version in example to match RN master

brunobar79:
The current gradle version doesn't match the version in the example.
If anyone wants to build React Native Android from source won't be able to find that string because of the version mismatch.

This PR sets the gradle version to match the latest RN version (0.56)


kelset:
Nice catch, thanks! 👏

PR #489, opened 2018-08-02
Update Android setup guide for API level 26

fiznool:
This PR updates the Android installation instructions to reference API level 26, as is now the default for RN 0.56.


morenoh149:
resolve conflicts? the docs do need to be updated to install api 26 at least.

fiznool:
Looks like #492 was merged instead so this is no longer relevant. Wish I hadn't spent so long on this now ☹️

morenoh149:
Don't feel bad. There's a lot of noise here and things get lost all the time. Thanks for your effort.

PR #490, opened 2018-08-02
Update performance docs

rafeca:
This PR changes the information about how to get the inline requires blacklist from the Metro `verboseNames`.

After https://github.com/facebook/metro/issues/204, the `verboseNames` in Metro contain already relative paths, so the process has got way simpler now

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -446,8 +446,9 @@ const config = {
   getTransformOptions: () => {
     const moduleMap = {};
     modulePaths.forEach(path => {
+      path = resolve(ROOT_FOLDER, path);
       if (fs.existsSync(path)) {
-        moduleMap[resolve(ROOT_FOLDER, path)] = true;
+        moduleMap[path] = true;
       }
     });
     return {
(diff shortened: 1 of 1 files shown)


docs/performance.md
```diff
@@ -496,12 +439,15 @@ const modulePaths = require('./modulePaths');
...
+// Update the following line if the root folder of your app is somewhere else.
+const ROOT_FOLDER = path.resolve(__dirname, '..');
+
 const config = {
   getTransformOptions: () => {
     const moduleMap = {};
     modulePaths.forEach(path => {
       if (fs.existsSync(path)) {
-        moduleMap[resolve(path)] = true;
+        moduleMap[resolve(ROOT_FOLDER, path)] = true;
```
jamesreggio: I think you'll want to resolve the absolute path and pass it into `existsSync` above. Otherwise, LGTM!

rafeca: oh yeah that's right 😆

kelset:
Awesome, thanks @rafeca !

PR #493, opened 2018-08-06
Renamed show/hide to toggle inspector

feargswalsh92:
I spotted this while trying to figure out how to change the port to something other than 8097 for the inspector. There isn't an issue related to it reported as far as I can tell. If you know how to do what I am trying to do please let me know 🙂


charpeni:
I didn't notice before, but we lost the chrome inspector and the node selector in the new gif.

I'll revert the gif as I prefer to have some wrong naming than a missing feature. Are you up to send a new PR?

feargswalsh92:
Yep I can do both again. It'll be tomorrow at the earliest though.

feargswalsh92:
Do you want me to make new gifs for both instances or just the one that explicitly mentioned Zonder?

PR #494, opened 2018-08-06
Renamed another instance of show/hide 

feargswalsh92:
@charpeni  This is the same idea.


charpeni:
I'd prefer if we don't promote an app (Zonder) in the documentation.

feargswalsh92:
Sure no problem I understand I'll upload a more generic gif

feargswalsh92:
#511 combines both of them.

PR #533, opened 2018-08-28
Fix Typo

abdullahtariq1171:
Change `require`d to `required`

See https://facebook.github.io/react-native/docs/appregistry


charpeni:
It looks like `require` refer to the `require` keyword.

So it makes sense to have `require`d.

abdullahtariq1171:
Not sure, maybe.
Actually, I think might be confusing to the new reader. I never came across such thing (`require`d).

Anyway feel free to close if you are confident.

abdullahtariq1171:
Sounds good. 
Updating in a few minutes

PR #519, opened 2018-08-22
Add "known issues" section to InputAccessoryView

Ashoat:
This PR adds a "known issues" section to `InputAccessoryView`.

PR #541, opened 2018-09-01
Add a new page documenting React Native's out-of-tree platform support

empyrical:
This PR adds a new page to the React Native docs that showcases other platforms that have been developed for React Native, and a bit documenting the API change that was added in https://github.com/facebook/react-native/pull/20825 . In the future, this should be able to be expanded with more information on making/maintaining React Native platforms.

### DIFF
--- a/docs/out-of-tree-platforms.md
+++ b/docs/out-of-tree-platforms.md
@@ -1,0 +1,38 @@
+---
+id: out-of-tree-platforms
+title: Out-of-Tree Platforms
+---
+
+React Native is not just for Android and iOS - there are community-supported projects that bring it to other platforms, such as:
+
+ * [React Native Windows](https://github.com/Microsoft/react-native-windows) - React Native support for Microsoft's Universal Windows Platform (UWP) and the Windows Presentation Foundation (WPF)
+ * [React Native DOM](https://github.com/vincentriemer/react-native-dom) - An experimental, comprehensive port of React Native to the web. (Not to be confused with [React Native Web](https://github.com/necolas/react-native-web), which has different goals)
+ * [React Native Desktop](https://github.com/status-im/react-native-desktop) - A project aiming to bring React Native to the Desktop with Qt's QML. A fork of [React Native Ubuntu](https://github.com/CanonicalLtd/react-native/), which is no longer maintained.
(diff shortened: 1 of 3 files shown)


empyrical:
I almost wonder if I should make a PR that throws an error if you pass a `--platform` that doesn't exist. Also Haul would be a good thing to mention, yes. But IIRC it only supports iOS/Android - a PR to remedy that could probably be made too!

docs/out-of-tree-platforms.md
```diff
@@ -0,0 +1,38 @@
...
+
+ * [React Native Windows](https://github.com/Microsoft/react-native-windows) - React Native support for Microsoft's Universal Windows Platform (UWP) and the Windows Presentation Foundation (WPF)
+ * [React Native DOM](https://github.com/vincentriemer/react-native-dom) - An experimental, comprehensive port of React Native to the web. (Not to be confused with [React Native Web](https://github.com/necolas/react-native-web), which has different goals)
+ * [React Native Desktop](https://github.com/status-im/react-native-desktop) - A project aiming to bring React Native to the Desktop with Qt's QML. A fork of [React Native Ubuntu](https://github.com/CanonicalLtd/react-native/), which is no longer maintained.
+ * [React Native macOS](https://github.com/ptmt/react-native-macos) - An experimental React Native fork targeting macOS and Cocoa
+
+## Creating your own React Native platform
+
+Right now the process of creating a React Native platform from scratch is not very well documented - one of the goals of the upcoming re-arcitecture ([Fabric](https://facebook.github.io/react-native/blog/2018/06/14/state-of-react-native-2018)) is to make maintaining a platform easier.
```
kelset: `arcitecture` typo

kelset:
related to 
> It would be great to include the first couple of error messages (e.g. when Accessibility.platform.js is missing) and how to fix them.

We have been discussing about adding a "known issue" section to each component/page - you can check out how it would look here -> https://github.com/facebook/react-native-website/pull/519

So I think that if you want to add it you should follow that "approach"

docs/out-of-tree-platforms.md
empyrical: Whoops, fixed!

PR #548, opened 2018-09-07
add a line regarding giff&webP on android in Update image.md 

ajulien42:
android app doesn't work without it


charpeni:
Fixed in 3eb585d66bbed6b9adee0008f84b1b8f5a74a283.

PR #557, opened 2018-09-13
fix DatePickerIOS example

mgcrea:
Using an arrow function prevents `this.setState is not a function` error with the current example (missing bind).


charpeni:
I'll close this PR as the method `setDate` was already bound in the constructor, and the review is still unaddressed. Feel free to resubmit this PR if you think it should be the case.

PR #569, opened 2018-09-19
update a dead link

esthor:
Updating the link to the Metro CLI config file


vreality64:
👍 but it's exactly same with #567

charpeni:
Fixed in #567.

PR #572, opened 2018-09-21
Update switch.md

javiercr:
It seems for some reason `thumbTintColor` has been renamed to `_thumbColor` according to this warning:

```
// react-native/Libraries/Components/Switch/Switch.js:112
'Switch: `thumbTintColor` is deprecated, use `_thumbColor` instead.',
```

### DIFF
--- a/docs/switch.md
+++ b/docs/switch.md
@@ -18,7 +18,7 @@ This is a controlled component that requires an `onValueChange` callback that up
 * [`ios_backgroundColor`](switch.md#ios_backgroundColor)
 * [`onValueChange`](switch.md#onvaluechange)
 * [`testID`](switch.md#testid)
-* [`_thumbColor`](switch.md#_thumbColor)
+* [`thumbColor`](switch.md#thumbColor)
 * [`tintColor`](switch.md#tintcolor)
 * [`value`](switch.md#value)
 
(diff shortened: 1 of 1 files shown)


hramos:
`thumbTintColor` has been deprecated. Generally, props with a _ prefix are not considered part of the public API.

Instead of renaming this prop, can you simply add a deprecation warning to the docs?

javiercr:
Sorry I don't quite get it. Is there a reason why this has been deprecated? It still seems [a valid property for iOS switches](https://developer.apple.com/documentation/uikit/uiswitch/1623684-thumbtintcolor?language=objc). Does this mean in the future we won't be able to control this property from RN?

Is there any example in the docs of how to document a property that has been deprecated?

Thanks!

hramos:
You may usually find additional context behind a given change by looking up the commit that implemented it. Looking through the [history of Switch.js](https://github.com/facebook/react-native/commits/master/Libraries/Components/Switch/Switch.js), I found [965adee109b86cc36f39a713237ade9592f13dad](https://github.com/facebook/react-native/commit/965adee109b86cc36f39a713237ade9592f13dad#diff-6ee4e62c750fcd87064f152f53214a05) which does say that, in fact, `thumbTintColor` has been renamed to `thumbColor`:

```
- Renamed `thumbTintColor` to `thumbColor`.
```

So I was wrong in claiming `thumbTintColor` had been deprecated.  I don't know why the message you saw used `_thumbColor`, as it should have said `thumbColor`. We should fix that.

I'm OK with renaming the prop in the docs here. I'll edit your change to use `thumbColor` and get this merged. Thanks for pointing this out and helping us keep the docs up to date!

PR #575, opened 2018-09-24
Documentation update for Switch "thumbTintColor" to "thumbColor"

HerrX2000:
"thumbTintColor" is deprecated so I would recommend changing the documentation to "thumbColor".


stianjensen:
The same change has now been made and merged in #572 
https://github.com/facebook/react-native-website/pull/620 backports this to 0.57 docs

PR #578, opened 2018-09-25
fix contextMenuHidden link

bosung90:
fix contextMenuHidden link


hramos:
Do we need to apply the same fix to the docs/ variant?

bosung90:
yes, I was looking for master doc but couldn't find one

hramos:
It's in the `docs/` folder. I can see that version has the same mistake: https://github.com/facebook/react-native-website/blob/master/docs/textinput.md

hramos:
Alright, fixed.

PR #581, opened 2018-09-27
File structure should include the 1x version too.

watadarkstar:
Its unclear in the docs that the 1x version should be included too when providing @2x and @3x images. I believe that you do need it, correct me if I'm wrong?

File structure should be:

```
└── img
    ├── check.png
    ├── check@2x.png	    
    └── check@3x.png	    
```

not

```
└── img
    ├── check@2x.png	    
    └── check@3x.png	    
```


watadarkstar:
Thanks for the merge @hramos 😄

PR #589, opened 2018-10-03
Text component behaviour inside another Text component

vikasverma92:


charpeni:
Actually, `| |` are referring to the screen's borders. 

The examples were right but could be misleading, I've rewritten the examples to add a full width and a flow one. What do you think?

vikasverma92:
> Actually, `| |` are referring to the screen's borders.
> 
> The examples were right but could be misleading, I've rewritten the examples to add a full width and a flow one. What do you think?

@charpeni That makes sense. Thanks for the change.

PR #582, opened 2018-09-28
Update Getting Started to use Expo CLI

fson:
This PR implements the proposal "Use Expo CLI in the Quick Start section of the docs":
https://github.com/react-native-community/discussions-and-proposals/issues/23

* Updates Getting Started with steps to get started using Expo CLI and up-to-date instructions for getting support.
* Updates CRNA references in the docs so that they also make sense for projects created with Expo CLI.

PR #593, opened 2018-10-04
Update Getting Started in v0.57 (latest) docs

fson:
Add changes from #582 to the latest (v0.57) docs.


fson:
@hramos I noticed the v0.57 docs were still not updated

PR #597, opened 2018-10-05
Add docs for ImageBackground

filiptdz:
The `ImageBackground` component is currently hidden inside the Images docs. Since many people only skim through the sidebar when looking for a solution, that most likely will pass unnoticed. Besides, with a page of its own, the docs now include the prop reference for ImageBackground, no longer requiring that the developer reads through the source code to find the available props.


hramos:
This is so good. Thanks!

PR #600, opened 2018-10-06
Lighter big button color change on hover

joerez:
Up to preference, but this may look a bit nicer compared to the old hover effect for the big buttons in the hero section.


charpeni:
@hramos thoughts?

PR #603, opened 2018-10-08
.big-button nicer hover effect

joerez:
I added a slight transform on the .big-button hover effect that does:

```
transform: scale(1.05)
```

and I added a transition to the .big-button class that transitions this change as well.


charpeni:
I like that! @hramos thoughts?

PR #610, opened 2018-10-10
Unnecessary linking removed from getting started page

ravirupareliya:
There was some unnecessary linking text for whole paragraph, which was not needed.


charpeni:
Fixed in b159246.

PR #611, opened 2018-10-10
Improper closing of anchor tag.

iamraviprakash:
Improper closing of <a> or anchor tag at 158 line.


charpeni:
Fixed in b159246.

Issue #617, opened 2018-10-13
Broken Anchor Tag - Getting Started Docs


jigarakatidus:
`<a href="getting-started.html" onclick="displayTab('guide', 'quickstart')">Expo/a>`

Notice the incorrect closing anchor tag. Complete paragraph is hyperlinked due to this.

charpeni:
Fixed in b159246.

PR #618, opened 2018-10-13
minor tag fixed in docs/getting-started

jigarakatidus:
This will fix: #617


jigarakatidus:
My bad, just noticed that there are already many PR which fixes this.

charpeni:
Fixed in b159246.

PR #628, opened 2018-10-15
Update getting-started.md

omargarde:
Missing an less-than symbol for the closing anchor tag on line 158.


charpeni:
Fixed in b159246.

PR #629, opened 2018-10-17
Removed performance section from stylesheet as no longer true

danilobuerger:
See https://github.com/facebook/react-native/commit/a8e3c7f5780516eb0297830632862484ad032c10
`StyleSheet.create` is now an identity function, the `ReactNativePropRegistry` (`StyleSheetRegistry`) was removed.


brunolemos:
Now that it's an identity function, it's not clear the value the StyleSheet.create method provides, if any.

image

PR #646, opened 2018-10-29
Modified webview.md for `injectJavaScript`

ifsnow:
I think `injectJavaScript` is a method, not a props.


docs/webview.md
```diff
@@ -541,3 +531,11 @@ stopLoading();
...
 
 Stop loading the current page.
+
+### `injectJavaScript(str)`
+
+```javascript
+injectJavaScript("... javascript string ...");
+```
+
+Executes the JavaScript string.
```
levibuzolic:
```suggestion
Function that accepts a string that will be passed to the WebView and executed immediately as JavaScript.
```

cpojer:
We are removing Webview from React Native. It is now part of a separate repo: https://github.com/react-native-community/react-native-webview so please send your change to that one instead.

PR #648, opened 2018-10-29
Fix broken link in understanding-cli.md

teameh:
[@orta commented on 9 Aug: ](https://github.com/facebook/react-native-website/pull/462#issuecomment-411766140)

> Looks like it got pulled out into it's own repo - the right link is probably https://github.com/facebook/metro/blob/master/packages/metro-config/src/configTypes.flow.js - want to update that?

Fixed in this commit


teameh:
See conversation at https://github.com/facebook/react-native-website/pull/462

PR #655, opened 2018-10-31
Accessibility Escape

rustle:
Documentation updates related to https://github.com/facebook/react-native/pull/22047


docs/view.md
```diff
@@ -181,6 +182,16 @@ When `accessible` is `true`, the system will invoke this function when the user
 
 ---
 
+### `onAccessibilityEscape`
+
+When `accessible` is `true`, the system will invoke this function when the user performs the escape gesture.
+
+| Type     | Required |
```
elucaswork: A `Platform` column is also needed here, right? with an `iOS` for its value, same as [accessibilitytraits](https://facebook.github.io/react-native/docs/view#accessibilitytraits)

PR #656, opened 2018-10-31
Update guides with latest content across all versions

hramos:
Files listed in `website/versionsBlacklistDocs.json` should remain consistent across all versions. I've updated the `blacklist-check.js` script to automatically fix these docs.

This PR includes both the updates to the script, as well as the result of running the script. In a future PR, we may automatically run this script as part of the pre-commit checks. This will solve a lot of the confusion around which files to update (answer: always update "docs/" unless you know what you're doing).


charpeni:
Amazing @hramos! 🎉

PR #674, opened 2018-11-10
I think that Swift is more relevant for implied iOS development

yassineim:


cpojer:
Thanks! I made a similar change in this PR: https://github.com/facebook/react-native-website/pull/730 which is a bit bigger so I hope you don't mind if I'm closing your PR.

yassineim:
Yup no problem.

Thanks!

PR #677, opened 2018-11-12
Update Docs - Grammar Check

sirjmkitavi:
Updating a few grammar errors in the endeavour to make this docs more epic than they already are. 😎


charpeni:
"All your base are belong to us" is actually a reference to a popular meme: https://en.wikipedia.org/wiki/All_your_base_are_belong_to_us.

sirjmkitavi:
Thank You @charpeni

PR #688, opened 2018-11-29
Add Stremio to the showcase

NikolaBorislavovHristov:
The android version of Stremio is written mostly in react native. It combines streaming server, libVLC bindings and chromecast support.
Can you add it to showcase?


charpeni:
Congrats with your app! Using React Native with libVLC bindings and chromecast support sounds really interesting. I'd love to read a blog post about your experience.

PR #693, opened 2018-12-03
Update layoutanimation.md

amitavroy:
Modified the documentation of LayoutAnimation. It is good to mention that we can set the flags inside the constructor of the component. Makes it a little more easy to figure out.


amitavroy:
While doing the development I felt it will be a little more easy to show the code of setting up the flag inside the constructor function.

That's why I modified the line.

cpojer:
Thanks for your PR. Unfortunately I don't think it makes the documentation more clear so I'm going to close this PR. I'm happy to hear suggestions on how we can make this more clear, possibly with an improved code example?

PR #694, opened 2018-12-05
Add uphold to RN showcase page

SandroMachado:
The v2 uphold mobile apps are built using React Native. This PR adds the apps to the showcase page.


charpeni:
Congrats with your app!

PR #705, opened 2018-12-12
Create Updated blog post on how to use TypeScript

janhesters:
TypeScript is getting ever [more popular](https://medium.com/npm-inc/this-year-in-javascript-2018-in-review-and-npms-predictions-for-2019-3a3d7e5298ef). Which is why I took the time to updated and enhance the existing blog post about using React Native with TypeScript. This update is more comprehensive that the few lines that have been added to the old post a couple of days ago.

In addition to getting TypeScript and Jest to work, it also covers TSLint, Enzyme and React Native Testing Library.

### DIFF
--- a/website/blog/2018-12-14-using-typescript-with-react-native.md
+++ b/website/blog/2018-12-14-using-typescript-with-react-native.md
@@ -154,7 +154,7 @@ Tada 🎉. It works.
 
 ---
 
-**Note:** Most people either use Enzyme or React Native Testing Library. So you probably only want to follow one of 4. a) and 4. b).
+**Note for the following section:** Most people use either Enzyme or React Native Testing Library. So you probably only want to follow one of 4. a) and 4. b).
 
 ## 4. a) Setting Up Enzyme
 
--- a/website/blog/2018-12-14-using-typescript-with-react-native.md
+++ b/website/blog/2018-12-14-using-typescript-with-react-native.md
@@ -16,7 +16,7 @@ In this tutorial, you are going to learn how to set up a React Native project wi
 
 An older [blog post](http://facebook.github.io/react-native/blog/2018/05/07/using-typescript-with-react-native) on the React Native site describes how to set up React Native with TypeScript. It's a little outdated, though.
 
-Things have changed since then. Since version 0.57 the framework ships with TypeScript support. And there is an [excellent template](https://github.com/emin93/react-native-template-typescript) for RN with TypeScript. This post is an update and an enhancement for the instructions of that old article. In addition to [Jest](https://jestjs.io/), we will also go into detail on how to set up linting, [Enzyme](https://airbnb.io/enzyme/) and [React Native Testing Library](https://www.npmjs.com/package/react-native-testing-library). Let's get started.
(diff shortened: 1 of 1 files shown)


kelset:
I feel that this should not be a blogpost, but it could be reworked to fit into the docs - as we are seeing more and more people wanting to use TS with RN.

cc @LinusU @ashfurrow would you mind checking it and LMK what you think about the content?

ashfurrow:
Hmm, this is really tricky. The content of the blog post is really great (well done janhesters!) but I share @kelset’s concerns. However, I’m not sure this would necessarily fit as-is inside docs; it really is a tutorial that covers an expansive setup, much more than _just_ RN and TS.

I think this would be a really helpful resource for learners, but I also see the benefit of having docs instead, that can be updated over time. In six months, will someone else need to rewrite this post again? Maybe, but even if they did, that’s not so bad. The benefits of having this available to learn from outweigh the concerns of docs-vs-blogs, imo. But that’s just me, I’ll leave it up to the React Native team 👍

LinusU:
Haven't had time to read it all yet, just skimmed the first sections, will read all soon.

One thing I thought about was the Jest setup, I didn't realize that it doesn't work out of the box with TypeScript, as the normal TypeScript support does.

Could we fix this in the `react-native` Jest preset? I would be happy to submit PRs for this.

There is also a `tsconfig.jest.json` referenced in the config, but no more info about that file. What does it do? 🤔

janhesters:
@LinusU Whoopsie, forgot to add that part. I updated it! It just extends the normal tsconfig for ts-jest so that Jest gets common JS code (and doesn't complain e.g. about `import`. See [here](https://kulshekhar.github.io/ts-jest/user/config/).

Also the fact that Jest and linting don't just work out of the box was why I wanted to create this blog post and preferably share it directly on the official website. These just the common stap (which I had archived for myself in a private Slack conversation) which make setting up new RN projects with TypeScript - and remembering all steps - easy 👍.

janhesters:
@kelset @LinusU @ashfurrow So guys, what is the consensus? 🤗

kelset:
So, are the overall feedback I got from what written above by Ash + Linus + other people in the core that commented on the Discord is that this should not be a blogpost on the main website, hence two roads:

- this gets reworked into a Documentation page, which means more or less:
	- removing the references to the other blogpost
	- define clearly why you are installing the other packages (ex: enzyme, react-native-testing)
	- provide a minimal github repo
	- add a link to the older blogpost referencing the new page of the docs
- you publish it on your own platform of choice

As stated elsewhere, I'd really prefer to have a dedicated section in the docs for this - a conversation about having a "proper" Tutorial section also came out in another occasion last week. But it's your decision 😊

All of this said, from FB I also received a feedback that they would comment on this PR after they made up their mind about it. So don't take my feedback as "final", as they *may* want to chime in with a different take.

janhesters:
@kelset Thank you for answering, I was worried this PR would be dead now 😅

Okay, I'll wait for the Facebook's teams feedback.

If they agree with you, I'll happily create a PR as I don't care about the clout and I merely like to have all information in one source. Furthermore, I don't care about the format as long as all (= setup + linting + testing) is in one place (which might be better as a blog 🤔), since these are the most common steps for each TS project 🚀

janhesters:
@rickhanlonii Thank you for your answer. Okay, I'll do it and then take care of the PR as soon as I can (I'm currently facing [RSI](https://de.wikipedia.org/wiki/Repetitive-Strain-Injury-Syndrom) though 😣)

How can I get ReactNativeComm to retweet? 😊

charpeni:
To retweet, you can reach us here with a link.

cpojer:
Thank you for your hard work on this. We tweeted it on the React Native twitter account: https://twitter.com/reactnative/status/1093515749493227521

PR #707, opened 2018-12-14
Remove absoluteFillObject

MiLeung:
It literally points to the same reference as absoluteFill.

https://stackoverflow.com/questions/53676723/react-native-what-is-the-difference-between-stylesheet-absolutefill-and-styl/53775303#53775303


cpojer:
Thank you!

PR #708, opened 2018-12-14
Document the `progressViewOffset` property

jaredly:
had to dig through the source code for a while to find it 🙃


levibuzolic:
Might be better to give this prop its own section, adding it to the list of props at the top and specifying that it's android only.

levibuzolic:
Actually looks like it already exists on [`RefreshControl`](https://facebook.github.io/react-native/docs/refreshcontrol#progressviewoffset)

cpojer:
I'm fine adding this small note, thanks for the PR!

PR #709, opened 2018-12-15
Document Animated.loop config in the same style as other functions

levibuzolic:
The `Animated.loop` documentation didn't explicitly show it takes an optional `config` object as a second argument nor did it follow the same format as other functions on the same page.

Also went back and updated the docs from `0.44` onwards. That's as far back as the docs mention the `loop` function from what I could see.


cpojer:
Thank you! Thanks for updating the old docs too, I'm sure that is helpful to some people :)

PR #714, opened 2018-12-17
Text: Indicate that `letterSpacing` works on Android

rigdern:
The docs say that `letterSpacing` is iOS-only but it also works on Android.

Why do we document the `Text` styles in both `text.md` and `text-style-props.md`?

Adam Comella
Microsoft Corp.


cpojer:
Cool, thanks for the update! I guess it didn't work back when this was initially written.

PR #718, opened 2018-12-21
Updated packager config to include project dir

dotconnor:
The initial packager config doesn't have the project directory in the config, and running it without it causes the packager look for files in the packager folder, not the intended project folder.


cpojer:
Actually, instead of suggesting this change, how about changing the overall documentation to stop recommending people to create a `packager` folder? Let's just recommend people to create a `metro.config.js` file in the root of their project. That won't require additional config, and it won't ask people to create an unnecessary folder.

cpojer:
Closing this PR because it is stale.

PR #721, opened 2018-12-24
Update images.md

yongching:
I find this `asset:/ scheme` a little bit misleading and it doesn't work, until I see this https://github.com/facebook/react-native/issues/17782. Correct me if i'm wrong...


cpojer:
Yeah, this doesn't look right.

PR #722, opened 2018-12-26
confusing use of "props"

TrentonAdams:
If I'm not mistaken, the use of the word "prop" here implies it's a reactive component property, but that's incorrect terminology is it not?


TrentonAdams:
Nope, I'm wrong.

PR #724, opened 2018-12-29
Add no support for numberOfLines on nested Text

mloureiro:
This feature is difficult to implement at the native level.

Original issue: https://github.com/facebook/react-native/issues/22811


cpojer:
Yeah, I think it's worth mentioning this.

PR #726, opened 2019-01-01
Update getting started for AMD users.

cptiwari20:
As I faced problem in setting up the react-native environment in my windows machine with AMD processor. It is hard to understand where the problem is. Most of the time we guide on the basis of Operating system but sometimes all the things are not working same in all the processors. Here 'android-studio' is not working correctly. The android emulator is not able to open as there is no intel HAXM. This is simple update for the AMD users.


cpojer:
Thank you! I made a small modification so that there is less text in the Getting Started section, which should be kept simple.

cptiwari20:
:)

PR #727, opened 2019-01-03
Updated docs for legacyImplementation property

LuckyLuky:
Updated docs for `legacyImplementation` property of `FlatList` and `SectionList`.

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -401,6 +401,8 @@ Set this when offset is needed for the loading indicator to show correctly.
 
 ### `legacyImplementation`
 
+Uses legacy `MetroListView` instead of default `VirtualizedSectionList`.
+
 May not have full feature parity and is meant for debugging and performance comparison.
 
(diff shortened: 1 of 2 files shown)


cpojer:
Can you explain what `MetroListView` is and where this is coming from?

LuckyLuky:
I could explan `MetroListView`, which I've found in the source code of version `0.57.4`.

But I wonder if this pull request is even valid, since you're removing the `legacyImplementation` completely, as I've just noticed in [this commit](https://github.com/facebook/react-native/commit/636d01bbd0b9a2e9617f8063713183a0a3a4d5fe#diff-85e8f96a713d23be94fbd08687332bff). 

If you think it is, I could provide more information. Docs in the source code say this: 

> This is just a wrapper around the legacy ListView that matches the new API of FlatList, but with some section support tacked on

charpeni:
Let's closed this PR as Eli removed the legacy implementation.

PR #728, opened 2019-01-04
Lower cased “versions”

mattsidjohn:
Changed it to be consistent with the rest of the titles with “versions” in it.


cpojer:
Thank you!

PR #730, opened 2019-01-04
More modest introduction to React Native.

cpojer:
I am planning a more thorough overhaul of the home page but I want to get these small changes in for now, even if they won't stay around for too long. After reading through the React Native website I got the feeling that the React Native project is somewhat pretentious. I'm not trying to blame anyone, I believe this accumulated over years and 2019 is quite different from 2015 – I would have probably written things in a similar style back then. I would like to thoroughly change the website to give a more modest and mature impression of React Native. Instead of touting how amazing React Native can be, I believe we should aim to be highly informative and simplistic in our language instead of joking around. Let the technology speak for itself and give people that wow effect.

The particular changes in this PR are:
* Changed the first component name from `WhyReactNativeIsSoGreat` to `HelloReactNative`.
* Slightly reworded the first component's example text to be shorter.
* Changed the text about React Native being a "real app" to remove the quotes around "mobile web app". I get what the paragraph is trying to say, but it sounds like mobile web apps are inherently bad. We don't need to imply anything like that.
* Removed `Awkward` from one component example.
* Changed the example image from a silly animated gif to the React logo (boo, I'm boring, sue me).
* Renamed `SomethingFast` and `TheGreatestComponentInTheWorld` to a possible real world example.
* Changed the formatting to be in line with the project's source code (no spaces around curly braces).

Again, there is a ton more to be done here, I just wanna put some patches in place before overhauling everything in a bigger change.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -13,9 +13,7 @@ const Container = CompLibrary.Container;
 
 const siteConfig = require(process.cwd() + '/siteConfig.js');
 
-const pinnedApps = siteConfig.users.filter(app => {
-  return app.pinned;
-});
+const pinnedApps = siteConfig.users.filter(app => app.pinned);
 
 class Button extends React.Component {
   render() {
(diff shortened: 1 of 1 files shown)


alextes:
I like it!

and yes, booo boring! (😉)

PR #738, opened 2019-01-14
Fix typo

jacob-israel-turner:

### DIFF
--- a/docs/linking-libraries-ios.md
+++ b/docs/linking-libraries-ios.md
@@ -41,7 +41,7 @@ Done! All libraries with native dependencies should be successfully linked to yo
 
 #### Step 1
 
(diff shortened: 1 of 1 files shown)


docs/linking-libraries-ios.md
```diff
@@ -41,7 +41,7 @@ Done! All libraries with native dependencies should be successfully linked to yo
 
 #### Step 1
 
-If the library has native code, there must be a `.xcodeproj` file inside it's folder. Drag this file to your project on Xcode (usually under the `Libraries` group on Xcode);
+If the library has native code, there must be a `.xcodeproj` file inside its folder. Drag this file to your project on Xcode (usually under the `Libraries` group on Xcode);
```
rickhanlonii: Want to change "a xcodeproj" to "an xcodeproj" while we're here?

```suggestion
If the library has native code, there must be an `.xcodeproj` file inside its folder. Drag this file to your project on Xcode (usually under the `Libraries` group on Xcode);
```

PR #740, opened 2019-01-15
Android: difference between screen and window

sidferreira:
Explains the difference between screen and window for Android


charpeni:
Related information on Stack Overflow: https://stackoverflow.com/a/44979327

PR #745, opened 2019-01-19
changing `column` to `row`

felipeamarante:
The default value according to the css docs is `row` not `column`
I testes myself and it's `row`

https://www.w3schools.com/cssref/css3_pr_flex-direction.asp

### DIFF
--- a/docs/flexbox.md
+++ b/docs/flexbox.md
@@ -11,7 +11,7 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
 
 #### Flex Direction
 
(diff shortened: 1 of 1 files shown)


charpeni:
React Native uses [Yoga](https://yogalayout.com/) for his flexible layouts, and `flexDirection` default value is `column`.

https://github.com/facebook/react-native/blob/6b0512c8196da5356272cbdfc1f16f23891d4f9e/ReactAndroid/src/main/java/com/facebook/react/uimanager/LayoutShadowNode.java#L262-L265

PR #754, opened 2019-01-25
Update alertios.md

EricaThompson:


charpeni:
Thanks for proof-reading the documentation.

This is mostly intentional, "All your base are belong to us" is actually a reference to a popular meme: https://en.wikipedia.org/wiki/All_your_base_are_belong_to_us.

PR #756, opened 2019-01-26
Bump docs to v0.58

hramos:
This is the result of running the following commands:

```
cd website
docusaurus-version 0.58
yarn sync-guides
```

I've also removed a link to the old discussion forum (now shut down).

Test Plan:

Ran `yarn start` and verified Docusaurus built the site successfully.


hramos:
Strange, I didn't have any issues locally.

PR #757, opened 2019-01-26
Reduce versions to build on Netlify

charpeni:
During the build process, we build a complete website for each version which is 9.2 MB / version.

On Netlify, our deploy preview infrastructure, we have to upload every changed file. Each time we cut a new version or making multiple changes we're facing timeout errors from Netlify trying to upload megabytes of HTML files.

This script makes sure we're not deploying unnecessary versions on Netlify.

We only need 0.5 which is the first draft of each doc, and the recent ones.

🐎 Also, this would significantly decrease the time needed to build a deploy preview.

⚠️ `imagebackground` and `safeviewarea` were missing in `versioned_docs/version-0.5`.


hramos:
ImageBackground and SafeAreaView were added later on. We probably don't want them displayed in the docs under version 0.5.

cpojer:
@charpeni mind addressing the feedback from @hramos?

charpeni:
I removed ImageBackground and SafeAreaView from version 0.5, and instead, I edited the script to keep version 0.46 (ImageBackground) and 0.50 (SafeAreaView).

So, from now, on Netlify we'll have version 0.5, 0.46, 0.50, and all versions from 0.56. That should save us some time on building deploy previews and should prevent the timeout issue with Netlify on cutting releases.

I'll monitor this on the next release and I'll remove the script if it didn't help as planned.

_Note: Netlify is only used for deploy previews._

Issue #773, opened 2019-02-01
Remove NavigatorIOS from the docs


hramos:
This component was removed in https://github.com/facebook/react-native/commit/0df92afc1caf96100013935d50bdde359b688658. References to it need to be removed from navigation.md, and the navigatorios.md doc will need to be updated.

I believe that simply removing navigatorios.md from versioned_docs/0.58 and docs/ may not work, as Docusaurus might have trouble handling the removal. We may need to add a notice instead, explaining the component has been removed from core.

gedeagas:
PR Submitted https://github.com/facebook/react-native-website/pull/776. 
what do you think about the wording @hramos?

PR #776, opened 2019-02-04
Deprecating NavigatorIOS

gedeagas:
Fix https://github.com/facebook/react-native-website/issues/773


docs/navigatorios.md
```diff
@@ -3,6 +3,8 @@ id: navigatorios
 title: NavigatorIOS
```
TheSavior: Should we just delete this file?

hramos: I thought this would cause issues with Docusaurus, but after testing it seems fine, as long as the entry is removed from sidebars.json.

I'll pull in this pull request to get the changes to docs/navigation.md in, and I'll remove navigatorios.md in a separate PR.

PR #777, opened 2019-02-05
add command to start server for running app section

buncismamen:
add step to run react-native server before installing the app 

idk but there is no step/command to run the server, it could confuse some people especially if they jump straight to react-native environment


buncismamen:
idk seems my pull request is useless

I used to run the server manually `yarn start` after running `react-native run-android` with real device
because I always get can't connect error after opening the installed apk

but today I run `react-native run-android` with emulator as my target then I not need to run `yarn start` manually

charpeni:
Let's close this then, thanks for proof-reading the documentation.

PR #790, opened 2019-02-14
Add ecbo cloak to the showcase

tsukushi0104:
Hi team!
Could you please add `ecbo cloak` to the showcase?
Thanks!


charpeni:

PR #823, opened 2019-03-08
Update Share.md

aboobakkar:
Worked Code Sample.


rickhanlonii:
Hey @aboobakkar, thanks for the update - what's the motivation for this change? Can you include a screenshot/

aboobakkar:
Previous code was not working. So debug current code and found real code.

cpojer:
Thanks for this PR! We are actually extracting this code into a separate repository so we won't be changing the docs here any more.

PR #830, opened 2019-03-12
Update `in` code to be consistent with `out` and `inOut`

grifotv:


charpeni:
Hmm, I'm sure I already merged a PR to fix this.

Something must be wrong with Prettier, can you try to run Prettier on your branch to see if it will be reverted again?

PR #837, opened 2019-03-17
Update state.md

blackwolf08:
ES6 and new react version does not need super or constructor


blackwolf08:
@charpeni looks good, will look towards implementing that

blackwolf08:
@charpeni Are we good now?

PR #839, opened 2019-03-19
Update Linking adding the brand new openSettings method

elucaswork:
Add documentation for `Linking.openSettings()`

https://github.com/facebook/react-native/pull/23965


elucaswork:
thanks @charpeni

PR #843, opened 2019-03-19
fix: update CLI paths

thymikee:
## Summary

Update the CLI paths to `yarn react-native`, because that's how we're supposed to run it.

I also added a `performance.md` to versioned 0.59 docs, because the sooner we make people to start official APIs, the better.


thymikee:
Thanks! Looks like removing the versioned version was easier, so I went that path :D

PR #853, opened 2019-03-26
Remove BackHandler with using NativeEventSubscription

asukiaaa:
I failed to remove `BackHandler` with using document style.
I succeeded in removing with keeping `BackHandler` as a value of `NativeEventSubscription`.
Thank you.

Reference: https://stackoverflow.com/a/52797669/3774895


asukiaaa:
Thank you approving and merging.

PR #857, opened 2019-03-28
Manual Upgrade Web Tool v1

watadarkstar:
The work here is based on the web work done by @pvinis here https://pvinis.github.io/purge-web/

The is the first iteration in a series of iterations we want to do around the web upgrading process.

Why the use of `dangerouslySetInnerHTML`? Unfortunately, docusaurus does not allow us to use `setState` and only supports static pages so react doesn't work as it should. 

Screen Shot 2019-04-08 at 4 50 45 PM

### DIFF
--- a/docs/direct-manipulation.md
+++ b/docs/direct-manipulation.md
@@ -54,7 +54,7 @@ render() {
 
 This is computationally intensive compared to the original example - React needs to re-render the component hierarchy each time the opacity changes, even though other properties of the view and its children haven't changed. Usually this overhead isn't a concern but when performing continuous animations and responding to gestures, judiciously optimizing your components can improve your animations' fidelity.
 
-If you look at the implementation of `setNativeProps` in [NativeMethodsMixin](https://github.com/facebook/react-native/blob/master/Libraries/Renderer/oss/ReactNativeRenderer-prod.js) you will notice that it is a wrapper around `RCTUIManager.updateView` - this is the exact same function call that results from re-rendering - see [receiveComponent in ReactNativeBaseComponent.js](https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeBaseComponent.js).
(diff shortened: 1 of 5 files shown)


iamclaytonray:
Is the third point worth the hassle of having a never-ending stream of "can you add an upgrade guide for X lib?" I think it's a cool concept but that would put a lot of strain on whoever (presumably you) is maintaining docs, specifically, this section.

I may be reading into this too far, though. Would this strictly be non-native libraries? That's the part where I think it'd be _really_ hard to maintain

JCMais:
I'm with @iamclaytonray if that section is for non-native libraries, probably allowing only native libraries there would be better.

Maintaining that would probably not be that hard, we could have a JSON file (or the like) with a structure similar to:
```
[library-name]: {
  [version-range]: 'react-native-version-range',
}
```

alpha0010:
For the most part, I find 3rd libraries straight forward to update* (usually just increment the version number).  Challenges that would be nice to have a centralized documented procedure for are [mysterious build failures](https://github.com/wonday/react-native-pdf/issues/316) and libraries [changing default template code](https://github.com/wix/react-native-navigation/issues/4853).

\* I am relatively new to the react native world, so take that with a healthy pinch of salt.

pvinis:
what's the URL to see this?

lucasbento:
@pvinis: you can see it here: https://deploy-preview-857--react-native.netlify.com/upgrade

lucasbento:
@iamclaytonray, @JCMais: I agree with both of you, we still need to have a decision about the upgrade instructions for each library so this won't be done right now, probably on a second iteration for this page.

watadarkstar:
Update: We are going to really simplify this for our first iteration and focus on replicating the work that has been done here: https://pvinis.github.io/purge-web/ by @pvinis 🎉

lucasbento:
@watadarkstar: can you update the PR description to have the simplified mockup?

watadarkstar:
@lucasbento done

pvinis:
just keep maybe the stuff you did in a branch or an issue so we can refer to and add later. :) great job!

watadarkstar:
This is ready to be reviewed as a first iteration but can someone kick off the react-native-bot to build a live preview :)

charpeni:
Sure!

watadarkstar:
Thank you @charpeni :)

watadarkstar:
Closing since I'm no longer working on this. And it's probably going to be separate from the docs.

PR #859, opened 2019-03-30
Suggest adding overflow: 'hidden' if borderRadius fails

murtraja:


murtraja:
Motivation: https://github.com/facebook/react-native/issues/24207

PR #888, opened 2019-04-15
Update textinput.md

SilukovKonstantin:


rickhanlonii:
@SilukovKonstantin thanks for the update - I think the current behavior is expected, from the code comment: 

> If you type something in the text box that is a color, the background will change to that color

PR #873, opened 2019-04-07
chore: remove accessibilityComponentType and accessibilityTraits props

elucaswork:
https://github.com/facebook/react-native/pull/24344

Remove deprecated `accessibilityComponentType` and `accessibilityTraits` props.

PR #890, opened 2019-04-18
fix accessibility docs with conflicts

elucaswork:
https://github.com/facebook/react-native-website/pull/873 got merged with conflicts. Fixing it now. My bad.

cc @charpeni @cpojer


charpeni:
Thank you ❤️

PR #891, opened 2019-04-18
Remove autoComplete prop from 0.58 & 0.59

### LINKED ISSUE #882: TextInput's autoComplete prop isn't available it but it's documented as available
I just found misinformation in the docs: in [TextInput](https://facebook.github.io/react-native/docs/0.58/textinput#autocomplete) says the `autoComplete` prop is available (since 0.58) but this feature is not available it in any version, it's only in #master, as can be seen here https://github.com/facebook/react-native/commit/f15145639dab1e8d7a1c79a127b7d45c91d025a8.

`autoComplete` prop's PR https://github.com/facebook/react-native-website/pull/606.

Is there a way to update 0.58 & 0.59 docs to remove it?
Just let me know how If I can help to solve this issue.

cc @charpeni

charpeni:
Fixes #882.

### DIFF
--- a/website/versioned_docs/version-0.58/textinput.md
+++ b/website/versioned_docs/version-0.58/textinput.md
@@ -98,7 +98,6 @@ Note that on Android performing text selection in input can change app's activit
 
 * [`allowFontScaling`](textinput.md#allowfontscaling)
 * [`autoCapitalize`](textinput.md#autocapitalize)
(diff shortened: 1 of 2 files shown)


gianpaj:
Should we add `autoCompleteType` (only Android), introduced in 0.59?
https://github.com/facebook/react-native/commit/179d490#diff-b48972356bc8dca4a00747d002fc3dd5

PR #894, opened 2019-04-20
Docs - Alert.alert() does not have a type argument

JonnyBurger:
As can be seen in the implementation, there is no type argument for `Alert.alert`:

https://github.com/facebook/react-native/blob/master/Libraries/Alert/Alert.js#L47


rickhanlonii:
Thanks @JonnyBurger! Looks like type is an arg for Alert.prompt

PR #901, opened 2019-04-25
Update Headless Tasks for Android

rpander93:
Hi all,

Seems to me the docs for Headless Tasks on Android are incorrect. Instead of registering a function that require()'s the task, you should return the task function instead.

Also see https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/AppRegistry.js#L265. `taskProvider()` returns the task function which then gets called with the data.


cpojer:
Thank you! This is actually the same code, so either should be fine.

PR #902, opened 2019-04-26
Recommend using Android App Bundle for publishing Android app

kmagiera:
This PR updates signed-apk-android with a recommendation to use Android App Bundle as opposed to the traditional method of building APK (or split APKs).

The benefits of using Android App Bundle is that it allows for automatic APK splitting done by Google Play. This way saving developers from distributing natively compiled binaries (especially JSC) compiled for four CPU architectures to all their users.

I described the motivation behind this more on a blog post here https://blog.swmansion.com/make-your-react-native-app-3x-smaller-44c993eda2c9

Notes: I decided to rename the section to "Publishing to Google Play Store" as the previous title was no longer relevant but keeping the "id" the same to keep the same URL

### DIFF
--- a/docs/signed-apk-android.md
+++ b/docs/signed-apk-android.md
@@ -1,17 +1,17 @@
 ---
 id: signed-apk-android
-title: Generating Signed APK
+title: Publishing to Google Play Store
 ---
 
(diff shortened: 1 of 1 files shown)


docs/signed-apk-android.md
```diff
@@ -140,3 +138,7 @@ To enable Proguard, edit `android/app/build.gradle`:
  */
 def enableProguardInReleaseBuilds = true
 ```
+
+### Migrating old Android React Native apps to use App Signing by Google Play
```
TheSavior: This seems like something we should mention in the release notes for the next release, right? cc @turnrye

turnrye: @TheSavior I think we should mention it -- thanks for bringing this to my attention!

PR #909, opened 2019-05-04
Remove extra commas from example.

tmartin1:


cpojer:
Thanks for your PR! We actually adopted this extra comma style across many codebases because it means that adding a new argument will not trigger a diff (in a commit/PR) that affects two lines (one with an added comma + the new line) and instead will only show a single changed line. I'd prefer to keep it that way and React Native is set up to deal with this.

PR #911, opened 2019-05-06
Adds Drop app to showcase

drop-george:
Adds the Drop app to the showcase page. Please see screenshot below: 
image


rickhanlonii:
Hey @drop-george, unfortunately we do not accept any PRs that modify the showcase

PR #912, opened 2019-05-06
Add additional accessibility roles and states to the documentation.

marcmulcahy:
This PR adds the additional accessibility roles and states added in https://github.com/facebook/react-native/pull/24095


elucaswork:
@marcmulcahy do you mind to update [`docs/view.md`](https://github.com/facebook/react-native-website/blob/master/docs/view.md) as well because it also has references `accessibiltyRole` and `accessibilityStates` (https://facebook.github.io/react-native/docs/view#accessibilityrole). Thanks

PR #920, opened 2019-05-10
Add detail of `Platform.nightMode`

turnrye:
This is the user documentation for https://github.com/facebook/react-native/pull/24790.

I added a brief example that follows some of the Material Design paradigms, as well as a link to the Android resource for additional details.

I don't love that this landed on the "Platform Specific Code" page, but it seems like the best part of the documentation for now.


cpojer:
Closing as we are not moving forward with this API right now.

PR #927, opened 2019-05-15
Default fresco dependency added in doc, which is necessary for gif support.

adityasonel:
Default fresco dependency is needed for gif and webp support on Android/IOS. 

Currently default fresco dependency not included in docs, developer confused sometime by the gif is not working after following the docs correctly. But the problem is that they don't have default fresco dependency in there project.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -78,6 +78,9 @@ You will need to add some optional modules in `android/app/build.gradle`, depend
 
 ```gradle
 dependencies {
+  //Add fresco for gif support (if not included in your project)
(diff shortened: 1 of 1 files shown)


cpojer:
Hello! I'm not sure if this is correct. Fresco is actually shipped together with React Native :)

adityasonel:
I actually tried this and the error occurred. That's why I suggested change.

And If we add one more line in the docs, I think that's no problem for anyone. But any developer in the world is still stuck on that line then that's the problem!

PR #932, opened 2019-05-16
Emphasize non-Oracle OpenJDK distributions in Getting Started guide

mikehardy:
As discussed with @cpojer - we should stop directing people to a JDK download site where it is seemingly impossible to fetch a JDK8 install, and even if it were possible would be subject to sub-optimal update and license policies.

Commit message with resasoning for the specific change and a link for more context follows:

Markdown: emphasize non-Oracle OpenJDK distribution
   
As of April 2019 Oracle restricts it's non-OpenJDK distributions to
non-commercial use, and the Oracle OpenJDK builds will not be updated
promptly for security patches

non-Oracle OpenJDK versions will still be distributed as users would expect
(open-source and unrestricted commercial use, plus regular updates), so the
getting started guide should emphasize non-Oracle OpenJDK versions

Of those, the AdoptOpenJDK project appears to hae solid delivery for macOS
via brew and for non-system-packager Linux, so it is linked first

Context: https://medium.com/@javachampions/java-is-still-free-2-0-0-6b9aa8d6d244

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -301,7 +301,7 @@ You will also need to install the Xcode Command Line Tools. Open Xcode, then cho
 
 ### Java Development Kit
 
-React Native works best with version 8 or newer of the Java SE Development Kit (JDK). You may download and install [OpenJDK](http://openjdk.java.net) from [AdoptOpenJDK](https://adoptopenjdk.net/) or your system packager. You may also [Download and install Oracle JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) if desired.
+React Native requires version 8 or newer of the Java SE Development Kit (JDK). You may download and install [OpenJDK](http://openjdk.java.net) from [AdoptOpenJDK](https://adoptopenjdk.net/) or your system packager. You may also [Download and install Oracle JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) if desired.
 
 <block class="native mac linux windows android" />
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -214,18 +214,22 @@ While you can use any editor of your choice to develop your app, you will need t
 
 <block class="native mac ios android" />
 
-### Node, Watchman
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -293,11 +297,11 @@ You will also need to install the Xcode Command Line Tools. Open Xcode, then cho
 
 ![Xcode Command Line Tools](/react-native/docs/assets/GettingStartedXcodeCommandLineTools.png)
 
-<block class="native mac linux android" />
+<block class="native linux android" />
 
 ### Java Development Kit
 
-React Native requires a recent version of the Java SE Development Kit (JDK). [Download and install Oracle JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) if needed. You can also use [OpenJDK 8](http://openjdk.java.net/install/) as an alternative.
+React Native works best with version 8 or newer of the Java SE Development Kit (JDK). You may download and install [OpenJDK](http://openjdk.java.net) from [AdoptOpenJDK](https://adoptopenjdk.net/) or your system packager. You may also [Download and install Oracle JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) if desired.
```
kelset: I think we should remove the "or newer" part - I seem to recall there were issues with the more recent JDKs.

mikehardy: Agreed - I vaguely remember the same but don't have the institutional knowledge you'd have on that.

cpojer: I updated it.

mikehardy:
:zap: :+1:  what's the lag on publish to https://facebook.github.io/react-native/docs/getting-started so I can make sure I didn't bust the page's style?

kelset:
I think that by default this change will only be visible when the version is changed to 0.60 - I think that to modify the one current out you'd have to also change the generated one

mikehardy:
Okay - I'll check then - thanks

mikehardy:
Just verified in the RN0.60 preview #967 that my changes did not break the fancy/cool OS-selection tabs etc - this change appears to be good to go / no touch-ups needed

PR #934, opened 2019-05-16
Update Docusaurus to 1.9.0

yangshun:
### Summary

- Update Docusaurus to 1.9.0
- Enable collapsible side bar
- Clean up some hrefs that were pointing to `.html` extensions. This should have been done when the `cleanUrl` option was enabled but it still works because we generate both `/path/index.html` and `/path.html` in v1. This will break in v2 so better change it now.
- Enable the scroll-to-top button

Screen Shot 2019-05-16 at 11 53 26 AM

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -68,8 +68,7 @@ Let's break down what's happening here. In the `FadeInView` constructor, a new `
 
 When the component mounts, the opacity is set to 0. Then, an easing animation is started on the `fadeAnim` animated value, which will update all of its dependent mappings (in this case, just the opacity) on each frame as the value animates to the final value of 1.
 
-This is done in an optimized way that is faster than calling `setState` and re-rendering.  
-Because the entire configuration is declarative, we will be able to implement further optimizations that serialize the configuration and runs the animation on a high-priority thread.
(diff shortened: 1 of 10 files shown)


rickhanlonii:
Whaaaat, thanks @yangshun!

cpojer:
Thanks so much @yangshun, this is so much better :)

I'm wondering if there is a way to split up the Docs and API pages into two separate top-level entry points both with their own sidebars? For Docs, it should only have the Guides sections and for API it should have the Components and API sections. Is this possible?

charpeni:
Thank you, @yangshun!

yangshun:
Welcome! I'm always available to answer any Docusaurus-specific questions. Cheers! 🎉

cpojer:
I just noticed that this also collapses the sidebar in the blog section. Is it possible to always expand it there?

yangshun:
This is a limitation of the collapsible sidebar. I'll modify Docusaurus core to not collapse for blogs.

yangshun:
PR for the fix in https://github.com/facebook/Docusaurus/pull/1519. Will update RN website when that PR is merged.

PR #937, opened 2019-05-17
Remove TimerMixin

cpojer:
It ded.


mrbrentkelly:
@cpojer Is there anything in the 0.60 changelog about this? Does this mean in React Native you no longer need `react-timer-mixin` (or something like https://github.com/fractaltech/react-native-timer) to use `setTimeout`?

PR #941, opened 2019-05-19
Add documentation about Linking.canOpenURL on iOS and promise rejection

thib92:
[The official Apple documentation about `UIApplication::canOpenURL`](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl) warns about making too many calls to this method under certain conditions.

I added this warning and a link to the page on the relevant React-Native doc page.

Moreover, I saw on the source code for iOS and Android that the promise can be rejected. Even though we could see it in the example source code, there was no clear documentation about it. I added explanations.


docs/linking.md
```diff
@@ -206,6 +208,9 @@ The method returns a `Promise` object. When it is determined whether or not the
 
 > As of iOS 9, your app needs to provide the `LSApplicationQueriesSchemes` key inside `Info.plist` or canOpenURL will always return false.
 
+However, be careful as this method has limitations on iOS 9+. From [the official Apple documentation](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl):
```
rickhanlonii: How about this?

```suggestion
> This method has limitations on iOS 9+. From [the official Apple documentation](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl):

If your app is linked against an earlier version of iOS but is running in iOS 9.0 or later, you can call this method up to 50 times. After reaching that limit, subsequent calls always return false. If the user reinstalls or upgrades the app, iOS resets the limit.
```

docs/linking.md
```diff
@@ -196,6 +196,8 @@ Determine whether or not an installed app can handle a given URL.
 
 The method returns a `Promise` object. When it is determined whether or not the given URL can be handled, the promise is resolved and the first parameter is whether or not it can be opened.
 
+The `Promise` will reject on Android if it was impossible to check if the URL can be opened, and on iOS if you didn't add the specific scheme in the `LSApplicationQueriesSchemes` key inside `Info.plist` (see bellow).
```
rickhanlonii: What if this was in a note block below with the other notes as:

> On Android, the `Promise` will reject if it was impossible to check if the URL can be opened.

thib92:
So should I target the 0.59 branch with this PR? Or open another PR?

rickhanlonii:
In this PR you can copy the whole Linking.md file you updated over to `website/versioned_docs/version-0.59/Linking.md`, that will update Linking.md in 0.59

docs/linking.md
thib92: I thought it would make sense to add this bellow the description of the returned promise. The note about iOS 9+ is much more of a side thing, isn't it?

thib92:
Okay thanks! Will add when the last comment is resolved 😄

docs/linking.md
rickhanlonii: Ah yeah that makes sense 👍

thib92:
@rickhanlonii just added the page for the `0.59` docs. Thanks for the review 😉

rickhanlonii:
Looks good! 

Screenshot 2019-05-20 16 30 16

rickhanlonii:
Thanks @thib92!

Issue #883, opened 2019-04-11
Typo in ScrollView documentation


wagnermaciel:
There seems to be a typo here in scroll view documentation for the `keyboardShouldPersistTaps` prop. In particular, the sentence describing the "handled" option.

**Current**:
"the keyboard will not dismiss automatically when the tap was handled by a children"

**Proposed change**:
"the keyboard will not dismiss automatically when the tap was handled by children of the scroll view"

[Link to the documentation](https://facebook.github.io/react-native/docs/scrollview.html#keyboardshouldpersisttaps)

rickhanlonii:
Fixed, thanks!

PR #944, opened 2019-05-21
Typo in ScrollView documentation #883

Conviley:
Original issue #883 

#### Changed
"the keyboard will not dismiss automatically when the tap was handled by a children"
#### to
"the keyboard will not dismiss automatically when the tap was handled by children of the scroll view"

as suggested by @wagnermaciel

This is my first time ever making a pull request to an open source project and i'm not entierly sure about the changes in `website/i18n/en.json`  but they were automatically generated and i saw that that particular file was supposed to be included according to the `.gitignore` so i left them there and brought them along.


Conviley:
@rickhanlonii If it should, I can surely add it. I changed it there at first however it did not reflect on the website so i left it.😅

rickhanlonii:
Yeah updating both will put it live on the site (for 0.59) and in the docs for the next release 👍

rickhanlonii:
Thanks @Conviley!

PR #947, opened 2019-05-21
add definition and example of how flexbox works

ecreeth:

### DIFF
--- a/docs/flexbox.md
+++ b/docs/flexbox.md
@@ -15,7 +15,6 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
 
 In the following example the red, yellow and the green views are all children in the container view that got `flex: 1`. The red view got `flex: 1` , the yellow view got `flex: 2` and the green view got `flex: 3` . **1+2+3 = 6** which means that red view will get `1/6` of the space, the yellow `2/6` of the space and the red `3/6` of the space.
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Flex](https://cdn-images-1.medium.com/max/800/1*PhCFmO5tYX_sZSyCd4vO3w.png)
 
 #### Flex Direction
@@ -52,7 +51,7 @@ export default class FlexDirectionBasics extends Component {
 // skip this line if using Create React Native App
 AppRegistry.registerComponent('AwesomeProject', () => FlexDirectionBasics);
 ```
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
+
 ![Flex Direction](https://cdn-images-1.medium.com/max/800/1*rA7IbuUsJWsx6evKAsabVw.png)
 
 ### Layout Direction
@@ -104,7 +103,7 @@ export default class JustifyContentBasics extends Component {
 // skip this line if using Create React Native App
 AppRegistry.registerComponent('AwesomeProject', () => JustifyContentBasics);
 ```
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
+
 ![Justify Content](https://cdn-images-1.medium.com/max/800/1*i5TVlme-TisAVvD5ax2yPA.png)
 
 #### Align Items
@@ -153,14 +152,12 @@ export default class AlignItemsBasics extends Component {
 AppRegistry.registerComponent('AwesomeProject', () => AlignItemsBasics);
 ```
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Align Items](https://cdn-images-1.medium.com/max/800/1*evkM7zfxt-9p-HJ1M0Bh2g.png)
 
 ### Align Self
 
 [`alignSelf`](https://facebook.github.io/react-native/docs/layout-props#alignself) has the same options and effect as `alignItems` but instead of affecting the children within a container, you can apply this property to a single child to change its alignment within its parent. `alignSelf` overrides any option set by the parent with `aligItems`.
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Align Self](https://cdn-images-1.medium.com/max/800/1*J1JCoKwLCokX9JXVBvP71g.png)
 
 ### Align Content
@@ -181,7 +178,6 @@ See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-n
 
 learn more [here](https://yogalayout.com/docs/align-content)
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Align Content](https://cdn-images-1.medium.com/max/800/1*cC2XFyCF_igp20Ombt4wBw.png)
 
 ### Flex Wrap
@@ -189,9 +185,8 @@ See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-n
 The [`flexWrap`](https://facebook.github.io/react-native/docs/layout-props#flexwrap) property is set on containers and controls what happens when children overflow the size of the container along the main axis. By default children are forced into a single line (which can shrink elements).
 If wrapping is allowed items are wrapped into multiple lines along the main axis if needed. 
 
-When wrapping lines `alignContent` can be used to specify how the lines are placed in the container. learn more [here](https://yogalayout.com/docs/flex-wrap)
+When wrapping lines `alignContent` can be used to specify how the lines are placed in the container. learn more [here]
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Flex Wrap](https://cdn-images-1.medium.com/max/800/1*_7v4uQhSsuCn1cfeOMVfrA.png)
 
 ### Flex Basis, Grow, and Shrink
@@ -228,7 +223,6 @@ The `position` type of an element defines how it is positioned within its parent
 
 `absolute` When positioned absolutely an element doesn't take part in the normal layout flow. It is instead laid out independent of its siblings. The position is determined based on the `top`, `right`, `bottom`, and `left` values. 
 
-See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
 ![Absolute & Relative Layoutp](https://cdn-images-1.medium.com/max/800/1*NlPeRQCQK3Vb5nyjL0Mqxw.png)
 
 #### Going Deeper
@@ -238,3 +232,5 @@ You can play a little with the [yoga playground](https://yogalayout.com/playgrou
 We've covered the basics, but there are many other styles you may need for layouts. The full list of props that control layout is documented [here](./layout-props.md).
 
 We're getting close to being able to build a real application. One thing we are still missing is a way to take user input, so let's move on to [learn how to handle text input with the TextInput component](handling-text-input.md).
+
+See some examples from [Wix Engineers](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
--- a/docs/flexbox.md
+++ b/docs/flexbox.md
@@ -11,7 +11,7 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
 
 ### Layout Direction
 
-Layout direction specifies the direction in which children and text in a hierarchy should be laid out. Layout direction also affects what edge `start` and `end` refer to. By default Yoga lays out with LTR layout direction. In this mode `start` refers to left and `end` refers to right.
+Layout direction specifies the direction in which children and text in a hierarchy should be laid out. Layout direction also affects what edge `start` and `end` refer to. By default React Native lays out with LTR layout direction. In this mode `start` refers to left and `end` refers to right.
 
 - `LTR` (**default value**) Text and children and laid out from left to right. Margin and padding applied the start of an element are applied on the left side.
 
--- a/docs/flexbox.md
+++ b/docs/flexbox.md
@@ -11,15 +11,15 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
 
 ### Flex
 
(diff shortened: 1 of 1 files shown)


docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
 
 > Flexbox works the same way in React Native as it does in CSS on the web, with a few exceptions. The defaults are different, with `flexDirection` defaulting to `column` instead of `row`, and the `flex` parameter only supporting a single number.
 
+### Layout Direction
```
rickhanlonii: What do you think about moving this section down? It may flow better if something like `flex` or `flex direction` is first

docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
...
+
+Layout direction specifies the direction in which children and text in a hierarchy should be laid out. Layout direction also affects what edge `start` and `end` refer to. By default Yoga lays out with LTR layout direction. In this mode `start` refers to left and `end` refers to right.
```
rickhanlonii: Do we need to mention Yoga here? It isn't introduced or used anywhere else in the doc

```suggestion
Layout direction specifies the direction in which children and text in a hierarchy should be laid out. Layout direction also affects what edge `start` and `end` refer to. By default React Native lays out with LTR layout direction. In this mode `start` refers to left and `end` refers to right.
```

docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
...
+
+- `LTR` (**default value**) Text and children and laid out from left to right. Margin and padding applied the start of an element are applied on the left side.
+
+- `RTL` Text and children and laid out from right to left. Margin and padding applied the start of an element are applied on the right side.
+
+### Flex
+
+[`flex`](https://facebook.github.io/react-native/docs/layout-props#flex) will define how your items are going to **“fight”** over the available space along your primary axis. Most of the time you will want your app container to be flex:1 to take all of the screen height. Space will be divided according to each element flex property. In the following example the red, yellow and the green views are all children in the container view that got flex:1. The red view got flex:1 , the yellow view got flex:2 and the green view got flex:3 . 1+2+3=6 which means that red view will get 1/6 of the space, the yellow 2/6 of the space and the red 3/6 of the space. I think you got it . From [wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c) engineers
```
rickhanlonii:
```suggestion
[`flex`](https://facebook.github.io/react-native/docs/layout-props#flex) will define how your items are going to fill the available space along your primary axis. Space will be divided according to each element flex property.

In the following example the red, yellow and the green views are all children in the container view that got flex:1. The red view got flex:1 , the yellow view got flex:2 and the green view got flex:3 . 1+2+3=6 which means that red view will get 1/6 of the space, the yellow 2/6 of the space and the red 3/6 of the space.
```

- Switch "fight" to "fill"
- Remove "most of the time  you will use flex: 1" because I don't think this is true 
- Add whitespace
- Remove "i think you got it"
- Remove "from wix engineers" because the next paragraph has the same content

docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
...
+
+See some examples of [wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c) engineers
+![Flex](https://cdn-images-1.medium.com/max/800/1*PhCFmO5tYX_sZSyCd4vO3w.png)
+
 #### Flex Direction
 
+[`flexDirection`](https://facebook.github.io/react-native/docs/layout-props#flexdirection) controls the direction in which children of a node are laid out. This is also referred to as the main axis. The main axis is the direction in which children are laid out. The cross axis the the axis perpendicular to the main axis, or the axis which wrapping lines are laid out in.
```
rickhanlonii:
```suggestion
[`flexDirection`](https://facebook.github.io/react-native/docs/layout-props#flexdirection) controls the direction in which children of a node are laid out. This is also referred to as the main axis. The cross axis the the axis perpendicular to the main axis, or the axis which wrapping lines are laid out in.
```

- Remove repeated statement

docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
...
+
+- `row` (**default value**) Align children from left to right. If wrapping is enabled then the next line will start under the first item on the left of the container.
```
rickhanlonii: **Important: the default value in React Native is column**

This may be worth calling out in a not somewhere because it is a common issue people run into

docs/flexbox.md
```diff
@@ -9,10 +9,37 @@ You will normally use a combination of `flexDirection`, `alignItems`, and `justi
...
+
+- `column` Align children from top to bottom. If wrapping is enabled then the next line will start to the left first item on the top of the container.
+
+- `row-reverse` Align children from right to left. If wrapping is enabled then the next line will start under the first item on the right of the container.
+
+- `column-reverse` Align children from bottom to top. If wrapping is enabled then the next line will start to the left first item on the bottom of the container.
+
 Adding `flexDirection` to a component's `style` determines the **primary axis** of its layout. Should the children be organized horizontally (`row`) or vertically (`column`)? The default is `column`.
```
rickhanlonii: This sentence refers to the "primary axis" while the above text refers to the "main axis", we should be consistent in the naming

docs/flexbox.md
```diff
@@ -33,11 +60,27 @@ export default class FlexDirectionBasics extends Component {
 // skip this line if using Create React Native App
 AppRegistry.registerComponent('AwesomeProject', () => FlexDirectionBasics);
 ```
+See some examples of [wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c) engineers
```
rickhanlonii:
```suggestion
See some examples from [Wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c):
```

docs/flexbox.md
```diff
@@ -33,11 +60,27 @@ export default class FlexDirectionBasics extends Component {
...
+
+- `flex-start`(**default value**) Align children of a container to the start of the container's main axis.
+
+- `flex-end` Align children of a container to the end of the container's main axis.
+
+- `center` Align children of a container in the center of the container's main axis.
+
+- `space-between` Evenly space of children across the container's main axis, distributing remaining space between the children.
+
+- `space-around` Evenly space of children across the container's main axis, distributing remaining space around the children. Compared to `space between` using `space around` will result in space being distributed to the beginning of the first child and end of the last child.
```
rickhanlonii:
```suggestion
- `space-around` Evenly space of children across the container's main axis, distributing remaining space around the children. Compared to `space-between` using `space-around` will result in space being distributed to the beginning of the first child and end of the last child.
```

- Add hyphens in variable names

docs/flexbox.md
```diff
@@ -33,11 +60,27 @@ export default class FlexDirectionBasics extends Component {
...
+
 Adding `justifyContent` to a component's style determines the **distribution** of children along the **primary axis**. Should children be distributed at the start, the center, the end, or spaced evenly? Available options are `flex-start`, `center`, `flex-end`, `space-around`, `space-between` and `space-evenly`.
```
rickhanlonii: This paragraph is now just repeating what's stated right before it, should we remove it?

rickhanlonii: Let's make sure to keep the yoga link

docs/flexbox.md
```diff
@@ -63,13 +106,29 @@ export default class JustifyContentBasics extends Component {
 // skip this line if using Create React Native App
 AppRegistry.registerComponent('AwesomeProject', () => JustifyContentBasics);
 ```
+See some examples of [wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c) engineers
```
rickhanlonii: I think we should consider adding a `"Source: Wix Engineers"` with a link below the images instead of each one having a leading sentence

docs/flexbox.md
```diff
@@ -63,13 +106,29 @@ export default class JustifyContentBasics extends Component {
...
+![Justify Content](https://cdn-images-1.medium.com/max/800/1*i5TVlme-TisAVvD5ax2yPA.png)
 
 #### Align Items
 
+[`alignItems`](https://facebook.github.io/react-native/docs/layout-props#alignitems) describes how to align children along the cross axis of their container. Align items is very similar to `justifyContent` but instead of applying to the main axis, `align items` applies to the cross axis.
```
rickhanlonii:
```suggestion
[`alignItems`](https://facebook.github.io/react-native/docs/layout-props#alignitems) describes how to align children along the cross axis of their container. Align items is very similar to `justifyContent` but instead of applying to the main axis, `alignItems` applies to the cross axis.
```

- Fix variable

docs/flexbox.md
```diff
@@ -63,13 +106,29 @@ export default class JustifyContentBasics extends Component {
...
+
+- `flex-start` Align children of a container to the start of the container's cross axis.
+
+- `flex-end` Align children of a container to the end of the container's cross axis.
+
+- `center` Align children of a container in the center of the container's cross axis.
+
+- `baseline` Align children of a container along a common baseline. Individual children can be set to be the reference baseline for their parents.
+
 Adding `alignItems` to a component's style determines the **alignment** of children along the **secondary axis** (if the primary axis is `row`, then the secondary is `column`, and vice versa). Should children be aligned at the start, the center, the end, or stretched to fill? Available options are `flex-start`, `center`, `flex-end`, and `stretch`.
```
rickhanlonii: Again, this paragraph just restates the content before it

docs/flexbox.md
```diff
@@ -98,8 +157,88 @@ export default class AlignItemsBasics extends Component {
 AppRegistry.registerComponent('AwesomeProject', () => AlignItemsBasics);
 ```
 
+See some examples of [wix](https://medium.com/wix-engineering/the-full-react-native-layout-cheat-sheet-a4147802405c) engineers
```
rickhanlonii: Ditto comment above about replacing with a source comment

docs/flexbox.md
```diff
@@ -98,8 +157,88 @@ export default class AlignItemsBasics extends Component {
...
+
+- `stretch` wrapped lines to match the height of the container's cross axis.
+
+- `center` Align wrapped lines in the center of the container's cross axis.
+
+- `space-between` Evenly space wrapped lines across the container's main axis, distributing remaining space between the lines.
+
+- `space-around` Evenly space wrapped lines across the container's main axis, distributing remaining space around the lines. Compared to space between using space around will result in space being distributed to the begining of the first lines and end of the last line.
+
+learn more [here](https://yogalayout.com/docs/align-content)
```
rickhanlonii: This is on a new line so it should be capitalized (same for all of the Learn more text in this doc)

docs/flexbox.md
```diff
@@ -98,8 +157,88 @@ export default class AlignItemsBasics extends Component {
...
+![Flex Wrap](https://cdn-images-1.medium.com/max/800/1*_7v4uQhSsuCn1cfeOMVfrA.png)
+
+### Flex Basis, Grow, and Shrink
+
+- [`flexGrow`](https://facebook.github.io/react-native/docs/layout-props#flexgrow) describes how any space within a container should be distributed among its children along the main axis. After laying out its children, a container will distribute any remaining space according to the flex grow values specified by its children.
+
+  flexGrow accepts any floating point value >= 0, with 0 being the default value. A container will distribute any remaining space among its children weighted by the child’s flex grow value.
+
+- [`flexShrink`](https://facebook.github.io/react-native/docs/layout-props#flexshrink) describes how to shrink children along the main axis in the case that the total size of the children overflow the size of the container on the main axis. flex shrink is very similar to flex grow and can be thought of in the same way if any overflowing size is considered to be negative remaining space. These two properties also work well together by allowing children to grow and shrink as needed.
```
rickhanlonii:
```suggestion
- [`flexShrink`](https://facebook.github.io/react-native/docs/layout-props#flexshrink) describes how to shrink children along the main axis in the case that the total size of the children overflow the size of the container on the main axis. Flex shrink is very similar to flex grow and can be thought of in the same way if any overflowing size is considered to be negative remaining space. These two properties also work well together by allowing children to grow and shrink as needed.
```

- Capitalize

docs/flexbox.md
```diff
@@ -98,8 +157,88 @@ export default class AlignItemsBasics extends Component {
...
+
+  Flex shrink accepts any floating point value >= 0, with 1 being the default value. A container will shrink its children weighted by the child’s flex shrink value.
+
+- [`flexBasis`](https://facebook.github.io/react-native/docs/layout-props#flexbasis) is an axis-independent way of providing the default size of an item along the main axis. Setting the flex basis of a child is similar to setting the `width` of that child if its parent is a container with `flex direction: row` or setting the `height` of a child if its parent is a container with `flex direction: column`. The flex basis of an item is the default size of that item, the size of the item before any flex grow and flex shrink calculations are performed.
+
+learn more [here](https://yogalayout.com/docs/flex)
+
+### Width and Height
+
+The `width` property in Yoga specifies the width of the element's content area. Similarly height property specifies the `height` of the element's content area.
```
rickhanlonii:
```suggestion
The `width` property in React Native specifies the width of the element's content area. Similarly height property specifies the `height` of the element's content area.
```

docs/flexbox.md
```diff
@@ -98,8 +157,88 @@ export default class AlignItemsBasics extends Component {
...
+
+Both `width` and `height` can take following values:
+
+- `auto` Is the default Value, Yoga calculates the width/height for the element based on its content, whether that is other children, text, or an image.
+
+- `pixels` Defines the width/height in absolute pixels. Depending on other properties set on the Yoga node this may or may not be the final dimension of the node.
```
rickhanlonii:
```suggestion
- `pixels` Defines the width/height in absolute pixels. Depending on other styles set on the component, this may or may not be the final dimension.
```

docs/flexbox.md
ecreeth: You're right! Thank you

docs/flexbox.md
ecreeth: Thanks, I think those who start will understand it much better that way.

docs/flexbox.md
ecreeth: thanks

docs/flexbox.md
ecreeth: Thanks for seeing that big mistake, I do not know what I was thinking about 😵

docs/flexbox.md
ecreeth: Excellent!

PR #953, opened 2019-05-22
[Formatting] Run prettier

rickhanlonii:
## Overview

Runs prettier on master


docs/easing.md
```diff
@@ -234,7 +234,7 @@ A useful tool to visualize cubic bezier curves can be found at http://cubic-bezi
 ### `in()`
 
 ```javascript
-static in(easing);
+static in easing;
```
rickhanlonii: This looks wrong?

cpojer: Yeah it does that for some reason. I'll manually revert.

PR #958, opened 2019-05-23
[FlatList] Add note on multi-column layouts

hramos:
Call out flexWrap anti-pattern for multi-column layouts.
Clean up renderItem docs a bit.


rickhanlonii:
Looks great

Screenshot 2019-05-23 20 23 04

...

Screenshot 2019-05-23 20 22 29

PR #961, opened 2019-05-26
Add .idea to gitignore

mohrash92:
Just a small one... I have included the `.idea` folder in the `.gitignore` for developers who use any editor from the IDEA family like myself. I figured if something like `.DS_Store` is in there then it's probably useful to have this 😄.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 .DS_Store
+.idea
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Sounds good - fwiw you can add .idea to your global .gitignore as well 👍

PR #965, opened 2019-05-28
Update actionsheetios.md

temitope:
added an example that was in previous documentation for the latest to give exlarity on what the _excludedActivityTypes_ referred to


cpojer:
We are actually planning on removing this module in favor of a third-party module, so we aren't looking to change the documentation any longer.

PR #966, opened 2019-05-28
Typo in index.js.

TonyChampion:
Should read "React Native are mobile web apps" instead of "aren't"


charpeni:
Doesn't look like a typo to me, because React Native apps aren't powered by HTML/CSS like a mobile web app with Cordova would have.

PR #970, opened 2019-05-30
Add the "optimizing-flatlist-configuration" topic

MohamadKh75:
Hello

As mentioned [here ](https://github.com/filipemerker/flatlist-performance-tips/issues/2), there is some useful information about `FlatList` which can be added to the official docs. 

I started to import them but there is a question (@cpojer): as the version 0.60 is already cut ([here](https://github.com/facebook/react-native-website/pull/967)), should my PR have 0.60 version? 

cc @kelset @filipemerker


cpojer:
Yeah, let's have a 0.60 and a master version.

filipemerker:
Wow @MohamadKh75, amazing work on the PR! I would definitely not be able to integrate it so well. When I get home today, I can take a look and help you with something!

website/versions.json
```diff
@@ -1,4 +1,5 @@
 [
+  "0.60",
```
filipemerker: Just change this to [0.60-RC](https://github.com/facebook/react-native-website/blob/master/website/versions.json#L2) and I think it will be fine.

MohamadKh75: Done!

kelset:
uhm just one thing - it seems that the PR is trying to add files for all the docs, I think you may have to rebase the PR and remove the versioned 0.60 files you generated

kelset:
cc @cpojer / @rickhanlonii I think it's ready for merging

PR #979, opened 2019-05-31
Use absolute links in Components and APIs

hramos:
Fixes https://github.com/facebook/react-native/issues/25104.

This introduces another concern: the links will always lead to the latest version of the docs, even if you're looking at an older version.

This whole page will be redone as part of the new website design, so that's less of a concern than plain broken links when looking at the RC docs.


brunolemos:
Check https://github.com/facebook/react-native-website/pull/981

hramos:
Resolved in https://github.com/facebook/react-native-website/commit/2eb9249853442b490046e45e4b154401c1fc78c6.

PR #980, opened 2019-06-01
Fix a few typos within docs/

mohrash92:
Fixed a few typos in the documentation 😄

### DIFF
--- a/docs/signed-apk-android.md
+++ b/docs/signed-apk-android.md
@@ -117,7 +117,7 @@ You can create an APK for each CPU by changing the following line in android/app
 + def enableSeparateBuildPerCPUArchitecture = true
 ```
 
-Upload both these files to markets which support device targetting, such as [Google Play](https://developer.android.com/google/play/publishing/multiple-apks.html) and [Amazon AppStore](https://developer.amazon.com/docs/app-submission/getting-started-with-device-targeting.html), and the users will automatically get the appropriate APK. If you want to upload to other markets, such as [APKFiles](https://www.apkfiles.com/), which do not support multiple APKs for a single app, change the following line as well to create the default universal APK with binaries for both CPUs.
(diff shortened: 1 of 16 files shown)


docs/signed-apk-android.md
```diff
@@ -117,7 +117,7 @@ You can create an APK for each CPU by changing the following line in android/app
...
 ```
 
-Upload both these files to markets which support device targetting, such as [Google Play](https://developer.android.com/google/play/publishing/multiple-apks.html) and [Amazon AppStore](https://developer.amazon.com/docs/app-submission/getting-started-with-device-targeting.html), and the users will automatically get the appropriate APK. If you want to upload to other markets, such as [APKFiles](https://www.apkfiles.com/), which do not support multiple APKs for a single app, change the following line as well to create the default universal APK with binaries for both CPUs.
+Upload both these files to markets which support device targeting, such as [Google Play](https://developer.android.com/google/play/publishing/multiple-apks.html) and [Amazon AppStore](https://developer.amazon.com/docs/app-submission/getting-started-with-device-targeting.html), and the users will automatically get the appropriate APK. If you want to upload to other markets, such as [APKFiles](https://www.apkfiles.com/), which do not support multiple APKs for a single app, change the following line as well to create the default universal APK with binaries for both CPUs.
```
rickhanlonii: What changed here?

kelset:
> @kelset which versions docs should we update this in to get it into the 0.60 release?

there should be a versioned_doc version called 0.60RC?

rickhanlonii:
Are we cutting the next release docs from master? If so, we could wait until the next version, but if we're cherry-picking the docs then we should add it now?

kelset:
I am not sure about that, @hramos will have a better answer, but I think the RC version will be "bumped" to become the 0.60 once we reach stable in terms of release.

rickhanlonii:
Awesome, thanks! @mohrash92 do you want to add these updates to website/versioned_docs/version-0.60-RC as well so they go live with the next release?

docs/signed-apk-android.md
mohrash92: `targetting` changed to `targeting`. Not sure why it's not highlighting here but if you select the "Display the rich diff" button you'll be able to see it.

mohrash92:
> Awesome, thanks! @mohrash92 do you want to add these updates to website/versioned_docs/version-0.60-RC as well so they go live with the next release?

Sure, thats done!

PR #984, opened 2019-06-03
Updating component links on Components And APIs page

millerbennett:
On the page displayed here: https://facebook.github.io/react-native/docs/components-and-apis#basic-components

Many of the component links are not working due to the links referring to a file in the parent folder. I've removed the relative parent folder reference ( ../ ) from the component links.


millerbennett:
I see this is a duplicate.

PR #1000, opened 2019-06-08
Update examples link

antronic:
I found that the examples link are outdated. So, I find some similar link to update it.


rickhanlonii:
Thanks!

PR #1010, opened 2019-06-12
Fix invalid markdown formatting

cpojer:
This was broken after exporting it from the tool that I used to write the post. Oops.


cpojer:
I'll merge https://github.com/facebook/react-native-website/pull/1011 instead :)

Issue #929, opened 2019-05-15
☂️ Website Redesign and Documentation Rewrite


cpojer:
The content and look of the React Native website are outdated. Many of the existing guides and the overall flow of the documentation were written in 2015 and do not reflect well what it is like to build apps with React Native in 2019. When we asked what people [dislike about React Native](https://github.com/react-native-community/discussions-and-proposals/issues/104), documentation issues came up many times. We received [more concrete feedback](https://github.com/facebook/react-native-website/issues/917) when we asked specifically.

This umbrella issue aims to outline the work necessary to update the website both with a new look and new content. While we'll handle the redesign work outside of this issue, we'd love to get your help in updating the entire documentation: if there is any item below issue that you'd love to help with, please send a Pull Request to this repo or get in touch otherwise! The best way to help right now is by updating the API docs with the actual implementations (see below) to ensure things are up-to-date for the upcoming release.

# Goals

* A fresh look for React Native that represents the framework in 2019.
* Guides for beginners and advanced developers that are best in class.
* Always up-to-date API documentation that is synced with actual code and examples.

# Redesign

The redesign is currently being worked on by @orta and @tesseralis. We are excited for you all to get a look very soon!

# New Guide Structure
The current structure is quite outdated. It's hard to know upfront exactly what the new flow should be like but here is the proposal for a new structure which is open for discussion:

* **Basics**
  * Setup & Getting Started
  * Basics & React – this should mirror reactjs.org with a focus on RN, gently introduce RN and React, and link to reactjs.org for concepts.
  * ~~Props~~ – Removed and rolled into basics, with links to reactjs.org.
  * ~~State~~ – Removed and rolled into basics, with links to reactjs.org.
  * Style, Height and Width – Rolled up into a new guide "Styling & Dimensions"
  * Layout or "Thinking in flexbox" – Significantly expand this guide with more screenshots and possibly a live playground similar to yogalayout.com (first improvements were merged in https://github.com/facebook/react-native-website/pull/947).
  * Handling Text Input and Handling Touches – Merge into a single "Handling User Interactions" guide
  * Using a ScrollView and Using a List Views – Merge into a single guide
  * ~~Networking~~ – Ideally we'll factor data fetching into the preceding guides, like populating a ScrollView content, for example.
* **New Guides**
  * Guide on "The basics of native development for Android and iOS" aimed at web developers including Xcode, Android Studio and Cocoapods and Gradle
  * Include docs on RN integration into Kotlin and Swift apps.
[...]

PR #1021, opened 2019-06-13
update slider.md

ecreeth:
This is part of #929 

### Was updated

-  The [`thumbTintColor`](https://github.com/facebook/react-native/blob/master/Libraries/Components/Slider/Slider.js#L112) prop


rickhanlonii:
Thanks!

PR #1022, opened 2019-06-13
update toastandroid.md

ecreeth:
This is part of #929 

> The properties of this component has not changed

### was added
- The [`ToastAndroid`](https://github.com/facebook/react-native/blob/master/Libraries/Components/ToastAndroid/ToastAndroid.android.js#L36) properties definitions

### DIFF
--- a/docs/toastandroid.md
+++ b/docs/toastandroid.md
@@ -137,6 +137,8 @@ static showWithGravityAndOffset(message, duration, gravity, xOffset, yOffset)
 
 ### `SHORT`
 
+Toast duration constant
+
 ```javascript
 ToastAndroid.SHORT;
 ```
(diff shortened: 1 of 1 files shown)


rachelnabors:
Hey there, @ecreeth! Thanks for your contribution! 

I am working with the React Native docs these days and am super looking forward to accepting your pull request! But also, I was wondering if you could provide some more context around what these values do? Like could you expand on `TOP`, `BOTTOM`, and `CENTER` to explain how they change the interaction? That would be so helpful for people new to this space!

If you can update your request, I'm happy to work with you to get it into the site!

Thanks again!

ecreeth:
Charmed! but I will have to delete this PR and open another one, since I have removed the fork before and it does not allow me to edit it. :smile:

PR #1028, opened 2019-06-15
Update stylesheet.md

ecreeth:
This is part of #929 

### was added

- The [`absoluteFillObject`](https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheet.js#L255) prop
- The [`compose`](https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheet.js#L263) method


rickhanlonii:
Thanks!

MrEdinLaw:
I see that you have updated the documentation there but the current code doesn't seem to reflect the current code, the `stylesheet.compose` seems to be simply returning both inputs in an array:
`return ([style1, style2]`
https://github.com/facebook/react-native/commit/50a481d

Is there something i do not understand here? Or is this function still under construction?

PR #1033, opened 2019-06-17
remove alertios.md

ecreeth:
This is part of #929 

## The **AlertIOS** was deleted because it was merged with **Alert**

Please see [#23318](https://github.com/facebook/react-native/pull/23318)


ecreeth:
> Sorry, I just sbumitted #1045 to reconsider removing these completely and instead replace them with a shell page directing to thee correct location

You're right, let's find a better way!

usrbowe:
Would it be possible to merge this change.
 It took me some time to find a release notes, that `AlertIOS` was merged into `Alert`. So current docs made me bit confused.
I agree that completely removing is not ideal, so the redirect link would be very useful.

rickhanlonii:
@ecreeth what do you think about adding text like "AlertIOS has been moved to Alert <link>"? We would update this PR for you but it looks like the fork was deleted

PR #1044, opened 2019-06-19
Add doc for countdown

harveyconnor:
Countdown is actually a native prop in iOS and can be used.


rickhanlonii:
Thanks for your first contribution @harveyconnor!

PR #1053, opened 2019-06-22
Change docs to use "contain" instead Image.resizeMode.contain

elrussord1:
Image no longer has a resizeMode property.

INSTALLINH


Arcanemagus:
@cpojer Just a heads up that this PR was from a spammer/bot account, they pulled a random commit from https://github.com/facebook/react-native-website/pull/449... luckily here without changing anything so it didn't actually cause any harm.

rickhanlonii:
Woah weird, cc @TheSavior since that was your commit

PR #1081, opened 2019-07-06
Anchor links bug fix

icreaterain:
On the page https://facebook.github.io/react-native/docs/webview anchor links have the form like 'webview#onmessage'. But in the menu at the top of the page we have links like 'webview.md#onmessage'. And on click we have 404 error. It fixes that bug.


charpeni:
Can you also apply this to the .md file in `docs` instead of `versioned_docs`?

icreaterain:
@charpeni, with pleasure, but there is no webview.md file in ./docs
https://github.com/facebook/react-native-website/tree/master/docs

PR #1085, opened 2019-07-08
Remove mention to deprecated Navigator in Modal page

tijunoi:
Navigator was removed from React Native in v0.44, and the navigation guide refers to `react-navigation` and `react-native-navigation` that have different implementations of navigators and this tip does not apply.


charpeni:
I feel like this is helpful as other libraries are offering a top-level navigator.

What do you think?

tijunoi:
I agree that the tip is useful if it points in the right direction, but there is not a reference link and the React Native top-level navigator does not exist anymore. Perhaps it would be better to change it to: 
>  Note: If you need more control over how to present modals over the rest of your app, consider consider using a navigation library [link to navigation.md].
But react-navigation/react-navigation which is the currently recommended library does not handle modal presentation, so it could be confusing if you're trying to figure it out.

What do you think? Should I update the PR?

PR #1086, opened 2019-07-09
Level the playing field for navigation libraries

grahammendick:
The "Navigation Between Screens" documentation was biased towards React Navigation. As per [my chat with @cpojer](https://twitter.com/grahammendick/status/1148215786760212481), I’ve rewritten it to give all navigation libraries a fair crack of the whip. To include a new navigation library you add a row to the feature table at the top. I’ve picked features that give an architectural overview of navigation libraries.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -3,72 +3,26 @@ id: navigation
 title: Navigating Between Screens
 ---
 
-Mobile apps are rarely made up of a single screen. Managing the presentation of, and transition between, multiple screens is typically handled by what is known as a navigator.
+By default, React Native renders your components onto a single screen. To create multiple screens and navigate between them you need a navigation library. The following table helps you choose the one that's right for you. Each column identifies a different feature and each row represents a different navigation library (React Router Native isn't included because it doesn't have stack navigation). These features aren't the be-all and end-all of what makes a good navigation library but they are important. Keep reading to find out why.
 
-This guide covers the various navigation components available in React Native. If you are just getting started with navigation, you will probably want to use [React Navigation](navigation.md#react-navigation). React Navigation provides an easy to use navigation solution, with the ability to present common stack navigation and tabbed navigation patterns on both iOS and Android.
+|                                                                                                           | Uses native API | Has one React root | Runs in Expo | Runs on the Web |
+| --------------------------------------------------------------------------------------------------------- | --------------- | ------------------ | ------------ | --------------- |
+| [React Navigation](https://reactnavigation.org/docs/en/getting-started.html)                              | No              | Yes                | Yes          | No              |
+| [React Native Navigation](https://wix.github.io/react-native-navigation/)                                 | Yes             | No                 | No           | No              |
+| [The Navigation router](https://grahammendick.github.io/navigation/documentation/native/hello-world.html) | Yes             | Yes                | No           | Yes             |
 
-If you'd like to achieve a native look and feel on both iOS and Android, or you're integrating React Native into an app that already manages navigation natively, the following library provides native navigation on both platforms: [react-native-navigation](https://github.com/wix/react-native-navigation).
+## Uses native API
 
-## React Navigation
+React Native uses the same fundamental UI building blocks as regular apps. Both iOS and Android have fundamental navigation building blocks Android has `Activities` or `Fragments` and iOS has `UIViewControllers`. If your navigation library doesn't use them then you can end up with an app that doesn't have a native look and feel. You'll also miss out on iOS navigation bar features like large titles and search bars.
 
-The community solution to navigation is a standalone library that allows developers to set up the screens of an app with just a few lines of code.
+## Has one React root
 
-The first step is to install in your project:
+If your navigation library renders all screens under a single React root then they share the React `Context`. That means you can use any state management library, or none, to pass data between them. If screens don't share the React `Context` then you have to hold the shared data in a global store outside of the React tree. That rules out state management solutions like Relay and Apollo and only rules in Redux.
 
-```
-npm install --save react-navigation
-```
+## Runs in Expo
 
-The second step is to install react-native-gesture-handler
+Expo provides a shared native runtime so you don't write native code. You focus on writing your React app in JavaScript. You don't have to worry about iOS or Android specific settings, or even opening up Xcode. Managed Expo projects have their own workflow including Expo CLI (a command line interface) and Expo Dev Tools (a web UI) to simplify the development and deployment process.
 
-```
-yarn add react-native-gesture-handler
-# or with npm
-# npm install --save react-native-gesture-handler
-```
+## Runs on the Web
 
-Now we need to link our react-native to react-native-gesture-handler
-
-```
-react-native link react-native-gesture-handler
-```
-
-Then you can quickly create an app with a home screen and a profile screen:
-
-```javascript
-import {createStackNavigator, createAppContainer} from 'react-navigation';
-
-const MainNavigator = createStackNavigator({
-  Home: {screen: HomeScreen},
-  Profile: {screen: ProfileScreen},
-});
-
-const App = createAppContainer(MainNavigator);
-
-export default App;
-```
-
-Each screen component can set navigation options such as the header title. It can use action creators on the `navigation` prop to link to other screens:
-
-```javascript
-class HomeScreen extends React.Component {
-  static navigationOptions = {
-    title: 'Welcome',
-  };
-  render() {
-    const {navigate} = this.props.navigation;
-    return (
-      <Button
-        title="Go to Jane's profile"
-        onPress={() => navigate('Profile', {name: 'Jane'})}
-      />
-    );
-  }
-}
-```
-
-React Navigation routers make it easy to override navigation logic. Because routers can be nested inside each other, developers can override navigation logic for one area of the app without making widespread changes.
-
-The views in React Navigation use native components and the [`Animated`](animated.md) library to deliver 60fps animations that are run on the native thread. Plus, the animations and gestures can be easily customized.
-
-For a complete intro to React Navigation, follow the [React Navigation Getting Started Guide](https://reactnavigation.org/docs/getting-started.html), or browse other docs such as the [Intro to Navigators](https://expo.io/@react-navigation/NavigationPlayground).
+React Native for Web runs your native app on the Web. You don't have to change any code because React Native for Web renders your React Native components as DOM elements. For example, it converts Views into divs. But this only works if your navigation library supports React Native for Web. To avoid compromising the Web experience, make sure that your navigation library can generate Hyoerlinks and render on the server-side (SSR).


docs/navigation.md
```diff
@@ -3,72 +3,26 @@ id: navigation
 title: Navigating Between Screens
 ---
 
-Mobile apps are rarely made up of a single screen. Managing the presentation of, and transition between, multiple screens is typically handled by what is known as a navigator.
+By default, React Native renders your components onto a single screen. To create multiple screens and navigate between them you need a navigation library. The following table helps you choose the one that's right for you. Each column identifies a different feature and each row represents a different navigation library (React Router Native isn't included because it doesn't have stack navigation). These features aren't the be-all and end-all of what makes a good navigation library but they are important. Keep reading to find out why.
```
rickhanlonii: Could we either include React Router Native and add a column for "stack natigation", or list all the the features that are uses as criteria used for inclusion. For example:

> Navigation libraries included below must all implement x, y, and z. These features aren't the be-all and end-all of what makes a good navigation library but they are important to cover a broad range of user experiance expectations.

rickhanlonii:
Here's a screenshot:

deploy-preview-1086--react-native netlify com_docs_next_navigation

@grahammendick I'll review the content but is it possible to make the table layout better?

docs/navigation.md
```diff
@@ -3,72 +3,26 @@ id: navigation
...
+| [React Native Navigation](https://wix.github.io/react-native-navigation/)                                 | Yes             | No                 | No           | No              |
+| [The Navigation router](https://grahammendick.github.io/navigation/documentation/native/hello-world.html) | Yes             | Yes                | No           | Yes             |
 
-If you'd like to achieve a native look and feel on both iOS and Android, or you're integrating React Native into an app that already manages navigation natively, the following library provides native navigation on both platforms: [react-native-navigation](https://github.com/wix/react-native-navigation).
+## Uses native API
 
-## React Navigation
+React Native uses the same fundamental UI building blocks as regular apps. Both iOS and Android have fundamental navigation building blocks Android has `Activities` or `Fragments` and iOS has `UIViewControllers`. If your navigation library doesn't use them then you can end up with an app that doesn't have a native look and feel. You'll also miss out on iOS navigation bar features like large titles and search bars.
```
rickhanlonii: > If your navigation library doesn't use them then you can end up with an app that doesn't have a native look and feel.

This doesn't seem accurate, if you don't use the platform features you can still have a native look and feel if you emulate the platform features. This section should include the advantages of not using the platform if it want's to be unbiased.

docs/navigation.md
```diff
@@ -3,72 +3,26 @@ id: navigation
...
 
-The community solution to navigation is a standalone library that allows developers to set up the screens of an app with just a few lines of code.
+## Has one React root
 
-The first step is to install in your project:
+If your navigation library renders all screens under a single React root then they share the React `Context`. That means you can use any state management library, or none, to pass data between them. If screens don't share the React `Context` then you have to hold the shared data in a global store outside of the React tree. That rules out state management solutions like Relay and Apollo and only rules in Redux.
```
rickhanlonii: This section should explain the benefits of having separate roots to remain unbiased

docs/navigation.md
```diff
@@ -3,72 +3,26 @@ id: navigation
...
-  }
-}
-```
-
-React Navigation routers make it easy to override navigation logic. Because routers can be nested inside each other, developers can override navigation logic for one area of the app without making widespread changes.
-
-The views in React Navigation use native components and the [`Animated`](animated.md) library to deliver 60fps animations that are run on the native thread. Plus, the animations and gestures can be easily customized.
-
-For a complete intro to React Navigation, follow the [React Navigation Getting Started Guide](https://reactnavigation.org/docs/getting-started.html), or browse other docs such as the [Intro to Navigators](https://expo.io/@react-navigation/NavigationPlayground).
+React Native for Web runs your native app on the Web. You don't have to change any code because React Native for Web renders your React Native components as DOM elements. For example, it converts Views into divs. But this only works if your navigation library supports React Native for Web. To avoid compromising the Web experience, make sure that your navigation library can generate Hyoerlinks and render on the server-side (SSR).
```
rickhanlonii: What is "React Native for Web"?

rickhanlonii:
@osdnk would you be able to review this as well?

osdnk:
Also, I see that these ticks and crosses are not black or white: 
e.g. RNN does not work with Expo? How about detaching?
RN does not work for web? PR is pending and it's matter of few weeks
RN does not use Native API? How about `react-native-screens`?

grahammendick:
Thanks @rickhanlonii  and @osdnk  for the great feedback. I wanted the documentation to be unbiased but I didn’t want it to be unopinionated. For example, I wanted to make it clear that React Router Native doesn’t pass the most basic test and so shouldn’t be used. But I can see that’s going to be too controversial.

Let’s change the format to match your suggestions. We can have a section for each library and reach out to library maintainers for their pitch/anti-pitch paragraphs. Are we all happy with that?

osdnk:
Good idea. I believe that the most important part is to explain motivations of these libraries, different approaches and problems solved with them.

docs/navigation.md
radex: > This doesn't seem accurate

But it is. I love `react-navigation` but the emulated IOS and Android navigation UI is merely an approximation or the real thing. It’s “good enough”, but it’s not perfect and you can really feel it.

I think we can tweak the wording to reflect it — it’s not that non-Native implementations suck, but it is a real tradeoff

brentvatne: I don't think this tradeoff is binary, it's not native or not native. For example, we used to use DrawerLayoutAndroid for the drawer navigator implementation, which I suppose you might call native navigation then. However, we found that it was buggy and not customizable enough so we re-implemented it with react-native-gesture-handler and react-native-reanimated. Going in the other direction, @kmagiera has explored building a set of primitives that navigation libraries build upon with react-native-screens. When he first prototyped it, he made a native stack that you could build react-navigation on top of. He has since returned to that and is exploring a backend for the React Navigation Stack Navigator that uses UINavigationController behind the scenes. The result, if it works out well, could be that you can have a fully customizable RNGH/Reanimated powered implementation and one that is driven by native primitives (although limited to a subset of the options).

brentvatne: The particular sentence here is also incorrect because you can opt-in to using react-native-screens with React Navigation by just doing `import { useScreens } from 'react-native-screens'; useScreens()` and your app will use [UIViewControllers](https://github.com/kmagiera/react-native-screens/blob/4dd751811addf7493bd355e5d557270c086979ed/ios/RNSScreen.m) and [Fragment]s(https://github.com/kmagiera/react-native-screens/blob/ed826d7b91acd84ced6fe933db74281fed469d69/android/src/main/java/com/swmansion/rnscreens/ScreenContainer.java) to represent screens.

docs/navigation.md
```diff
@@ -3,72 +3,26 @@ id: navigation
...
+|                                                                                                           | Uses native API | Has one React root | Runs in Expo | Runs on the Web |
+| --------------------------------------------------------------------------------------------------------- | --------------- | ------------------ | ------------ | --------------- |
+| [React Navigation](https://reactnavigation.org/docs/en/getting-started.html)                              | No              | Yes                | Yes          | No              |
+| [React Native Navigation](https://wix.github.io/react-native-navigation/)                                 | Yes             | No                 | No           | No              |
+| [The Navigation router](https://grahammendick.github.io/navigation/documentation/native/hello-world.html) | Yes             | Yes                | No           | Yes             |
```
brentvatne: I'm not sure how to say this without it coming off as unkind, so I am sorry @grahammendick, but this library should not be listed here. As far as I know nobody uses it in production. It may be reasonable to include it at some point in time but it needs to prove itself through usage.

cpojer: I agree. Can you remove it? I have never heard of this library before.

grahammendick: Thanks @brentvatne, no offence taken. I hope you won’t take any from what I have to say.

It’s interesting to see React gatekeeping out in the open. It’s obvious it goes on but it’s usually more subtle than that.

If you’re concerned about libraries proving themselves first then why does [React Navigation promote React Native Router](https://reactnavigation.org/docs/en/alternatives.html)? React Router Native doesn’t even have stack navigation because it doesn’t hold visited screens in memory. This is the lowest bar for any native navigation library.

Why didn’t Airbnb’s Native Navigation library have to prove itself? [It was promoted on this page](https://github.com/facebook/react-native-website/pull/532) for ages despite it not being production-ready and having threadbare documentation. The project’s been disbanded.

It’s obvious that React gatekeeping isn’t working. It lets through projects that should be turned away and closes the door to projects that should be welcomed. I wanted to rewrite this page as a corrective to the corrosive nature of React gatekeeping. Is that going to be possible?

cpojer: The current guidance for the React Native website is to list repositories that are either part of the RN community org ( https://github.com/react-native-community ) or have significant adoption. If any library part of the React Native ecosystem could be listed on the website it will confuse people with too many packages and increase maintenance burden for us to check which packages work with the latest RN releases.

grahammendick: @cpojer Can you explain what checks you run? Do you plan to update this page to say that React Native Navigation doesn't work with RN 0.60?

cpojer: As you pointed out yourself, even with a small number of libraries mentioned on the website it is still hard to ensure they are all working with the latest versions of React Native. Imagine increasing the amount of libraries mentioned significantly and how that would make people lose trust in the RN documentation.

React Native Navigation is currently being made ready for 0.60: https://github.com/wix/react-native-navigation/issues/5228

docs/navigation.md
danilobuerger: Its untrue that you can't use Apollo if you have multiple roots like in react-native-navigation. Of course you can share a context with multiple roots.

docs/navigation.md
brentvatne: @grahammendick - there is no gatekeeping happening here, we are just trying to provide a good initial experience for new users and I don't think that guiding them towards an unproven library is a responsible thing to do. I'd appreciate a bit more empathy for our situation rather than assuming that folks are acting in bad faith to keep certain libraries in favor and limit the popularity of others. I'll repeat to you again - as a maintainer of React Navigation, I would love for another library to come along and be better in every way and replace it, then the React Native community would be better and I would have less work to do. I encourage you to keep working on your library, kick the tires in some production apps and get some early adopters.

satya164: > If you’re concerned about libraries proving themselves first then why does React Navigation promote React Router Native

IMO we should remove it from there

satya164: > Why didn’t Airbnb’s Native Navigation library have to prove itself? It was promoted on this page for ages despite it not being production-ready and having threadbare documentation. The project’s been disbanded.

It was backed by a big company which already used it in production. So it was fair to assume that they'll keep maintaining it. We can't predict future and we make mistakes in our assumptions.

But I feel it doesn't matter now since we've already removed it when we got to know that it's not maintained anymore.

docs/navigation.md
satya164: I think we shouldn't include React Router Native. It's not really feature-complete enough to use for production apps, imo.

docs/navigation.md
grahammendick: These mistakes only ever happen in one direction. They always promote the powerful. Companies and people whose voices are already amplified in the React community. This is what I mean by gatekeeping @brentvatne . I don’t think you’re acting in bad faith, I think you’re suffering from confirmation bias.

Thanks @satya164  for responding to my specific points. I’ll be very happy if you remove React Router Native because It will help prevent users taking a wrong turn.

brentvatne: https://github.com/react-navigation/react-navigation.github.io/commit/921b263d37649f05beb13ca6e2988d457071ae9b

sahrens: > They always promote the powerful. 

I think you are mixing correlation and causation. We try to promote production battle-tested libraries used by many engineers. There is a correlation with those libraries and powerful organizations because powerful organizations usually have lots of engineers and users doing that battle testing. But it's entirely possible for a lone developer to come up with an awesome library, and grow enough early adopters to prove it out, but they need to seed that growth themselves.

osdnk:
I'm afraid it's not a good place for this discussion. Can you, @grahammendick, open RFC in https://github.com/react-native-community/discussions-and-proposals/issues and sum up some thoughts. I'll happily follow up with come more general comments and thoughts, but now I'm feeling that every comment here is considered as a "review" for your changes. Let's talk about overview of navigation patterns with any bias.

osdnk:
And for now, can we close this PR? Don't get me wrong, @grahammendick, but I'm afraid we're not ready to merge in current state even with every of suggestions here included.

grahammendick:
No problem @osdnk. Thanks all for the suggestions

PR #1088, opened 2019-07-09
Fix anchor links to component props and methods

mileskrell:
Most of these were just caused by the link not being all in lowercase, e.g. `datepickerios.md#initialDate` instead of `datepickerios.md#initialdate`.

There were also a few links for Button that were erroneously linking to View props with the same names.


charpeni:
ToC has been removed.

PR #1090, opened 2019-07-11
Update pushnotificationios.md

anfriis:
Just tried the guide with RN 0.60.0 and the `PushNotificationIOS` library was automatically linked in the `Pods` project, so it cannot build if the library is also manually linked in the main project.


charpeni:
This is conflicting, can you verify if the current information on Pods and (your PR) automate linking is true?

anfriis:
It seems that another PR changing the same docs was merged before this one. 
However I think my point about RN 0.60.0 doing everything for you with autolink is worth adding, as well as the instruction for manual and CocoaPods linking for versions below 0.60.0

PR #1091, opened 2019-07-11
delete deprecated duration prop from ScrollView

ecreeth:
The `duration` prop was added in [#22884](https://github.com/facebook/react-native/pull/22884). But was revert [here](https://github.com/facebook/react-native/commit/4c69ccd0fb55e994bd4604e5194206e442ec5958#diff-f8f8b220fc0d1573e4f8b78c84415075).

See the issue [#25589](https://github.com/facebook/react-native/issues/25589) in the React Native repository.

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -718,23 +718,17 @@ Displays the scroll indicators momentarily.
 
 ```javascript
 scrollTo(
-  ([y]: number),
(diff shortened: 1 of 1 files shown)


ccfz:
The documentation still mentions the duration. Is that an accident or was this not yet deployed to the website?

PR #1104, opened 2019-07-15
Update Style.MD

nduevans:
This example is more straight forward for a new learner. The conflicting attributes(e.g color) in the current example where you have two styles modifying the same text  is confusing. In last TEXT, red is not rendered for example.


rachelnabors:
This section is about to get an overhaul... and the cascade of CSS likely to get covered. So removing it now would not make sense. Thank you for your suggestion, though! Well be sure to dig in an confuse learners less ;)

PR #1111, opened 2019-07-16
Change to export default class

niedem:
Old code did no longer work, in the new react-native versions this is supposed to be used


charpeni:
Thanks, this has been fixed on master, closing it.

PR #1113, opened 2019-07-16
Added instruction to install node modules and pods

akshaygore06:


akshaygore06:
@cpojer Can you please review this PR? Thanks

thymikee:
Pods are installed by default. If they fail, CLI should tell you about it. @charpeni this can be closed

PR #1117, opened 2019-07-17
Remove `openSettings` from docs for v0.59

Skn0tt:
"openSettings" is only available from v0.60 on.
See https://github.com/facebook/react-native/blob/0.59-stable/Libraries/Linking/Linking.js.


charpeni:
Thanks, can you also remove it from the table of contents?

PR #1136, opened 2019-07-22
Clean up usage of AppRegistry in examples

brentvatne:
Currently we include `AppRegistry.registerComponent(...)` in basic React Native Web Player examples, but we don't need to do this since the player automatically registering the default exported component. This is also the case for apps you initialize through `react-native init` -- `App.js` just does `export default class App..` and `AppRegistry` is invoked in `index.js`. This PR also removes the comment about "Create React Native App" which no longer is a thing.


brentvatne:
thx @rickhanlonii

PR #1138, opened 2019-07-23
Update showcase.json

Ishaan28malik:
I was going through the website and as I scrolled all the way down to the bottom , I observed  that out of 12 logos of the React Native users , three didn't show any updates on mouseovers . So just checked in with the source code and Updated it .

I think this was necessary because it doesn't seems nice to the viewer .
**### Referring to this** -

**`<div class="logos"><div><div class="showcase"><a</div>`**

An alternative can be till the time there is someone writes blogs for these three users , You can replace them with the any three users like 
Vogue , Soundcloud , Discord etc

Hope this is useful for the community and devs.

### DIFF
--- a/website/showcase.json
+++ b/website/showcase.json
@@ -28,6 +28,9 @@
     "linkAppStore": "https://itunes.apple.com/us/app/facebook/id1266461465",
     "linkPlayStore":
       "https://play.google.com/store/apps/details?id=com.facebook.analytics",
+    "infoLink":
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Thanks, this is a great first contribution to the website!

Ishaan28malik:
@brentvatne , @rickhanlonii What should I work upon next ? Can you assign me any task ..

PR #1139, opened 2019-07-23
Remove Expo SDK version mapping for embeds and use latest instead

brentvatne:
The Expo SDK version map for embeds isn't being kept up to date and I believe it makes most sense / is consistent with the React Native Web Player to just use the latest version all the time.

## Before

Screen Shot 2019-07-22 at 16 39 37

## After

Screen Shot 2019-07-23 at 10 45 47


brentvatne:
thx @rickhanlonii

Issue #1131, opened 2019-07-19
use [snack.io][WEB SUPPORT] instead of react-native-web-player


ecreeth:
As shown in the following image, snack.io now has web support.

The reason for this change is that the examples in the `snack` player are more similar to what they would look like on a real device.

![image](https://user-images.githubusercontent.com/20761166/61547097-b1c62280-aa18-11e9-89f0-7e901eb05297.png)

cc @hramos

hramos:
This seems fine to me. Are all existing uses of the react-native-web-player well supported when switched to use the Snack Web player?

ecreeth:
I'm not 100% sure, I've tried all the examples in the "The Basics" section and they all work. Maybe @brentvatne can clarify this for us.

brentvatne:
yeah they should work fine! 

@ecreeth - can you open a PR for this so we can test it out with the automatically deployed netlify preview? (eg: https://github.com/facebook/react-native-website/pull/1142#issuecomment-514781241)

ecreeth:
Of course!! :100: thanks for the clarification

ecreeth:
the PR was created. I have noticed that the web player option does not appear. Any ideas
![imagen](https://user-images.githubusercontent.com/20761166/61841721-a84b1900-ae63-11e9-9db5-c465ad79cc13.png)

tcdavis:
@ecreeth I believe this needs to be bumped to 33.0.0

ecreeth:
@brentvatne it seems that the snack Embed Preview does not have the option of the web player
Thank you @tcdavis, once you sent the message I thought that could be the problem
![image](https://user-images.githubusercontent.com/20761166/61907761-20fdb400-aefc-11e9-9e58-e5af8d956daf.png)

tcdavis:
@ecreeth Sorry about that.  Embeds should have the web preview now

PR #1143, opened 2019-07-25
use [snack.io][PLAYER] instead of react-native-web-player 

ecreeth:
### Use `Snack.io` player instead of the `react-native-web-player` in the following sections.

Related issue #1131 

- [x] The Basics
- [x]  Guides
- [x] Guides (IOS)
- [x] Guides (Android)
- [x] Components
- [x] API

### DIFF
--- a/docs/activityindicator.md
+++ b/docs/activityindicator.md
@@ -7,7 +7,7 @@ Displays a circular loading indicator.
 
 ### Example
 
-```ReactNativeWebPlayer name=activityindicator&platform=web
+```SnackPlayer name=activityindicator&platform=web
 import React, { Component } from 'react'
 import {
   ActivityIndicator,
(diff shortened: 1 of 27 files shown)


brentvatne:
hmm that deploy preview seems wrong 🤔

brentvatne:
hey @ecreeth, thanks for doing this! could you change SnackPlayer embeds to use `web` by default when they make sense on web? eg: most of the stuff in the the "basics" section except where we used SnackPlayer prior to this PR would make more sense with the `web` default.

to do this you can add it as an additional param for the code block:

```
```SnackPlayer name=Hello%20World&platform=web
```

ecreeth:
Thank you for this! I was about to do it but I didn't know how 😄

brentvatne:
looks good! @ecreeth - can you backport this to the 0.60 docs too? then I think it's ready to go!

brentvatne:
hey @ecreeth! I rebased against master and changed every instance of ReactNativeWebPlayer to SnackPlayer in https://github.com/facebook/react-native-website/pull/1174, so let's continue there. Your commit attribution of course remains in place!

ecreeth:
Great!!

PR #1145, opened 2019-07-26
Remove duplicate activeOpacity description

amandeepmittal:
The `activeOpacity` prop is described twice in the document. This PR resolves the duplication.


amandeepmittal:
I've closed this PR since there is the same which resolves this issue. #1129

PR #1152, opened 2019-07-29
Update toastandroid.md

ecreeth:
This is part of #929 

> The properties of this component has not changed

### was added
- The [`ToastAndroid`](https://github.com/facebook/react-native/blob/master/Libraries/Components/ToastAndroid/ToastAndroid.android.js#L36) properties definitions


rachelnabors:
Thanks for updating these!

PR #1159, opened 2019-07-31
Updates the JS environment

orta:
Think I got most of the new additions from the [metro defaults](https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/package.json)


orta:
good point, fixed

PR #1160, opened 2019-07-31
Added a first pass at TS docs for React Native

orta:
And here's the PR all the others have been working up to, this adds a quick overview on how to:

- start with TypeScript
- add TypeScript to an existing project
- talks about how it works via babel
- gives an example to look at
- gives some extra context on where to look at next

Open to feedback on what else could be in here?

I also added myself as an auto-reviewer to this file which should make it easy for me to validate changes if it's complex 👍

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -1,0 +1,186 @@
+---
+id: typescript
+title: Using TypeScript with React Native
+---
+
+[TypeScript][ts] is a language which extends JavaScript to add type definitions, a lot like [Flow][flow] which React Native is built in. React Native supports TypeScript and Flow out of the box by default.
+
+## Getting Started with TypeScript
+
+If you're starting a new project, you can use the [TypeScript template][ts-template] to get you there faster:
+
+```sh
+react-native init MyAwesomeTSProject --template typescript
+```
+
+Or if you're using expo:
+
+```sh
+npm install -g expo-cli
+expo init MyAwesomeTSProject
+```
+
+and pick one of the TypeScript templates.
+
+## Adding TypeScript to an Existing Project
+
+1. Add TypeScript, and the React Native and Jest types to your project. When you were using JavaScript these could have been added for you by your editor.
+
+```sh
+yarn add typescript @types/jest @types/react @types/react-native @types/react-test-renderer
+# or for npm
+npm install --save-dev @types/jest @types/react @types/react-native @types/react-test-renderer
+```
+
+2. Add a TypeScript config file. Create a `tsconfig.json` in the root of your project:
+
+```json
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "isolatedModules": true,
+    "jsx": "react",
+    "lib": ["es6"],
+    "moduleResolution": "node",
+    "noEmit": true,
+    "strict": true,
+    "target": "esnext"
+  },
+  "exclude": [
+    "node_modules",
+    "babel.config.js",
+    "metro.config.js",
+    "jest.config.js"
+  ]
+}
+```
+
+3. Create a `jest.config.js` file to configure Jest to use TypeScript
+
+```js
+module.exports = {
+  preset: 'react-native',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+};
+```
+
+4. Rename a JavaScript file to be `*.tsx`
+
+5. Run `yarn tsc` to type-check your new TypeScript files.
+
+## How TypeScript and React Native works
+
+Transforming your files to JavaScript works via the same [Babel infrastructure][babel] as vanilla JavaScript, this means that you should not rely on TypeScript to transform your `ts` and `tsx` files, and it's worth noting there are [one or two caveats][babel-7-caveats] if you have existing TypeScript code to work with.
+
+## What does React Native + TypeScript look like
+
+You can provide an interface for a React Component's [Props][props] and [State][state] via `React.Component<Props, State>` which will provide type-checking and editor auto-completing when working with that component in JSX.
+
+```ts
+// components/Hello.tsx
+import React from 'react';
+import {Button, StyleSheet, Text, View} from 'react-native';
+
+export interface Props {
+  name: string;
+  enthusiasmLevel?: number;
+}
+
+interface State {
+  enthusiasmLevel: number;
+}
+
+export class Hello extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    if ((props.enthusiasmLevel || 0) <= 0) {
+      throw new Error('You could be a little more enthusiastic. :D');
+    }
+
+    this.state = {
+      enthusiasmLevel: props.enthusiasmLevel || 1,
+    };
+  }
+
+  onIncrement = () =>
+    this.setState({enthusiasmLevel: this.state.enthusiasmLevel + 1});
+
+  onDecrement = () =>
+    this.setState({enthusiasmLevel: this.state.enthusiasmLevel - 1});
+
+  getExclamationMarks = (numChars: number) => Array(numChars + 1).join('!');
+
+  render() {
+    return (
+      <View style={styles.root}>
+        <Text style={styles.greeting}>
+          Hello{' '}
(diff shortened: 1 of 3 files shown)


iRoachie:
Can possibly link to the typescript docs on the reactjs website on typescript https://reactjs.org/docs/static-type-checking.html

And/or maybe a small tidbit on DefinitelyTyped

docs/typescript.md
```diff
@@ -0,0 +1,186 @@
...
+  ]
+}
+```
+
+3. Create a `jest.config.js` file to configure Jest to use TypeScript
+
+```js
+module.exports = {
+  preset: 'react-native',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
```
thymikee: Jest has it by default, however the order is different so this may be a bit more performant in resolving ts files

JReinhold:
Looks great, I have a few comments:

1. If we are giving examples on TypeScript usage, I think we should also have an example on functional components as well, using the `React.FC` type. Especially now, that class components are getting out of style.
2. It could also be an idea with examples on how to type the built-in hooks, but one could argue that that would be too much.
3. Another advanced topic, that I and others have had trouble with, is making path aliases work (again, maybe they don't need to be in the docs, depending on how comprehensive you want to be). Specifically making path aliases that point to folders outside the `src` directory work, like in a monorepo setup. I THINK it works out of the box for path aliases that points to internal directories under `src`, but I'm unsure. This both needs to be addressed with the Metro bundler and with Jest, and possibly in the babel config.

docs/typescript.md
orta: I thought this too, but the example test doesn't pass when you have it removed

```
orta@Perl ~/d/r/t/js> mv jest.config.js _something
orta@Perl ~/d/r/t/js> yarn jest
yarn run v1.17.3
$ /Users/orta/dev/react-native/tmp/js/node_modules/.bin/jest
 FAIL  __tests__/App-test.js
  ✕ renders correctly (15ms)

  ● renders correctly

    Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.

      at invariant (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:55:15)
      at createFiberFromTypeAndProps (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2056:11)
      at createFiberFromElement (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2077:15)
      at reconcileSingleElement (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:4629:23)
      at reconcileChildFibers (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:4686:35)
      at reconcileChildren (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:6508:28)
      at updateHostRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:6920:5)
      at beginWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:7745:14)
      at performUnitOfWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11413:12)
      at workLoop (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11445:24)

  console.error node_modules/react/cjs/react.development.js:188
    Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object.
```

orta:
@JReinhold  1 & 2 - I don't know all the new React syntax, so I followed a tutorial and probably got it right.

3 - I was originally going to say "I can link to a definitive source" but there wasn't one I could find, so I've added it here. 

@iRoachie - added, thanks

docs/typescript.md
SimenB: That should most definitely be fixed...

docs/typescript.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Getting Started with TypeScript
+
+If you're starting a new project, you can use the [TypeScript template][ts-template] to get you there faster:
+
+```sh
+react-native init MyAwesomeTSProject --template typescript
+```
+
+Alternatively, you can use some of the community bootstrapping tools which can get you to a complete app faster:
```
brentvatne: small nit on wording - is `react-native-cli` not a community bootstrapping tool as well? https://github.com/react-native-community/cli

brentvatne:
looks great :+1:

JReinhold:
@orta 

The functional component example looks good! 💪

However I'm unsure if the path alias example is exhaustive, I don't think it will cover a case, where we want to reference modules outside the root directory (which is very common in a monorepo structure).Taking an example structure like below:

```
├── react-native-ts-app
│   ├── package.json
│   ├── src
│   │   └── index.tsx
│   ├── test
│   │   └── some.test.ts
│   └── tsconfig.json
└── shared
    └── some-util.ts
```

I'm concerned that the docs right now will only work, as long as the paths reference something inside `react-native-ts-app`, and won't work when importing from the `shared` folder.
My experience is that both the metro bundler and jest needs to be told about the `shared` folder as well.

I've wanted to test out if your documentation actually solves this problem, but I haven't had time yet to investigate it, so I thought I would at least share my concern now, to not slow down the PR.
As soon as I get time, I will try to see if your docs mitigates the issue I'm describing. Anyone is of course welcome to beat me to it.

Again, I might be wrong, and your docs might be perfectly exhaustive.

docs/typescript.md
orta: Hah, I guess it is given that it lives in there

orta:
@JReinhold - I think you're welcome to expand on that once it's merged 👍

ashfurrow:
I really like these docs – succinct enough to be helpful without being overwhelming, with lots of links to further resources. Looks great!

docs/typescript.md
```diff
@@ -0,0 +1,222 @@
...
+
+const Hello: React.FC<Props> = (props) => {
+    const [enthusiasmLevel, setEnthusiasmLevel] = React.useState(props.enthusiasmLevel);
+
+    const onIncrement = () => setEnthusiasmLevel((enthusiasmLevel || 0) + 1);
+    const onDecrement = () => setEnthusiasmLevel((enthusiasmLevel || 0) - 1);
+
+    const getExclamationMarks = (numChars: number) => Array(numChars + 1).join('!');
+    return (
+        <View s style={styles.root}>
```
iRoachie:
```suggestion
        <View style={styles.root}>
```

orta:
OK, this should be good to go

docs/typescript.md
```diff
@@ -3,33 +3,33 @@ id: typescript
 title: Using TypeScript with React Native
 ---
 
-[TypeScript][ts] is a language which extends JavaScript to add type definitions, a lot like [Flow][flow] which React Native is built in. React Native supports TypeScript and Flow out of the box by default.
+[TypeScript][ts] is a language which extends JavaScript to add type definitions, much like [Flow][flow]. While React Native is built in Flow, it supports both TypeScript _and_ Flow by default.
```
iRoachie: Seems to be more confusing than what was previously there.

`React Native supports TypeScript and Flow out of the box by default.` was clearer

docs/typescript.md
```diff
@@ -0,0 +1,220 @@
...
+};
+```
+
+4. Rename a JavaScript file to be `*.tsx`
+
+5. Run `yarn tsc` to type-check your new TypeScript files.
+
+## How TypeScript and React Native works
+
+You should not rely on TypeScript to transform your `ts` and `tsx` files into JavaScript! Transforming your files to JavaScript works via the same [Babel infrastructure][Babel] as vanilla JavaScript. If you have existing TypeScript code, there are [one or two caveats][babel-7-caveats].
```
rachelnabors: Could we rephrased "Transforming your files to JavaScript works via the same Babel infrastructure as vanilla JavaScript" to be a bit clearer? For folks not familiar with how Babel works with vanilla JS (should we even still be calling it "vanilla JS?" Can we use a different phrase?)

docs/typescript.md
rachelnabors: "[TypeScript][ts] is a language which extends JavaScript to add type definitions, much like [Flow][flow]. React Native is built in Flow, and supports both TypeScript _and_ Flow by default."

orta:
This still needs merging :D

PR #1063, opened 2019-06-27
Add `upgrade-helper` as an upgrading option

lucasbento:
This PR is related to react-native-community/upgrade-helper#38 and adds [Upgrade Helper](https://github.com/react-native-community/upgrade-helper) tool as an option to the `Upgrading to new React Native versions` documentation. It also:

1. Puts "how to automatic upgrade with CLI without Git" in a troubleshooting section;
1. Recommends [react-native-clean-project](https://github.com/pmadruga/react-native-clean-project) as a way to clear project cache after you upgrade but have encountered errors with version mismatch;
1. Uses `yarn` instead of `npm` when asking for installing dependencies.

The second one is done mostly because there was work done by @mikehardy on turning the package into a plugin for the CLI so the user can simply run `react-native clean-project`.

I'm very much open for feedback on how to improve this.

PR #1161, opened 2019-07-31
Add upgrade docs to `0.60`

lucasbento:
This PR is the same as #1063 but for the main `upgrading.md` documentation.

Apparently there's an issue on how to fetch the correct file, for `0.59` the `upgrading.md` that lives under `version-0.5` but for `0.60` it's picking up the one under `docs`.

For now I'm fixing this as it is and we should have a plan to fix it in a proper way in the source.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -1,7 +1,6 @@
 ---
-id: version-0.5-upgrading
+id: upgrading
 title: Upgrading to new React Native versions
-original_id: upgrading
 ---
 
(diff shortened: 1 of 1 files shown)


lucasbento:
@charpeni: all fixed!

> Apparently there's an issue on how to fetch the correct file, for 0.59 the upgrading.md that lives under version-0.5 but for 0.60 it's picking up the one under docs.

Can you confirm if this is an actual issue?

PR #1163, opened 2019-08-01
Remove linking for gesture handler

JullietRUS:
As from 0.60 modules has to be linked automatically.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -27,12 +27,6 @@ yarn add react-native-gesture-handler
 # npm install --save react-native-gesture-handler
 ```
 
-Now we need to link our react-native to react-native-gesture-handler
-
-```
-react-native link react-native-gesture-handler
-```
-
 Then you can quickly create an app with a home screen and a profile screen:
 
 ```jsx


rickhanlonii:
Nice, thanks! Are there any other `react-native link` references in the docs we should remove? Can you also update this in the versioned_docs for 0.60?

JullietRUS:
> 
> 
> Nice, thanks! Are there any other `react-native link` references in the docs we should remove? Can you also update this in the versioned_docs for 0.60?

Hi, if I'll find some other 'react-native link', i'll let you know.

But wait a moment, where is not only to remove linking, but update an article with this (I made it in my own app based on react-native v0.60.4 and react-native-gesture-handler 1.3.0 for make gesture handler works):
**------------------------------------------------**
You don't need to execute "react-native link", but you need to add on "app/src/main/java/<your_path/>MainActivity.java:
```
import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;

public class MainActivity extends ReactActivity {
...
    @Override
    protected ReactActivityDelegate createReactActivityDelegate() {
      return new ReactActivityDelegate(this, getMainComponentName()) {
        @Override
        protected ReactRootView createRootView() {
         return new RNGestureHandlerEnabledRootView(MainActivity.this);
        }
      };
    }
...
}
```
------------------------------------------------

PR #1170, opened 2019-08-05
Tweak home page example

gaearon:
1. Remove unused `Component` export.
2. `export default () => ...` is not recommended because it doesn't get a name. Changing this to an explicit assignment. Happy to change to `App` or whatever but this seemed clear enough.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -23,11 +23,11 @@ one team can maintain two platforms and share a common technology—React.
   `,
   codeExample: `
 \`\`\`jsx
-import React, {Component} from 'react';
+import React from 'react';
 import {Text, View} from 'react-native';
 import {Header} from './Header';
 
-export default () => 
+const WelcomeScreen = () => 
   <View>
     <Header title="Welcome to React Native"/>
     <Text style={header}>Step One</Text>


noway:
While we are at it, can we define `header` variable which is used in `style={header}`s too? It's not defined, the example won't run 😅

gaearon:
>While we are at it, can we define header variable which is used in style={header}s too? It's not defined, the example won't run

What is it supposed to be?

orta:
It can just be a fake import IMO `import {headers} from "./style"` or something like that maybe?

noway:
agree with @orta, otherwise 
In https://github.com/facebook/react-native/blob/master/template/App.js it is defined as
```javascript
  sectionTitle: {
    fontSize: 24,
    fontWeight: '600',
    color: Colors.black,
  },
```
so it could be
```javascript
  const header = {
    fontSize: 24,
    fontWeight: '600',
    color: 'black',
  }
```
(replacing color to a string literal so that no extra dependency needed)

orta:
Please don't _actually_ put the styles in - the point is that it looks and feels like the above screenshots. That it could be the implementation. Not that it is.

kdawgwilk:
Do we want to mention "Fast Refresh" that isn't actually released yet?

PR #1177, opened 2019-08-07
Improve embedded examples

brentvatne:
This PR does a handful of things to improve the experience of using embedded Snack examples:

- Fixes a reference to an image that no longer exists (`favicon.png`), I added an equivalent image and called it `tiny_logo.png` (this won't work on the Netlify preview because the example points to the production site for the image URL)
- The default for SnackPlayer is now to use web. The only exception to this currently is for `InputAccessoryView`, which doesn't really make sense on web - it only applies to iOS.
- Use global `alert` rather than `Alert.alert` in handling touches example, this works on web whereas `Alert.alert` does not.
- Fix broken import for Image component. It was previously also broken with ReactNativeWebPlayer because the `favicon.png` file was removed, but also it depended on importing from a path relative to the docs page (`require('/react-native/img/favicon.png')`) rather than relative to file system. I replaced it with a static image that is hardcoded into Snack for this purpose because it's not currently possible to add files to a Snack workspace for an inline-embedded Snack (it is possible if you save the code on snack.expo.io and then embed that Snack, but that makes it trickier to maintain the docs).
- Make the Image style example actually stretch the image (using `resizeMode`) since the style is called 'stretch' and it seems to make sense with the example.
- Improving some styling of embeds in a couple of places (eg: direct manipulation)


brentvatne:
Unclear to me what is up with CI here, it seemingly can't find the branch to check out in the setup job even though it clearly exists :P Unrelated to the code in this PR.

Issue #1216, opened 2019-09-03
Height And Width Flex Playground Not Working


alexfinnarn:
<img width="830" alt="Screen Shot 2019-09-03 at 10 59 49 AM" src="https://user-images.githubusercontent.com/3640707/64184825-11965280-ce3a-11e9-9302-e695d837e3d0.png">

There appears to be some faulty Markdown causing an issue with rendering. Oddly, I swear this worked for me a few days ago and that document says it hasn't been updated in 28 days.

alexfinnarn:
Not sure if the four tick marks means something special like a directive in Docosaurus, but usually code blocks are three ticks.

charpeni:
This is working fine on my computer, at least. Can you try in Guest or Private browsing mode?

alexfinnarn:
I use the Brave browser, but in private mode still the same thing for me. I don't have Chrome installed so I can't say how that looks, but I have had issues rendering things with Chromium-based browsers that aren't Chrome before.

In Safari, the preview doesn't even show up for me.

<img width="795" alt="Screen Shot 2019-09-04 at 5 51 04 PM" src="https://user-images.githubusercontent.com/3640707/64294362-9bc6df80-cf3c-11e9-83d4-ed15a2f4aafa.png">

I feel like this wasn't the case last week with the same browser. I'm not sure how often you run builds but looks like something changed.

PR #1217, opened 2019-09-03
Issue #1216: Remove Faulty Markdown Syntax

alexfinnarn:
If your PR references an existing issue, please add the issue number below:
#1216


alexfinnarn:
Still doesn't look right in the preview so maybe it is four ticks?

charpeni:

PR #1219, opened 2019-09-03
imp: Snack examples for DatePickerIOS and DrawerLayoutAndroid

Esemesek:
### Description

Some more Snack examples, this time for `DatePickerIOS` and `DrawerLayoutAndroid`


rachelnabors:
Issue, on DrawerLayoutAndroid's web preview I get this error:
```
Cannot read property 'Left' of undefined
TypeError: Cannot read property 'Left' of undefined
    at App (module://App.js.js!transpiled:32:64)
    at Xo ([snack internals]
    at Qi ([snack internals]
    at za ([snack internals]
    at Ja ([snack internals]
    at Ds ([snack internals]
    at _s ([snack internals]
    at Es ([snack internals]
    at qa ([snack internals]
    at Object.enqueueSetState (https://s3.us-west-1.amazonaws.com/snack-web-player-staging/34/static/js/2.0ff5cc17.chunk.js:1:1000231)
```

Esemesek:
@raajnadar This is `Android` specific module.

raajnadar:
@Esemesek ??

Esemesek:
@raajnadar sorry for that. I meant to tag @rachelnabors . Stupid autocomplete.

rachelnabors:
Given that this is Android-only, I wonder if there's any way to not show web and iOS in the preview? cc @brentvatne

brentvatne:
it's not. possible to disable it on other platforms currently, i'm following up with @tcdavis on this because i think it could be a valuable option to have. currently a workaround is to just write the code such that it will tell it's not supported on unsupported platforms, eg: https://snack.expo.io/@notbrent/platform-unsupported-example?platform=android

rachelnabors:
@brentvatne Hmm, that workaround could be more confusing than useful. It would be better to wait to exclude those. (Our other option is to just not use Snacks in the components docs, but they solve problems for us like letting users see a visual of each component.)

Also wondering how we'd represent these variations in Markdown!

rachelnabors:
Handled in #1579

PR #1227, opened 2019-09-04
Update animations.md

antronic:
Fix typo in an Animated API's example

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -36,7 +36,7 @@ const FadeInView = (props) => {
     <Animated.View                 // Special animatable View
       style={{
         ...props.style,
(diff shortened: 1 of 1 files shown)


docs/animations.md
```diff
@@ -36,7 +36,7 @@ const FadeInView = (props) => {
     <Animated.View                 // Special animatable View
       style={{
         ...props.style,
-        opacity: fadeAdnim,         // Bind opacity to animated value
+        opacity: fadeAdmin,         // Bind opacity to animated value
```
charpeni: You made me realized that it should be `fadeAnim`, not `fadeAdmin`. 

Are you up to replace it elsewhere as well?

PR #1234, opened 2019-09-04
imp: Snack example for Modal and Picker

Esemesek:
### Description

Yet another PR with snack examples. C:


Esemesek:
@charpeni It works on Android and iOS simulators. It seems like there is no Modal implementation for web, so it is replaced with `div`.

Esemesek:
I can add `Platform` check to this one, and on `web` render some kind of message.

charpeni:
Hmm yeah. Related to #1219. This seems odd to me to default to web with those unsupported components 🤔 

Not sure what will be the best solution. cc @rachelnabors.

Esemesek:
I think we are able to pass a parameter to our RemarkablePlugin. https://github.com/facebook/react-native-website/blob/master/website/core/RemarkablePlugins.js#L85 . I'll try to fix it tomorrow.

docs/picker.md
```diff
@@ -5,16 +5,39 @@ title: Picker
...
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  picker: {
+    width: 200,
+    height: 300,
```
ecreeth: Maybe **height: 30** :smile:

Esemesek: Thanks. It makes more sense for it to me 30 :P

rachelnabors:
Folks, what's the status on this one? Looks like modal on web still isn't working.

cpojer:
@Esemesek can you revert the change to Modal?

cpojer:
Actually I just merged https://github.com/facebook/react-native-website/pull/1433 which adds the snack player to Picker, so let's just close this PR for now and take care of Modal later :)

PR #1236, opened 2019-09-04
Fix fadeAdmin typo

giautm:


docs/animations.md
```diff
@@ -36,7 +36,7 @@ const FadeInView = (props) => {
     <Animated.View                 // Special animatable View
       style={{
         ...props.style,
-        opacity: fadeAdnim,         // Bind opacity to animated value
+        opacity: fadeAdmin,         // Bind opacity to animated value
```
charpeni: It should be `fadeAnim`.

PR #1238, opened 2019-09-04
Bad variable name leading code example to fail.

stardisblue:
Array destructuring `[fadeAdmin]` instead of `[fadeAnim]` in the SnackPlayer code example.

Renaming to `fadeAnim` in the example.


stardisblue:
closing because of dupe

Issue #1151, opened 2019-07-29
Add in docs that "location" is now an extracted module


giacomocerquone:
We should add in this page: https://facebook.github.io/react-native/docs/geolocation
that the module is extracted here (and so it's not global anymore): https://github.com/react-native-community/react-native-geolocation

If someone gives me the ok, I'll prepare the PR

abhinavsagar:
Go for it.

giacomocerquone:
@abhinavsagar sorry but I can't find the geolocation md file inside the docs folder but I can't find it under versioned_docs.
What should I do? As you see under 0.60 the geolocation api is still there: https://facebook.github.io/react-native/docs/geolocation

Simek:
The deprecation note has been added in the PR #1673 so the reported issue should be fixed now.

PR #1239, opened 2019-09-05
Create geolocation.md

Hyllesen:
For some reason the geolocation.md file was no longer in repo, but was up on the website mentioning the now outdated navigator global.
#1151

### DIFF
--- a/docs/geolocation.md
+++ b/docs/geolocation.md
@@ -1,0 +1,166 @@
+---
+id: version-0.60-geolocation
+title: Geolocation
+original_id: geolocation
+---
+
+The Geolocation API extends the [Geolocation web spec](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation).
+
+This API is available through the [react-native-geolocation](https://github.com/react-native-community/react-native-geolocation) module.
+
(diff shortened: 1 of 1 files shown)


jamonholmgren:
I agree with @radko93 -- we need to point them to the community module which has been extracted.

giacomocerquone:
I authored the #1151 issue.
What warning are you talking about? @radko93 Do you mean a warning while developing when requesting the old module? I'd say that sounds good too, but this is the react-native-website repo and I was asking for a way to include it in the docs and there are a couple of things I'm not getting.

1. How can that page be up if it's missing in this repo? Isn't the whole website built with netlify from this repo?
2. Where can I edit it then?

radko93:
@giacomocerquone look at the warning here https://facebook.github.io/react-native/docs/webview

giacomocerquone:
oh yes ok, thank you, I was just a little bit confused! I've also seen that this pr brings that warning! So nothing to do here 🙂

cpojer:
Geolocation was removed intentionally as part of Lean Core. We should add it to the list of community modules, which should be done in a separate PR.

PR #1241, opened 2019-09-05
Correction in Animation: fadeAdmin example

ivp199:
The code is breaking and giving the error about fadeAdmin. Corrected the issue.

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -36,7 +36,7 @@ const FadeInView = (props) => {
     <Animated.View                 // Special animatable View
       style={{
         ...props.style,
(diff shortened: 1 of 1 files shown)


docs/animations.md
```diff
@@ -36,7 +36,7 @@ const FadeInView = (props) => {
     <Animated.View                 // Special animatable View
       style={{
         ...props.style,
-        opacity: fadeAdnim,         // Bind opacity to animated value
+        opacity: fadeAdmin,         // Bind opacity to animated value
```
charpeni: It should be `fadeAnim`.

PR #1247, opened 2019-09-07
imp: Snack example for SectionList

Esemesek:
### Example

Snack example for SectionList component.


charpeni:
Thanks for doing this @Esemesek!!

PR #1252, opened 2019-09-09
Update height-and-width.md incorrect code block

LuigiQuarta:
Little fix in the last code block, the code incorporated the last line of text and the example didn't worked.


charpeni:

Issue #968, opened 2019-05-30
Allow me to link to the native docs steps, not the expo one


brunolemos:
I'd like to be able to send a link to a beginner (e.g. https://facebook.github.io/react-native/docs/getting-started.html#native) with the `React Native CLI Quickstart` tab open, not the `Expo CLI Quickstart` one.

Is that currently possible?

![image](https://user-images.githubusercontent.com/619186/58601717-7735da00-8260-11e9-8e5d-dfe0d63de01d.png)

raajnadar:
No, You can't at-least with the current page structure.

rachelnabors:
Technically it could be done. You'd have to do some work in **tabs.js** to do something like:

- When user activates a tab
- Add metadata to URL

And 
- When landing on a page
- Check URL for metadata
- If metadata found
- Fire appropriate show/hide functionality

> Note: This should only be done for CLI vs Expo info. The `devNotes` data should do this same thing to apply a single user's preferences in localstorage, not the URL

PR #1254, opened 2019-09-09
imp: Add anchors to the getting started tabs

pureinx:
Referenced to issue #968. 
Added links with anchors to the quickstart tabs at the Getting Started page to make possible send a link for specific quickstart method.


jamonholmgren:
This would be quite helpful!

rachelnabors:
I see what you're trying to do here, and I love your intentions! However, this change only works if people know to copy the link to the h-tag. Most won't realize that. 

What would be EVEN MORE AWESOME would be to update the URL with a hash when the user selects from these options so they are preselected when the user copy-pastes from their URL. Would love you to take a crack at this in **tabs.js**

PR #1261, opened 2019-09-11
Update event emitter code examples for Android

alnorris:


charpeni:
Thanks for looking into this. 

Would you mind adding some description on why those things changed?

alnorris:
Hey @charpeni , The code is out of date, (mixins aren't used anymore in React.), and DeviceEventEmitter is now deprecated. See (https://github.com/facebook/react-native/blob/1490ab12ef156bf3201882eeabfcac18a1210352/Libraries/EventEmitter/RCTDeviceEventEmitter.js#L32).  Using NativeEventEmitter seems to now work in Android, and it works exactly like the iOS version.

alnorris:
Additionally, the examples have references to reactContext, so I added it to the constructor so it's easy for someone to reference it in their own class, and then easily copy and paste from the examples. This is in addition to WritableMap and Arguments

PR #1264, opened 2019-09-13
Flex dimensions example doesn't work

JuanELo-O:
The emulator crashes because the markdown has a little mistake.


charpeni:

JuanELo-O:
Your welcome, great work by the way.

PR #1271, opened 2019-09-15
.htaccess file added

realabbas:
# .htaccess file 
For better browser leverage caching 👋


rachelnabors:
Unfortunately, Github doesn't allow `.htaccess` files. Thank you for your kind and clever thought! Sorry we could not use it.

PR #1282, opened 2019-09-22
Add guide for Fast Refresh

rickhanlonii:
## Overview

This PR adds a new guide for Fast Refresh starting in 0.62.

### DIFF
--- a/docs/fast-refresh.md
+++ b/docs/fast-refresh.md
@@ -1,0 +1,38 @@
+---
+id: fast-refresh
+title: Fast Refresh
+---
+
+Fast Refresh is a React Native feature that allows you to get near-instant feedback for changes in your React components. Fast Refresh is enabled by default, and you can toggle "Enable Fast Refresh" in the [React Native developer menu](/intern/wiki/React_Native_Intro/common_workflows/#trigger-react-native-dev). With Fast Refresh enabled, most edits should fall within a second or two.
+
+## How It Works
+
+- If you edit a module that **only exports React component(s)**, Fast Refresh will update the code just for that module, and re-render your component. You can edit anything in that file, including styles, rendering logic, event handlers, or effects.
+- If you edit a module with exports that _aren't_ React components, Fast Refresh will re-run both that module, and the other modules importing it. So if both `Button.js` and `Modal.js` import `Theme.js`, editing `Theme.js` will update both components.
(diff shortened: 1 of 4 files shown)


rickhanlonii:
Preview here: https://deploy-preview-1282--react-native.netlify.com/docs/next/fast-refresh

docs/fast-refresh.md
```diff
@@ -0,0 +1,38 @@
+---
+id: fast-refresh
+title: Fast Refresh
+---
+
+Fast Refresh is a React Native feature that allows you to get near-instant feedback for changes in your React components. Fast Refresh is enabled by default, and you can toggle "Enable Fast Refresh" in the [React Native developer menu](/intern/wiki/React_Native_Intro/common_workflows/#trigger-react-native-dev). With Fast Refresh enabled, most edits should fall within a second or two.
```
cpojer: "most edits should fall within a second or two." – I think "fall" is hard to understand for ESL folks. Maybe "Should take a second or two"?

docs/fast-refresh.md
```diff
@@ -0,0 +1,38 @@
...
+
+- Local state is not preserved for class components (only function components and Hooks preserve state).
+- The module you're editing might have _other_ exports in addition to a React component.
+- Sometimes, a module would export the result of calling higher-order component like `createNavigationContainer(MyScreen)`. If the returned component is a class, state will be reset.
+
+In longer term, as more of your codebase moves to function components and Hooks, you can expect state to be preserved in more cases.
+
+## Tips
+
+- Fast Refresh preserves React local state in function components (and Hooks!) by default.
```
cpojer: remove the !

rickhanlonii:
There were discussions about it being it's own guide (I don't remember if it was with @gaearon or @rachelnabors or both) so that's what I went with here but I don't have a strong opinion where this lives 👍

docs/fast-refresh.md
rickhanlonii: Changed to "most edits should be visible within a second or two"

docs/fast-refresh.md
rickhanlonii: Done

rachelnabors:
Weighing in here!

So the "guides" section itself is a bit bloated at the moment. Needs some organizing. I think it totally makes sense to have this as its own page. But in a separate effort, I'd love for me/us to go through the guides, weed out the least useful, and then sort the remainders into sidebar categories like "development flow" and "design." Let's take this page and work on sorting out that sidebar now!

rachelnabors:
Looks great! Thanks, @rickhanlonii!

PR #1294, opened 2019-09-26
Getting Started guide: Suggest npx instead of globally installing the cli

vp2177:
Also, yarn is an option, but isn't even used in this guide, so don't suggest installing it.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -148,7 +148,6 @@ While you can use any editor of your choice to develop your app, you will need t
 We recommend installing Node, Watchman, and JDK using [Homebrew](http://brew.sh/). Run the following commands in a Terminal after installing Homebrew:
 
 ```
-brew install yarn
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -148,7 +148,6 @@ While you can use any editor of your choice to develop your app, you will need t
 We recommend installing Node, Watchman, and JDK using [Homebrew](http://brew.sh/). Run the following commands in a Terminal after installing Homebrew:
 
 ```
-brew install yarn
```
charpeni: Yarn is used to start the packager and install dependencies.

What do you think about letting this brew install, but keeping your other changes?

vp2177: Ok that's fine. I've put it back.

I thought `npm` is used if `yarn` is not found when installing deps, but still it's needed to start the packager.

PR #1296, opened 2019-09-27
defaultVersion 0.61

acagastya:
Since 0.61 is now the latest stable versoin


ecreeth:
This was fixed in #1309 Thanks!

PR #1307, opened 2019-10-01
Update run instructions on IOS

RafaelDoering:


charpeni:
@brentvatne I thought the QR code has been removed from iOS?

brentvatne:
@charpeni - the Expo client app doesn't have a QR code scanner built in but the iOS Camera app will scan the QR code and open the deep link into the Expo client

PR #1322, opened 2019-10-03
fix pod name yoga -> Yoga

bolasblack:
I encountered the error from cocoapods:

图片

Then I found the pod name should be `Yoga`, not `yoga`


bolasblack:
I found that most of the example code for Podfiles is no longer applicable. Should I change them all?

* Current Podfile in documentation: https://facebook.github.io/react-native/docs/integration-with-existing-apps#configuring-cocoapods-dependencies
* Current Podfile in example: https://github.com/facebook/react-native/blob/v0.61.2/template/ios/Podfile

TheSavior:
Yeah, if more are wrong we should probably fix the rest. Thanks!

PR #1324, opened 2019-10-04
Update getting-started.md

madebyew:
Added details on installing a simulator in Xcode as a simulator is not installed automatically with Xcode.


rachelnabors:
Great catch! Thanks for this write up!

PR #1334, opened 2019-10-11
signed-apk-android.md

pradeep-gupta1:
The generated APK can be found under `android\app\build\outputs\apk\release\app-release.apk`, and is ready to be uploaded to Google Play.


cpojer:
Thanks for sending a PR. I believe you got something mixed up there and I think the documentation is actually correct :)

Issue #1338, opened 2019-10-14
Lint docs for insensitive and inconsiderate language


rickhanlonii:
## Overview

**Update: All of the docs have been claimed 🚧🎉**

Now that we have landed the lint process and fixed the lint issues in the `docs/` directory, we're looking to also lint all of the instances of insensitive and inconsiderate language in the `versioned_docs` directory (i.e. the docs for old versions of React Native.

## How
If you want to lint one of these directories, follow these steps!

### Install
1. Fork the repo
2. Clone your fork
3. Go to the `website` directory and then `yarn install`

### Lint
1. Pick a version, comment here
2. Remove your version from the list in [`.alexignore`](https://github.com/facebook/react-native-website/blob/master/.alexignore#L7-L12)
3. Run this command:

```
yarn lintv website/versioned_docs/version-{version number}
```

This will output all of your errors as in:
<img width="1438" alt="Screen Shot 2019-10-14 at 8 26 40 PM" src="https://user-images.githubusercontent.com/2440089/66777448-f3a02100-eec0-11e9-9c3d-4df867b77747.png">

Once you fix all of the errors, commit all of the changes including the .alexignore file.

To be considerate to the reviewer, please submit each version in different PRs.
## How to fix

In most cases the work can be removed or replace (e.g. "just" could be changed to "only").

In other cases, where a word is quoted or a prop and cannot be changed, we can add  this to the line above:

```
<!-- alex ignore {rule name} -->
```

Where `{rule name}` is the name of the rule you want to ignore. So if you want to ignore the 'host-hostess' rule on a line, you would add this to the line above it:

```
<!-- alex ignore host-hostess -->
```

If you're stuck on what to change, check out the way it's used in the current `docs/` directory. If that doc is no longer there, feel free to ask for advice!

## Submitting PR
To make it easier to merge your PRs, please enable this setting so we can fix any minor issues (like merge conflicts) and merge your PR right away:

![image](https://user-images.githubusercontent.com/2440089/66912004-9d8ac500-f009-11e9-8a6c-806aa5d52cc0.png)

## Versions

| status | owner | version |
| ----- | ----- | ----- |
| ✅ Done | @thackley | `website/versioned_docs/version-0.5` |
| ✅ Done | @carolstran | `website/versioned_docs/version-0.6` |
| ✅ Done | @carolstran | `website/versioned_docs/version-0.7` |
| ✅ Done | @thackley | `website/versioned_docs/version-0.8` |
[...]

PR #1355, opened 2019-10-15
Fix lint rules v0.5

thackley:
This PR handles the linting and fixes instances of insensitive language found in versioned_docs for v0.5

#1338

### DIFF
--- a/.alexignore
+++ b/.alexignore
@@ -10,7 +10,6 @@ website/blog/
 # Once you fix all of the errors, commit all of the changes including this file.
 #
 # To be considerate to the reviewer, please submit each version in different PRs.
-website/versioned_docs/version-0.6/
 website/versioned_docs/version-0.7/
 website/versioned_docs/version-0.8/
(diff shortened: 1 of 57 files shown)


.alexignore
```diff
@@ -10,8 +10,6 @@ website/blog/
 # Once you fix all of the errors, commit all of the changes including this file.
 #
 # To be considerate to the reviewer, please submit each version in different PRs.
-website/versioned_docs/version-0.5/
-website/versioned_docs/version-0.6/
```
rickhanlonii: I think removing this may have been a mistake, and doing so is causing the tests to fail

Don't worry about merge conflicts with master, I'll fix them 👍

rickhanlonii: Sorry, removing 0.6 may have been an mistake - 0.5 should be removed

thackley: Oh yes, removing 0.6 was definitely a mistake. I'll fix!

rickhanlonii:
Thanks so much!!

PR #1380, opened 2019-10-16
Fix/lint rules v0.17

salvariable:


salvariable:
> This one also looks branched off the previous one, could you put each in their own PR branched off master?

Will work on doing that, yes :)

PR #1406, opened 2019-10-18
Fix lint issues in version 0.25

mesaugat:
Fixes lint issues found in versioned_docs directory for version 0.25

Also, looks like alex does not catch the word `simplest` for some reason but I've changed that too.

#1338


mesaugat:
@rickhanlonii Thank you. Alex does not seem to catch the word `simplest`. Should I change those in  all the docs?

rickhanlonii:
Sorry for the delay - simplest seems fine because it's a comparison but if you want to make a case to re-text feel free to file an issue there! https://github.com/retextjs/retext-equality

PR #1409, opened 2019-10-18
More examples for Custom Path Aliase

JeffGuKang:
It made me confused when I tried config module-resolver, especially, in `tsconfig` and `module-resolver` alias part.
So I added another example code.


rachelnabors:
Thank you!

PR #1390, opened 2019-10-17
fix bug in documentation

ffMathy:

PR #1419, opened 2019-10-21
Fix getting started typescript command

ragecryx:
The previous command failed to complete successfully and in the error log seemed to be trying to check-out `react-native-template-react-native-template-typescript`. So I think some kind of refactor in the template param of the cli caused that and tried `npx react-native init MyApp --template typescript` which worked properly.


thymikee:
If this command fails, it means that you have `react-native-cli` installed globally. Please remove it and it will work.

cc @charpeni this can be closed

orta:
Dupe of https://github.com/facebook/react-native-website/pull/1390

PR #1431, opened 2019-10-24
Update textinput.md

retyui:
Add note about specific platform behavior

```js
<TextInput
  returnKeyType="next"
  // Never called
  onSubmitEditing={() => {}}
  keyboardType="phone-pad"
/>

```

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -513,6 +513,8 @@ Callback that is called when the text input's submit button is pressed with the 
 | -------- | -------- |
 | function | No       |
 
+Note that on iOS this method was never called when use `keyboardType="phone-pad"`.
+
 ---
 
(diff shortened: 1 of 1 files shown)


docs/textinput.md
```diff
@@ -513,6 +513,8 @@ Callback that is called when the text input's submit button is pressed with the
 | -------- | -------- |
 | function | No       |
 
+Note that on iOS this method was never called when use `keyboardType="phone-pad"`.
```
cpojer: Could you change this to:

```
Note that on iOS this method is not called when using `keyboardType="phone-pad"`.
```

retyui: Sure
https://github.com/facebook/react-native-website/pull/1527

PR #1437, opened 2019-10-27
Fix broken link on STYLEGUIDE

haruelrovix:
Fix following:

image


rachelnabors:
Nice, thanks for catching this!

PR #1452, opened 2019-10-31
Fix typescript template

iagopiimenta:
Fix error when init a new react native project using typescript template:

image


rachelnabors:
Thanks for sharing! Much obliged!

PR #1455, opened 2019-11-04
Add warning about AAB and Hermes

StephenDocP:
https://github.com/facebook/react-native/issues/26252

Currently Hermes does not support AAB, only APK for Android.

This update will add a warning to the documentation to help warn other developers when trying to build with Hermes and AAB.

### DIFF
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -3,9 +3,6 @@ id: hermes
 title: Using Hermes
 ---
 
-> ## Note about Android App Bundles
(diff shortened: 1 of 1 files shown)


SuhairZain:
But according to this https://github.com/facebook/react-native/issues/26400, hermes enabled builds with AAB doesn't cause a crash, but only that it shows a white screen.

radko93:
This PR is not telling the whole truth, as it's possible to use Hermes and Bundles in RN 0.61.5.

PR #1456, opened 2019-11-06
Updated getting-started.md

rmmariano:
I've updated this file to let explicit that I need to run 'npx react-native start' before running 'npx react-native run-android', because I'm beginner on react-native and I got an error when I run for the first time. (e.g. https://stackoverflow.com/q/55441230/8447990)

Thank you in advance.


rachelnabors:
Excellent catch! Thanks so much for paving the way for future learners!

PR #1458, opened 2019-11-08
Update doc-rating.js

AdamJNavarro:
Added the eventValue property so that the feedback value is sent to GA. Currently the event has no value passed.


rachelnabors:
Fixed in other PR. Thanks for submitting!

PR #1467, opened 2019-11-15
Added yarn command

ayyshim:
It's not huge but since react-native uses yarn package manager to manage libraries, I thought it would be nice if we get to see yarn command to install react-native-navigation.


rachelnabors:
Hey there, someone already patched this up. Thanks for your PR all the same!

PR #1475, opened 2019-11-23
Command doesn't work, need to add extra parameters

Bn8:
Added the parameters needed (project name and template 'blank').
Tested on windows 10.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -31,7 +31,7 @@ npm install -g expo-cli
 Then run the following commands to create a new React Native project called "AwesomeProject":
 
 ```sh
-expo init AwesomeProject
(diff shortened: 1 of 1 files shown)


rachelnabors:
Hey there! Can you pull down master to fix this language linting error?

Bn8:
> Hey there! Can you pull down master to fix this language linting error?

Hey, I fail to see the error in the language_lint test, not only I only changed 1 line but also can't see anywhere the use of the "he" word especially since the only line I changed is a command-line?

cpojer:
Based on Expo's docs, this should not be necessary. See https://expo.io/learn

Maybe something else went wrong when you tried it out?

PR #1476, opened 2019-11-24
Updating the troubleshoot document

indranil32:
For https://github.com/facebook/react-native/issues/26598 on windows.


cpojer:
Thanks! We are currently revamping this section and will consider including this type of troubleshooting information but for now I'll close this PR. Hope that's ok.

PR #1484, opened 2019-11-27
Android before iOS

hramos:
Place Android before iOS when both are referred to in the same sentence.


docs/text.md
```diff
@@ -556,7 +556,7 @@ The highlight color of the text.
 
 - **`textAlign`**: enum('auto', 'left', 'right', 'center', 'justify')
 
-  Specifies text alignment. The value 'justify' is only supported on iOS and Android Oreo (8.0) or above (API level >= 26). For lower android version it will fallback to `left`.
+  Specifies text alignment. The value 'justify' is only supported on Android and iOS Oreo (8.0) or above (API level >= 26). For lower android version it will fallback to `left`.
```
Simek: `iOS Oreo (8.0) or above` - more that order flip is needed here.

website/versioned_docs/version-0.60/text.md
```diff
@@ -562,7 +562,7 @@ The highlight color of the text.
 
 - **`textAlign`**: enum('auto', 'left', 'right', 'center', 'justify')
 
-  Specifies text alignment. The value 'justify' is only supported on iOS and Android Oreo (8.0) or above (API level >= 26). For lower android version it will fallback to `left`.
+  Specifies text alignment. The value 'justify' is only supported on Android and iOS Oreo (8.0) or above (API level >= 26). For lower android version it will fallback to `left`.
```
Simek: `iOS Oreo (8.0) or above` - more that order flip is needed here.

Simek:
@hramos @rachelnabors

Simek:
This PR also shadow fixed CI issue after 2526618f64169da8c652e2f4254dc5b4dcea9c5f commit. 

For the clarity can we expect such changes in future put in separate commit? It was hard to catch that fix in 240 lines of changes.

hramos:
What sort of changes would you like to be added separately?

Simek:
I have in mind the exact changes which fixed CI builds. 

In this case they were even put in as separate commit: https://github.com/facebook/react-native-website/pull/1484/commits/bdb5ccf21c1da747fc53faca4c652bd6d63ee6f5

But to be clear, that was only a question considering future. No need to take an action with this PR.

PR #1486, opened 2019-11-28
Add supportedPlatforms param to Snack embeds

brentvatne:
At the request of @rachelnabors, I added support to Snack for the `supportedPlatforms` URL query param that lets you specify what platforms a Snack will run on. This PR integrates the param into the Snack Remarkable plugin used here and gives an example of using it on the RefreshControl API reference page.

Here is what it looks like with the web option disabled on the RefreshControl page:

Screen Shot 2019-11-28 at 15 14 23


hramos:
Rebased and merged.

PR #1494, opened 2019-12-04
move props related pages to separate section in sidebar

Simek:
This PR adds new "Props" section to the "API" page sidebar which contains all props related pages earlier listed in section below. 

The order of sections can be changed, it's only a proposal aiming to tidy up the sidebar.

Screenshot 2019-12-04 at 14 08 16


Simek:
@rachelnabors What do you think about those changes? If you don't like that idea I can close this PR.

Issue #1272, opened 2019-09-16
React Native Docs Refresh plan through 2019


rachelnabors:
2019 React Native Docs Refresh

The docs site has grown organically since its conception, and because of this, has grown uneven in quality and quantity of content across different topics. Realigning the docs with what today’s learners need will require a bit of analysis to make sure our best efforts are put into the most important materials that learners need the most!

Over the past two months, I dove into analyzing our users, our offerings, and the landscape with:

* Content Analysis 
* Community Interviews 
* Community Surveys

## Goals for Rest of 2019

_The following goals informed by the above work and are subject to change as we proceed._

**Polish Component docs**—React Native developers refer to our components and API docs regularly each week. The following will make these docs even more useful to them.

- [ ] Round out code examples
- [ ] Refactor them to include hooks
- [ ] Add visual examples of each component
- [ ] Expand props definitions
- [ ] Bring props definitions up higher on the page via a sidebar
- [ ] Cross link to related or similar components
- [ ] Performance tips for each component (not lost in a guide in another location no one visits)

**Site retooling** including the following:

- [ ] **Version dropdown** because folks have been struggling to find version-specific information.
- [x] **Better analytics** to follow learners on their journey.
- [x] **“Was this helpful?” thumbs up/down** to help us measure how the new docs are doing
- [ ] **Reducing content size**—the site content has grown over the years. Conciseness and clarity, in addition to being a good measure of solid documentation, helps more people people learn faster.
- [x] **Style guide** for future contributions to be edited against (“we” vs “you”, “they” vs “he”, tone, etc).
- [x] **Automation** for checking contributions against the style guide, possibly using something like Alex.js (https://alexjs.com/); @rickhanlonii implementedintegrated.
- [ ] **Speak to everyone**—people with non-web backgrounds can have a longer ramp up with our getting starting docs. We can improve incorporating vocabulary and examples from Android and iOS communities.

**Reform the guides**—community interviews show that our learners need a bridge between our Getting Started tutorial and our Guides. At this intersection, they frequently go offroading on Google or Youtube or send their new hires to crash courses like Wix’s (https://github.com/wix/react-native-crash-course) —but we and they would prefer to come to the official docs as a Source of Truth for all things app dev related. If the guides were updated, culled, and organized, we could create a proper learning path for them to walk featuring:

- [x] A more skimmable sidebar in Docs
- [ ] Updated docs for readability
- [ ] Remove unused docs
- [x] Consolidate similar docs
[...]

PR #1496, opened 2019-12-05
Restructuring Guides IA and adding new sidebar to Docs

rachelnabors:
As per issue #1272, this restructuctures the guides into a sidebar. 

- Retains platform-specific sections
- Chunks guides out by topics of interest

? Maybe the sidebar should start expanded? But it's so much to show all at once!


website/sidebars.json
```diff
@@ -115,7 +114,6 @@
       "image-style-props",
       "interactionmanager",
       "keyboard",
-      "layout-props",
```
Simek: @rachelnabors Was `layout-props` page been removed intentionally?

rachelnabors: That was unintentional! Good catch :)

thymikee:
Hey, I admire your attempt at categorizing things, it's really hard to get right :D Looking at the new sidebar, I have a feeling that there's too many categories out there. I needed to open them one-by-one to see what's there, as some (like "Interoperability" or "JavaScript Internals") were not quite clear straight away. Thankfully we have an Algolia search for quick access, but still. I think we want to strive for as little categories as possible – that's my personal preference.

Some thoughts for possible improvements:
1. How about moving "Upgrading to new React Native versions" to "Workflow". The "Maintaining" category feels odd, and upgrading is a part of workflow, isn't it? 
2. How about renaming "JavaScript Internals" -> "JavaScript Runtime", since we're talking about JS engines and their runtime execution environments here?
3. Should we mix categories? E.g. "Using Hermes" fits into "Performance" as well, but I'm fine with current state.
4. "Platform Specific Code" fits into "Workflow" or "The Basics". Maybe there's no need for "Interoperability" category?

rachelnabors:
@thymikee Thanks for your feedback! These are all good suggestions. Let me breakdown two important things you brought up:

> I have a feeling that there's too many categories out there. I needed to open them one-by-one to see what's there, as some (like "Interoperability" or "JavaScript Internals") were not quite clear straight away. Thankfully we have an Algolia search for quick access, but still. 

The reason for making categories is that metrics and interviews indicated that readers were not opening "Guides" to find things anyway: they were already using the search but more often heading to Google. We should aim, with these categories, to use words that help people with questions to find the general area of the content providing answers. In that regard, I'm eager for pushback on the category terminologies and happy to work with contributors on them even more!

> Should we mix categories?

Technically, there's no reason we can't put Hermes under both _JS Runtime_ and _Performance_! But is that the right thing to do? It could confuse users to have a sense of deja vu. Perhaps what we really need is a section on Hermes, or a page on "Hermes for performance" in the performance section.

I'll make your tweaks and push them up shortly. Thanks!

thymikee:
I'm also not sold on mixing categories, was curious about your take on it. I think we're on the same page then.

PR #1500, opened 2019-12-08
Update fast-refresh.md

christianalfoni:
Hi there!

As a library author I found it puzzling how fast-refresh affects the hooks. It is nice to document how **useState** and **useRef** differs from the others. Also giving an example of how **useEffect** runs related to a fast-refresh. Because it runs as if the component was removed and added again, which is just nice to know 😄 

Awesome work on this!


gaearon:
I rewrote this a bit, as I don't think the original wording was clear enough. Hope it's ok.

christianalfoni:
For sure, iterations are always good, thanks! 👍 😄

PR #1506, opened 2019-12-11
Fix the spacing on the footer of prev/next buttons

lucasbento:
This PR fixes an issue with the bottom of the prev/next buttons passing the sidebar.

Before:

image

After:

image


orta:
Looks good to me, better than before. On tiny widths it could probably do with pulling in a bit more, but better than it is already 👍

PR #1514, opened 2019-12-16
add react-native-debugger to resources

bell-steven:


cpojer:
Thank you for your PR! We are currently rethinking this section and will consider adding this one, but we are also working on Chrome debugging and integrating with Flipper so we need to be careful about not adding too many suggestions here.

PR #1516, opened 2019-12-17
fix the missing line of run the metro server

AmilaViduranga:
Mention that to run metro server before execute " npx react-native run-android" command. If you guys not mention that line it will make conflicts for beginners.


cpojer:
Thank you for your PR! It says later in the docs that you should execute `run-ios` which will start Metro so we don't need to add this here.

PR #1520, opened 2019-12-22
BlinkApp → Blink

sidnt:
When setState is called, BlinkApp will re-render its Component. → After setState returns, Blink Component is re-rendered.


cpojer:
Thank you for your PR. I think the current text is good enough and changing it to the new version doesn't really help with readability, so let's keep it the way it is.

sidnt:
§1: `When setState is called, BlinkApp will re-render its Component.`
- `setState` is called on `Blink` component. §1 doesn't clarify this.
- Since `setState` is called on `Blink` component
  - `react` re-renders the `Blink` component
  - It is not the `BlinkApp` which re-renders its Component. `BlinkApp` is a component itself.
  - Thus §1 is not precise enough.

PR #1530, opened 2019-12-28
Adds snack preview for KeyboardAvoidingView

arrygoo:
* Adds snack preview for KeyboardAvoidingView
* Removes the recommendation to not set the behavior prop. As demonstrated in the snack, and manually tested by me on multiple Android and iOS, it appears that not setting the `behavior` prop causes the component to stop working


rachelnabors:
Ah, but someone submitted another one in the documentation drive. Thank you, but a duplicate! That said, if you'd like to make some other examples, we'd love your help over [here](https://github.com/facebook/react-native-website/issues/1579)!

rachelnabors:
(I'm taking your other edits, though ;)

PR #1535, opened 2019-12-31
New intro to React Native and React

rachelnabors:
Research indicated that our starter materials could use some updating and expansion to go along with the new information architecture.

Featuring:
* adds new intro to React Native
* combines the state and prop pages into a React refresher course as per issue #929 
* removes and redirects state and prop pages to react-intro
* removes and redirects tutorials to React Native intro
* adds a "how to use this guide" intro
* updates "More Resources" page


docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
+---
+id: intro-react-native-components
+title: Core Components and Native Components
+---
+
+React Native is an opensource framework for building Android and iOS applications using [React](https://reactjs.org/) and the app platform’s native capabilities. With React Native, you use JavaScript to access your platform’s APIs as well as to describe the appearance and behavior of your UI using React components, bundles of reusable, nestable code. You can learn more about React in the next section. But first, let's cover how components work in React Native.
+
+## Views and mobile development
+
+In Android and iOS development, a **view** is the basic building block of UI: a small rectangular element on the screen which can be used to display text, images, respond to user input. Even the smallest visual elements of an app, like a line of text or a button, are kinds of views. Some kinds of views can nest other views. It’s views all the way down!
```
TheSavior: > text, images, respond to user input.

text, images, *or* respond to user input.

docs/introduction.md
```diff
@@ -0,0 +1,119 @@
...
+export default function YourApp() {
+  return (
+      <View style={{ flex: 1, justifyContent: "center", alignItems: "center" }}>
+        <Text>Hello, world!</Text>
+      </View>
+  );
+}
+```
+
+> The sample code in this documentation uses [JavaScript version ES2015](https://babeljs.io/learn-es2015/) (also known as ES6). Since ES2015 is not yet supported by all browsers, React Native ships with ES2015 support out of the box thanks to [Babel](https://babeljs.io/) (which also handles your JSX!). This means you can use ES2015 without worrying about backwards compatibility.
```
TheSavior: I'm not sure that browsers supporting ES2015 impacts React Native. 

Should we just get rid of that part and say "The sample code in this documentation uses JavaScript version ES2015 (also known as ES6). React Native ships with ES2015 support out of the box thanks to Babel...

docs/more-resources.md
```diff
@@ -3,41 +3,41 @@ id: more-resources
...
+## Find, make, and share your own Native Components and Modules
 
-[App Center](https://appcenter.ms/signup?utm_source=Facebook%20React%20Native&utm_medium=Azure&utm_campaign=GitHub) is a service from Microsoft that allows you to deploy live updates to your React Native app. If you don't like going through the app store process to deploy little tweaks, and you also don't like setting up your own backend, give [App Center](https://appcenter.ms/signup?utm_source=Facebook%20React%20Native&utm_medium=Azure&utm_campaign=GitHub) a try.
+React Native has a community of thousands of developers like you making content, tools, tutorials—and Native Components!
 
-[Expo](https://docs.expo.io) is a development environment plus application that focuses on letting you build React Native apps in the Expo development environment, without ever touching Xcode or Android Studio. If you wish React Native was even more JavaScripty and webby, check out Expo.
+Can’t find what you’re looking for in the Core Components? Check [JS.coach](https://js.coach/react-native) or [Native Directory](https://www.native.directory/) to find what the community has been creating.
```
TheSavior: Same ordering question here

docs/more-resources.md
```diff
@@ -3,41 +3,41 @@ id: more-resources
...
-
-[Bugsnag](https://www.bugsnag.com/), [Microsoft App Center](https://appcenter.ms/signup?utm_source=Facebook%20React%20Native&utm_medium=Azure&utm_campaign=GitHub), and [Sentry](https://sentry.io/welcome/) all provide excellent crash and error monitoring services for React and React Native apps. These services allow you to proactively monitor crashes and issues occurring on your apps in real time so you can fix them quickly and improve user experience.
-
-The [React Developer Tools](debugging.md#react-developer-tools) are great for debugging React and React Native apps.
+Interested in making your own Native Component or Module? Making modules for your own use case and sharing them with others on NPM and github helps grow the React Native ecosystem and community! Read the guides to making your own Native Modules ([Android](native-modules-android.md), [iOS](native-modules-ios.md)) and Native Components ([Android](native-components-android.md), [iOS](native-components-ios.md)).
```
TheSavior: > github

GitHub

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+```
+
+<block class="classical functional" />
+
+## Make your own Native Components
+
+As you continue developing with React Native, you may want to use a kind of view on Android or iOS that isn’t in the Core Components. For this reason, React Native lets you to build your own Native Components for [Android](native-components-android.md) and [iOS](native-components-ios.md) to suit your app's unique needs. We also have a thriving ecosystem of these community-contributed components and modules. Check [JS.coach](https://js.coach/react-native) or [Native Directory](https://www.native.directory/) to find what the community has been creating.
```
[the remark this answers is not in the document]
rachelnabors: That was already in the More Resources section, so I brought it over here, too. It would be soooo great to have something like native.directory here!

docs/more-resources.md
rachelnabors: It's in that order already on the site. Would you recommend reordering it or removing JS.coach?

docs/intro-react.md
```diff
@@ -0,0 +1,540 @@
...
+- JSX
+- custom components
+- props
+- state
+
+If you want to dig deeper, we encourage you to check out [React's official documentation](https://reactjs.org/docs/getting-started.html).
+
+## Your first component
+
+Your first task is to make a pet component. To define your `MyPet` component, first use JavaScript’s [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) to React Native’s [`Text`](/react-native/docs/next/text) components:
```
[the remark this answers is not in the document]
rachelnabors: Ok! I'll provide some inline context then!

docs/introduction.md
rachelnabors: Switched over to "engines" to be technically correct ;)

docs/intro-react.md
```diff
@@ -0,0 +1,540 @@
...
+
+Props are set by the parent component and can only be changed by the parent component passing down new values. But user interfaces need to accept new data (input) directly and change according to human interaction. Fortunately, we have `state`, which lets us get and set data associated with a component. You can manipulate component state by calling [React’s `useState` Hook](https://reactjs.org/docs/hooks-state.html).
+
+> A Hook is a kind of function that lets you “hook into” React features. For example, useState is a Hook that lets you add React state to function components. You can learn more about [other kinds of Hooks in the React documentation.](https://reactjs.org/docs/hooks-intro.html)
+
+You can manipulate component state by calling `useState` inside its function:
+
+`const [currentValue, setNewValue] = useState("Initial value.");`
+
+Or you could set it like so:
```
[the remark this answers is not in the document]
rachelnabors: How about "You can use `useState` to track any kind of data, including numbers:"

docs/intro-react.md
```diff
@@ -0,0 +1,540 @@
...
+                <MyPet />
+            </View>
+        );
+    }
+}
+```
+
+<block class="functional classical webNote androidNote" />
+
+In the above example `MyPet` is a React component. In `MyPets` each `<MyPet>` renders a completely different pet. These pets can all have different data, which we pass to them via props.
```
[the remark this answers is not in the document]
rachelnabors: There was a bit of HTML missing that prevented an example from showing, urf!

docs/intro-react.md
```diff
@@ -0,0 +1,613 @@
...
+}
+```
+
+<block class="classical functional webNote androidNote" />
+
+To define your `MyPet` component, first use JavaScript’s [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) to import React and React Native’s [`Text`](/react-native/docs/next/text) component:
+
+```jsx
+import React from 'react';
+import {Text} from 'react-native';
```
brentvatne: nit: other code snippets seem to include spacing around imports, eg: `import { Text }`

docs/intro-react.md
```diff
@@ -0,0 +1,613 @@
...
+        );
+    }
+}
+```
+
+<block class="functional classical webNote androidNote" />
+
+You can build many things with `props` and the Core Components [`Text`](text), [`Image`](image), and [`View`](view)! But to build something interactive, you’ll need `state`.
+
+## State
```
brentvatne: do you think it's worth adding a class-based state explanation as well? might trip folks up if they follow that through the doc and get to the end to find only the hooks approach to state

docs/introduction.md
```diff
@@ -0,0 +1,121 @@
...
+      </View>
+    );
+  }
+}
+
+```
+
+<block class="classical functional webNote iosNote androidNote" />
+
+The above is a Snack Player. It’s [a handy tool created by Expo](https://expo.io/tools#snack) to embed and run React Native projects on and share how they render in platforms like Android and iOS. We will be using them a lot to demonstrate code samples throughout these docs. You can play around with sample code provided in the Snack, or you can [set up an environment on your local machine](http://localhost:3000/react-native/docs/next/getting-started) and paste the code into your `App.js` file there.
```
brentvatne:
```suggestion
The above is a Snack Player. It’s [a handy tool created by Expo](https://expo.io/tools#snack) to embed and run React Native projects on and share how they render in platforms like Android and iOS. We will be using them a lot to demonstrate code samples throughout these docs. You can play around with sample code provided in the Snack, or you can [set up an environment on your local machine](getting-started) and paste the code into your `App.js` file there.
```

docs/intro-react.md
rachelnabors: After some user testing and discussion with Dan and a few others, it seems the Best Thing might be removing the class examples and sharing a legacy example and explanation at the end of the intro.

brentvatne: that seems reasonable!

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
+---
+id: intro-react-native-components
+title: Core Components and Native Components
+---
+
+React Native is an opensource framework for building Android and iOS applications using [React](https://reactjs.org/) and the app platform’s native capabilities. With React Native, you use JavaScript to access your platform’s APIs as well as to describe the appearance and behavior of your UI using React components, bundles of reusable, nestable code. You can learn more about React in the next section. But first, let's cover how components work in React Native.
```
kelset:
```suggestion
React Native is an open source framework for building Android and iOS applications using [React](https://reactjs.org/) and the app platform’s native capabilities. With React Native, you use JavaScript to access your platform’s APIs as well as to describe the appearance and behaviour of your UI using React components, bundles of reusable, nestable code. You can learn more about React in the next section. But first, let's cover how components work in React Native.
```

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
+---
+id: intro-react-native-components
+title: Core Components and Native Components
```
kelset: I think that the name of this section is a bit "redundant" - what about "Components: Native and  Core"?

docs/introduction.md
```diff
@@ -0,0 +1,121 @@
+---
+id: introduction
+title: Introduction
+---
+
+So many different kinds of people use React Native: from master iOS developers to React JS beginners, to people getting started programming for the first time in their career. These docs were written for all learners, no matter their experience level or background.
```
kelset:
```suggestion
Many different kinds of people use React Native: from advanced iOS developers to React JS beginners, to people getting started programming for the first time in their career. These docs were written for all learners, no matter their experience level or background.
```

docs/introduction.md
```diff
@@ -0,0 +1,121 @@
...
+      </View>
+    );
+  }
+}
+
+```
+
+<block class="classical functional webNote iosNote androidNote" />
+
+The above is a Snack Player. It’s [a handy tool created by Expo](https://expo.io/tools#snack) to embed and run React Native projects on and share how they render in platforms like Android and iOS. We will be using them a lot to demonstrate code samples throughout these docs. You can play around with sample code provided in the Snack, or you can [set up an environment on your local machine](getting-started) and paste the code into your `App.js` file there.
```
kelset: I think the "call to action" to the docs about setting up your machine should be further up, before the "How to use these docs" section.

Something just like, "It's not necessary, but in case you want to setup your local machine before proceeding you can read the dedicated document". 

I feel that setting up one's machine is one of the most complicated parts of the "learning curve" and having the reference here makes it a bit less visible IMHO

docs/introduction.md
```diff
@@ -0,0 +1,121 @@
...
+
+<block class="iosNote classical functional" />
+
+> iOS developers may be familiar with this concept.
+
+<block class="webNote androidNote iosNote  classical functional" />
+
+---
+
+Now that you know how this guide works, it's time to get to know the foundation of React Native: [Native Components](intro-react-native-components.md).
```
kelset:
```suggestion
Now that you know how this guide works, it's time to get to know the foundation of React Native: [Native and Core Components](intro-react-native-components.md).
```

website/i18n/en.json
```diff
@@ -114,7 +114,7 @@
         "title": "Gesture Responder System"
       },
       "getting-started": {
-        "title": "Getting Started"
+        "title": "Setting up the development environment"
```
kelset: I think it would be better to fully rename the "old" getting started section to "setting-up-machine" or something in order to avoid confusion moving forward: I find kind of confusing to see the url still saying "getting started" when I click on the setting up the machine.

website/i18n/en.json
```diff
@@ -161,6 +161,15 @@
       "interactionmanager": {
         "title": "InteractionManager"
       },
+      "intro-react-native-components": {
+        "title": "Core Components and Native Components"
+      },
+      "intro-react": {
+        "title": "Key React Concepts"
```
kelset: tbh I feel that even without "Key" in front, this title would work.

website/i18n/en.json
```diff
@@ -161,6 +161,15 @@
       "interactionmanager": {
         "title": "InteractionManager"
       },
+      "intro-react-native-components": {
+        "title": "Core Components and Native Components"
```
kelset: Same comment as above, I'm not super fond of "XX Components and YYY Components" - feels redundant.

website/i18n/en.json
kelset: Also, I think that the button on the main page of the website will still link too this page? But I guess the idea would be to change that to point to the new introduction.

docs/intro-react-native-components.md
rachelnabors: I'm worried about internationalization for that one. Explicit sets of words are easier for translators to handle. Sometimes redundant is clearer for more folks!

docs/introduction.md
rachelnabors: Well, that was a bit intentional. Research was showing learners bogging down on environment setup, which is pretty complicated and can be discouraging to folks who just want to get a feel for React Native. The hope is that by moving it past the tutorial mode to a clearly labeled section and using in-browser examples, folks would feel more confident about investing so much time later. But let me have a try!

docs/intro-react-native-components.md
kelset: Makes sense :) Ok then totally ignore my comment 👍

website/i18n/en.json
kelset: removing base don previous comment

docs/introduction.md
kelset: Oh I see!
I think I see your point - and I think it totally makes sense to "move away" the setup from this area.

website/i18n/en.json
rachelnabors: At first it was "React Basics" to imply that it wasn't covering ALL of React. But then we wondered if there was a more inclusive way of expressing "just enough React to get by but not all the React." What do you think?

kelset: Fundamentals of React?
React foundations?

cpojer:
With this PR, selecting "next" for docs and then clicking the "Docs" in the header will select https://deploy-preview-1535--react-native.netlify.com/docs/next/getting-started instead of https://deploy-preview-1535--react-native.netlify.com/docs/next/introduction. Is that expected?

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+
+You can start here and read through these docs linearly like a book; or you can read the specific sections you need. Already familiar with React JS? You can skip that section—or read it for a light refresher.
+
+## Prerequisites
+
+To work with React Native, you will need to have an understanding of JavaScript fundamentals. If you’re new to JavaScript or need a refresher, you can [dive in](https://developer.mozilla.org/en-US/docs/Web/JavaScript) or [brush up](https://developer.mozilla.org/en-US/docs/Web/JavaScript/A_re-introduction_to_JavaScript) at Mozilla Developer Network.
+
+> While we do our best to assume no prior knowledge of React, iOS, or Android development, these are valuable topics of study for the aspiring React Native developer. Where sensible, we have linked to resources and articles that go more in depth.
+
+### Environment Setup
```
cpojer: I recommend moving this after "Interactive examples". Reading from the top down it wasn't clear to me what the next step was: should I click this link and go somewhere else or continue to scroll down? I think it's best to get people to place with RN asap, and then recommend "oh, if you don't wanna play with it here on the website, set up a local dev env here […] and then continue the guide."

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+
+Optionally, if you want to setup a local development environment, [you can follow our guide to setting up your environment on your local machine](getting-started). Our you can follow along with the interactive examples used throughout this guide.
+
+## Interactive examples
```
cpojer: Maybe this one could have a bit more woo, like "Get started immediately with interactive examples"?

docs/introduction.md
cpojer: Could we remove the env setup guidance from this paragraph and move the paragraph above this one here (See my comment above)? I believe the flow will work better that way:

* How do use the docs
* You need to know JS
* Play with it right here
* Optionally install a local setup and come back

vs:

* How do use the docs
* You need to know JS
* Optionally install a local setup and maybe come back?
* Play with it right here

The first one gets people to jump in right away, making the local setup a secondary flow for people who just want to get started.

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+      </View>
+    );
+  }
+}
+
+```
+
+<block class="classical functional webNote iosNote androidNote" />
+
+The above is a Snack Player. It’s [a handy tool created by Expo](https://expo.io/tools#snack) to embed and run React Native projects on and share how they render in platforms like Android and iOS. We will be using them a lot to demonstrate code samples throughout these docs. You can play around with sample code provided in the Snack, or you can [set up an environment on your local machine](getting-started) and paste the code into your `App.js` file there.
```
cpojer: One more thing: I got really confused on the website that "a handy tool created by Expo" was linked to some other page. I stopped reading there, clicked the link and got lost on the Expo website without knowing what to do next. Expo has its logo and name on the player that link to the website, and we mention it on almost every page. I think we could get away without turning this text into a link and just letting people play with it. If they are curious, they'll check out Expo on their own.

cpojer: Oops, I have one more thought: Could this paragraph encourage people to make a specific change? Maybe the example could have an emoji in it that is commented out and removing the comment will show a 🎉 emoji or something? That would show people the power of RN and Snack right away and get them hooked :)

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Function and Class Syntaxes
+
+You may have noticed the two tabs for “Functional Syntax” and “Class Syntax” in the Snack above. We thought it would be useful for you to have both syntaxes to study, though we will only go into depth with function components.
+
+Classes were the only way to manage state in a React component until [React 16.8 introduced an API called “Hooks” for managing state and more in function components](https://reactjs.org/docs/hooks-intro.html). In March 2019, [React Native 0.58 adopted Hooks as well](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059). While Hooks are the future-facing way to write your React-based components, you might be working in a codebase that still uses classes.
```
cpojer: The version numbers and dates will get outdated very quickly. The React version doesn't really mean anything to most RN devs. Could we instead assume hooks as the default and simply mentioning ("Hooks were introduced in React Native 0.58" with a link to the blog post, without dates added?

I'm looking at this as somebody from 2023 (yes, your docs will still be around by then). Reading "March 2019" would make me feel "wow that's old".

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+export default function YourApp() {
+  return (
+    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
+      <Text>Hello, world!</Text>
+    </View>
+  );
+}
+```
+
+> The sample code in this documentation uses [JavaScript version ES2015](https://babeljs.io/learn-es2015/) (also known as ES6). Since ES2015 is not supported by all JavaScript engines, React Native ships with ES2015 support out of the box thanks to [Babel](https://babeljs.io/) (which also handles your JSX!). This means you can use ES2015 without worrying about backwards compatibility.
```
cpojer: Could you remove this? All engines that RN runs on since 0.59 are using modern JavaScript (ES2015+) and as such Babel is not required to ship modern features, even though the transform is still enabled by default.

(cc @ide – hey, should we do anything about that? We should be able to disable almost all babel transforms in RN by default but we'd need to figure out a migration strategy. Good app size and possibly perf impact to be had.)

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Developer Notes
+
+People with many different developer experiences are learning React Native. You may have experience with a range of technologies, from web to Android to iOS and more. We try to write for developers from all backgrounds. Sometimes we provide explanations specifc to one platform or another like so:
```
cpojer: Maybe "Many people with a different developer experience"?

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
+      <Text>Hello, world!</Text>
+    </View>
+  );
+}
+```
+
+> The sample code in this documentation uses [JavaScript version ES2015](https://babeljs.io/learn-es2015/) (also known as ES6). Since ES2015 is not supported by all JavaScript engines, React Native ships with ES2015 support out of the box thanks to [Babel](https://babeljs.io/) (which also handles your JSX!). This means you can use ES2015 without worrying about backwards compatibility.
+
+## Developer Notes
```
cpojer: I love this. In a follow-up, we should probably make these tabs store their state in a cookie/localStorage so it persists across pages.

docs/introduction.md
```diff
@@ -0,0 +1,125 @@
...
+
+<block class="iosNote classical functional" />
+
+> iOS developers may be familiar with this concept.
+
+<block class="webNote androidNote iosNote  classical functional" />
+
+## Formatting
+
+Menu paths are written in bold and use carats to navigate submenus. Example: **Android Studio > Preferences**
```
cpojer: Nit: `carets`.

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+---
+
+React Native is an open source framework for building Android and iOS applications using [React](https://reactjs.org/) and the app platform’s native capabilities. With React Native, you use JavaScript to access your platform’s APIs as well as to describe the appearance and behaviour of your UI using React components, bundles of reusable, nestable code. You can learn more about React in the next section. But first, let's cover how components work in React Native.
```
cpojer: Nit: Use a colon instead of comma: "UI using React components: bundles of reusable, nestable code"

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Views and mobile development
+
+In Android and iOS development, a **view** is the basic building block of UI: a small rectangular element on the screen which can be used to display text, images, or respond to user input. Even the smallest visual elements of an app, like a line of text or a button, are kinds of views. Some kinds of views can nest other views. It’s views all the way down!
```
cpojer: Should "View" be capitalized here since I assume it is referring to the RN primitive?

cpojer: Maybe: "Some kinds of views can nest other views" -> "Some kinds of views can be composed of other views"?

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Native Components
+
+In Android development, you write views in Java or Kotlin; in iOS development, you use Objective-C or Swift. With React Native, you can invoke these views with JavaScript using React components. At runtime, React Native creates the corresponding Android and iOS views for those components. Because React Native components are backed by the same views as Android and iOS, React Native apps look, feel, and perform like any other apps. We call these platform-backed components Native Components.
```
cpojer: Given the recent popularity of newer languages, I'd reverse the order: "Kotlin or Java; in iOS development you use Swift or Objective-C"

Also may be nice to make "Native Components" at the end bold so that it connects with the headline. You are driving the point home so nicely about what they are but by the end of the paragraph I kinda forgot that was the focus of the paragraph (might just be because your writing is that good I got lost!).

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Core Components
+
+React Native comes with a suite of built-in components ready for you to use. These Core Components include:
+
+| React Native UI Component | Android View   | iOS View         | Web Analog               | Description                                                                                           |
```
cpojer: Our table rendering looks kinda off. We should figure out how to give this table the full width available, or a min-width.

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+| `<Text>`                  | `<TextView>`   | `<UITextView>`   | `<p>`                    | Displays, styles, and nests strings of text and even handles touch events                             |
+| `<Image>`                 | `<ImageView>`  | `<UIImageView>`  | `<img>`                  | Displays different types of images                                                                    |
+| `<ScrollView>`            | `<ScrollView>` | `<UIScrollView>` | `<div>`                  | A generic scrolling container that can contain multiple components and views                          |
+| `<TextInput>`             | `<EditText>`   | `<UITextField>`  | `<input type="text">`    | Allows the user to enter text                                                                         |
+
+React Native has many more Core Components for everything from form controls to activity indicators. You can find them [documented in the APIs section](components-and-apis). For now, we’re going to work with these five Core Components. In the next section you will start combinging them. Have a play with them here now!
```
cpojer: Nit: "API section"

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Make your own Native Components
+
+As you continue developing with React Native, you may want to use a kind of view on Android or iOS that isn’t in the Core Components. For this reason, React Native lets you to build your own Native Components for [Android](native-components-android.md) and [iOS](native-components-ios.md) to suit your app's unique needs. We also have a thriving ecosystem of these community-contributed components and modules. Check [Native Directory](https://www.native.directory/) to find what the community has been creating.
+
+---
+
+Because React Native uses the same API structure as React components, you’ll need to understand React component APIs to get started. The [next section](intro-react) makes for a quick introduction or refresher on the topic. However, if you’re already familiar with React, feel free to [skip ahead](handling-text-input).
+
+<img src="/react-native/docs/assets/react-native-components.png" srcset="/react-native/docs/assets/react-native-components2x.png 2x" width="1000" alt="A diagram showing React Native's Core Components are a subset of React Components that ship with React Native.">
```
cpojer: Is this diagram intentionally on the bottom after the text asking me to navigate away from this page?

docs/intro-react.md
```diff
@@ -0,0 +1,613 @@
...
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+
+export default function MyPet() {
+    return (
+        <View>
+            <Text>Hello, I am...</Text>
+            <TextInput
+                style={{ height: 40, borderColor: 'gray', borderWidth: 1 }}
+                value="Name me!"
```
cpojer: This example has a hardcoded `TextInput` value, making it not editable. I don't think that was the intention?

docs/introduction.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+<block class="classical syntax" />
+
+```SnackPlayer name=Hello%20World%20Class%20Component
+import React, { Component } from 'react';
+import { Text, View } from 'react-native';
+
+export default class HelloWorldApp extends Component {
```
gaearon: Nit: this one is `HelloWorldApp` but the other one is `YourApp`. Use the same name?

docs/introduction.md
```diff
@@ -0,0 +1,138 @@
...
+
+> iOS developers may be familiar with this concept.
+
+<block class="endBlock devNotes" />
+
+## Formatting
+
+Menu paths are written in bold and use carats to navigate submenus. Example: **Android Studio > Preferences**
+
+Inline code
```
gaearon: I'm not sure what this is referring to. Was this meant to be in backticks? Or is this a header?

Screen Shot 2020-01-15 at 8 07 34 PM

docs/intro-react-native-components.md
gaearon: A dash would work too. I thought this was a run-on sentence at first.

docs/intro-react-native-components.md
gaearon: Yeah, I got confused that "... Native Components" was immediately followed by "Core Components" and thought there's a terminology mistake. I didn't noticed that this section had a "Native Components" headline.

One possible fix would be to introduce the name in the beginning of this section.

For example:

>## Native Components

>Some components are provided by the platform. We call these platform-backed components *Native Components*. For example, in Android [...]

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,63 @@
...
+
+In Android and iOS development, a **view** is the basic building block of UI: a small rectangular element on the screen which can be used to display text, images, or respond to user input. Even the smallest visual elements of an app, like a line of text or a button, are kinds of views. Some kinds of views can nest other views. It’s views all the way down!
+
+## Native Components
+
+In Android development, you write views in Java or Kotlin; in iOS development, you use Objective-C or Swift. With React Native, you can invoke these views with JavaScript using React components. At runtime, React Native creates the corresponding Android and iOS views for those components. Because React Native components are backed by the same views as Android and iOS, React Native apps look, feel, and perform like any other apps. We call these platform-backed components Native Components.
+
+## Core Components
+
+React Native comes with a suite of built-in components ready for you to use. These Core Components include:
```
gaearon: After reading this, I don't have a super clear idea of what's the difference between Native and Core components, whether it matters, or whether they're the same thing. Maybe just clarifying that with a few words before jumping into the table could help.

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,63 @@
...
+React Native has many more Core Components for everything from form controls to activity indicators. You can find them [documented in the APIs section](components-and-apis). For now, we’re going to work with these five Core Components. In the next section you will start combining them. Have a play with them here now!
+
+```SnackPlayer name=Hello%20World
+import React from 'react';
+import { View, Text, Image, ScrollView, TextInput } from 'react-native';
+
+export default function App() {
+  return (
+    <ScrollView>
+      <Text>ScrollViews scroll!</Text>
```
gaearon: This one doesn't scroll for me though. At least you can't see it in the snack. Maybe we can remove this line?

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,63 @@
...
+        defaultValue="You can type in me"
+      />
+    </ScrollView>
+  );
+}
+```
+
+## Make your own Native Components
+
+As you continue developing with React Native, you may want to use a kind of view on Android or iOS that isn’t in the Core Components. For this reason, React Native lets you to build your own Native Components for [Android](native-components-android.md) and [iOS](native-components-ios.md) to suit your app's unique needs. We also have a thriving ecosystem of these community-contributed components and modules. Check [Native Directory](https://www.native.directory/) to find what the community has been creating.
```
gaearon: Nit: "a kind of view" is a bit awkward. Maybe just "a view"

docs/intro-react-native-components.md
gaearon: Diagram nit: I find most of these "s" at the end a bit hard to read as a non-native speaker. (Cause I don't draw it like this.) Overall I would prefer a tiny bit more polished diagram.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+function Cat() {
+  return <Text>Hello, I am your cat!</Text>;
+}
+```
+
+You can export your function component with JavaScript’s [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) for use throughout your app like so:
+
+```jsx
+export default function Cat() {
+  // element(s) go here!
```
gaearon: This snippet is focused on introducing `export default`. However, you're stealing attention from that by replacing `return <Text>Hello, I am your cat!</Text>;` from previous example with `// element(s) go here!`. Let's just keep the function contents as is now, since we already introduced the elements. Then `export default` will be more noticeable.

Maybe even:
```jsx
// We added "export default"
export default function Cat() {
  return <Text>Hello, I am your cat!</Text>;
}
```

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+Whatever a function component returns is rendered as a React element. `Cat` will render a `<Text>` element:
+
+```jsx
+function Cat() {
+  return <Text>Hello, I am your cat!</Text>;
+}
+```
+
+You can export your function component with JavaScript’s [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) for use throughout your app like so:
```
gaearon: Maybe say `export default` here since we added both keywords. Someone following along might not realize and add `export` alone.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+```jsx
+export default function Cat() {
+  // element(s) go here!
+}
+```
+
+<block class="classical syntax" />
+
+Class components have a little more complexity compared to function components in some ways.
```
gaearon: Complexity is a bit subjective. Maybe:

Class components tend to be a bit more verbose than function components.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+  }
+}
+```
+
+And as with function components, you can export your class component:
+
+```jsx
+export default function Cat() {
+  // element(s) go here!
+}
```
gaearon: 1. This should be a class
2. Let's keep the exact example you wrote above but add `export default` to it — don't change the render method

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+<block class="endBlock syntax" />
+
+> This is one of many ways to export your component. This kind of export works well with the Snack Player. However, depending on your app’s file structure, you might need to use a different convention. This [handy cheatsheet on JavaScript imports and exports](https://medium.com/dailyjs/javascript-module-cheatsheet-7bd474f1d829) can help.
+
+Now take a closer look at that `return` statement. `<Text>Hello, I am your cat!</Text>` is using a kind of JavaScript syntax that makes writing elements convenient: JSX.
+
+## JSX
+
+React and React Native use JSX, a syntax that lets you write elements inside JavaScript like so: `<Text>Hello, I am your cat!</Text>`. The React docs have [a comprehensive guide to JSX](https://reactjs.org/docs/introducing-jsx.html) you can reference to learn even more. Here’s how you would write the previous example with JSX:
```
gaearon: So here's an interesting thing. The doc you link to is an intro to JSX. But we also *do* have a [more in-depth look at JSX](https://reactjs.org/docs/jsx-in-depth.html). Whether it's useful here or not is up to you. Maybe a link at the end.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
```
gaearon: >Here’s how you would write the previous example with JSX:

The previous example already uses JSX. So maybe this should say something like "Let's take another look at this example, paying attention to the `<Text>` element we return" or something.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+import { Text } from 'react-native';
+
+export default function Cat() {
+  const name = "Maru";
+  return (
+    <Text>Hello, I am {name}!</Text>
+  );
+}
+```
+
```
gaearon: One way I like to conceptualize JSX is `{` and `}` let you "escape" back to the JavaScript land. It might not be obvious that any JavaScript expression is valid here, e.g. `formatName(name)`. Might be worth clarifying so people don't think it *has* to be a variable.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+export default function Cat() {
+  return (
+    <View>
+      <Text>Hello, I am...</Text>
+      <TextInput
+        style={{
+          height: 40,
+          borderColor: 'gray',
+          borderWidth: 1 }}
+        value="Name me!"
```
gaearon: defaultValue so it's editable

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+export default function Cat() {
+  return (
+    <View>
+      <Text>Hello, I am...</Text>
+      <TextInput
+        style={{
+          height: 40,
+          borderColor: 'gray',
+          borderWidth: 1 }}
```
gaearon: Nit: `}}` on next line for style. In general, you can use formatting that https://prettier.io/ does.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+  </span>
+</div>
+
+<block class="webNote devNotes" />
+
+> If you’re familiar with web development, `<View>` and `<Text>` might remind you of HTML! You can think of them as the `<div>` and `<p>` tags of application development.
+
+<block class="androidNote devNotes" />
+
+> On Android, you usually put your views inside LinearLayout, FrameLayout, RelativeLayout, etc. to define how the view’s children will be arranged on the screen. In React Native, `View` uses Flexbox for its children’s layout. You can learn more in [our guide to layout with Flexbox](https://facebook.github.io/react-native/docs/next/flexbox).
```
gaearon: The link should probably start with `/` rather than hardcoding the next branch?

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+    <View>
+      <Cat name="Maru" />
+      <Cat name="Jellylorum" />
+      <Cat name="Spot" />
+    </View>
+  );
+}
+```
+
+Most of React Native’s Core Components can be customized with props, too. For example, when using [`Image`](image), you pass it a `prop` named [`source`](image#source) to define what image it shows:
```
gaearon: Nit: don't backtick "prop" because it's not an API

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+```
+
+`Image` has [many different props](image#props), including [`style`](image#style), which accepts a JS object of design and layout related property-value pairs.
+
+> Notice the double curly braces `{{ }}` surrounding `style`‘s width and height. In JSX, JavaScript values are referenced with `{}`. This is handy if you are passing something other than a string as props, like an array or number (`<Cat food={["fish", "kibble"]} /> age={2}`). However, JS object are **_also_** denoted with curly braces: `{width: 193, height: 110}`. Therefore, to pass a JS object in JSX, you must wrap the object in **another pair** of curly braces: `{{width: 193, height: 110}}`
+
+You can add custom cat pictures to your component by:
+
+- adding an `Image` component to your `Cat` component
+- setting `Image`‘s `source` via `Cat`‘s `props` with `{props.picURI}`
```
gaearon: Nit: don't backtick props

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+- then passing `picURI` values in each `Cat`
+
+```SnackPlayer name=Custom%20Pictures
+import React from 'react';
+import { Text, View, Image } from 'react-native';
+
+function Cat(props) {
+  return (
+    <View>
+      <Image source={props.picURI} style={{width: 200, height: 200}}/>
```
gaearon: Style nit: add a space before />. Or use Prettier!

docs/intro-react.md
gaearon: Fun fact: JSX doesn't work if you don't have `import React from 'react'` at the top. Maybe worth mentioning, it trips people at first.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+    </View>
+  );
+}
+
+export default function Cafe() {
+  return (
+    <View>
+      <Cat name="Maru" picURI="https://placekitten.com/g/200/200" />
+      <Cat name="Jellylorum" picURI="https://placekitten.com/g/200/200" />
+      <Cat name="Spot" picURI="https://placekitten.com/g/200/200" />
```
gaearon: Kind of negates the point of this example to pass the same URIs. Can we come up with something?

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+    </View>
+  );
+}
+```
+
+You can build many things with `props` and the Core Components [`Text`](text), [`Image`](image), and [`View`](view)! But to build something interactive, you’ll need `state`.
```
gaearon: Don't backtick state.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+## State
+
+Props are set by the parent component and can only be changed by the parent component passing down new values. But for individual components to change according to human interactions, they need to be able to accept and store their own data. Fortunately, React components can have their own `state` you can get and set data from.
```
gaearon: Same: don't backtick state.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
```
gaearon: I find that this paragraph still leaves me fuzzy about the distinction between props and state. Maybe some analogy could work? Props are like "component arguments", and state is like "component variables". State is always owned by a particular component, which can pass it down as props to other components.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+<block class="functional syntax" />
+
+You can add state to a component by calling [React’s `useState` Hook](https://reactjs.org/docs/hooks-state.html).
+
+> A Hook is a kind of function that lets you “hook into” React features. For example, `useState` is a Hook that lets you add state to function components. You can learn more about [other kinds of Hooks in the React documentation.](https://reactjs.org/docs/hooks-intro.html)
+
+You can declare a component’s state by calling `useState` inside its function like this:
+
+`const [currentValue, setNewValue] = useState("Initial value.");`
```
gaearon: Make this a full code snippet for highlighting? Also it's easy to miss this happens inside the component. I suggest:

```js
function Something() {
  const [currentValue, setNewValue] = useState("Initial value.");`
  // ...
```

Also we try to follow the convention like `thing / setThing` so that they match 1:1.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+- it creates a “state variable” with an initial value
+- it creates a function to set that state variable’s value
+
+Let’s break `[a, b] = useState(c)` down:
+
+- `a` is the “state variable”—it refers to the current value of the component’s state
+- `b` is the name of the function you use to update the state variable
+- `c` is the value you initialize the state variable with
+
```
gaearon: I think at this point I'd still wonder: why do I need a "state variable"?

Maybe you could say: 

>Normally, when you exit a function, its variables disappear. But we want our components to remember some data —for example, the user input. This is where state variables come in. They let components remember things! Here's how to declare a state variable.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+  return (
+    <>
+      <Cat />
+      <Cat />
+    </>
+  );
+}
+```
+
+> See the `<>` and `</>` above? These bits of JSX are [fragments](https://reactjs.org/docs/fragments.html). Adjacent JSX elements must be wrapped in an enclosing tag. Fragments let you do that without nesting an extra, unnecessary wrapping element like `View`.
```
gaearon: By this point I'm overloaded trying to understand the state explanation, so adding more concepts overflows my brain. Maybe let's get to fragments somewhere later or in another section?

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+First, you will want to import `useState` from React like so:
```
gaearon: I feel like this section repeats the previous one. Maybe you can axe the previous one and just do this explanation.

docs/intro-react.md
gaearon: In general, Hooks are kind of a generalization of this. They let things "stick around".

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+const [name, setName] = useState("What's my name again?");
+```
+
+Then you add `onChangeText` to the `<TextInput>`. This fires every time the `<TextInput>`‘s value changes (every time a user adds a character of text).
+
+```jsx
+onChangeText={(text) => setName(text)}
+```
+
+This uses the `setName` you defined with the `useState` Hook above to update your state variable `defaultText` every time `onChangeText` fires. You might set state when you have new data from the server, or from user input, or even from a timer.
```
gaearon: The state variable is called `name`, not `defaultText`?

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+> You might’ve noticed that although `name` is a [const](https://developer.mozilla.org/Web/JavaScript/Reference/Statements/const), it is seemingly reassignable! What is happening is that every time a user inputs a letter into `<TextInput>`, `setName()` fires to update the state const `name`. When a state-setting function like `setName` is called, its component will re-render, running its function again. In this case `Cat` will re-render its `<Cat>`—discarding the old state const `name` in the process!
```
gaearon: This is good to call out.

docs/intro-react.md
gaearon: Or maybe they could be swapped around?

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+      <>
+        <Cat />
+        <Cat />
+      </>
+    );
+  }
+}
+```
+
+In class components, state is stored in a `state` object:
```
gaearon: Notice how much simpler the class section looks when you switch. This is not the impression we want to make.

docs/intro-react.md
```diff
@@ -0,0 +1,476 @@
...
+
+```jsx
+this.setState({name: text});
+```
+
+> Do not change your component's state directly by assigning it a new value with `this.state.name = text`. Calling `this.setState()` allows React to track changes made to state that trigger rerendering. Setting state directly can break your app's reactivity!
+
+<block class="endBlock syntax" />
+
+---
```
gaearon: In general, we might want to rethink this example. First, because you have the next chapter about TextInput anyway — but we use it earlier. Second, because text input involves explaining "controlled components" — like inputs with `value` prop. But most people find it counter-intuitive at first. What's the point of state if the input works "on its own" anyway? They expect input to remember its own state.

So an example that isn't an input might work better here. So that it's obvious *why* the component wants to "remember" something. This also ensures people who don't yet understand controlled components wouldn't struggle with state in general. State is simple.

docs/intro-react.md
rachelnabors: In the end, we did offer a class-based state example!

docs/intro-react.md
gaearon: In fact, "Handling Text Inputs" should probably explain the concept of a controlled component instead. And `value` vs `defaultValue`. It's ok if that's out of scope of this PR though.

docs/introduction.md
rachelnabors: So interestingly enough, it _was_ further down, but someone suggested it should be closer to the top for readers who want to set up their environment before before getting too far along. I tend to agree with you, though. I think the straighter flow is better. Let me see what Dan's feedback says :)

docs/introduction.md
rachelnabors: That would be a lovely thing. Maybe in a ticket for another push?

docs/intro-react-native-components.md
rachelnabors: "Some kinds of views can be composed of other views" feels clunky. What about "Some kinds of views can contain other views?" It's a little more common language.

rachelnabors: Re: "view", here I was trying to think of "the views that exist in iOS and Android development. We haven't quite gotten to introducing React Native's Core Component with a capital V: View (not quite yet). Thoughts?

docs/intro-react-native-components.md
rachelnabors: It should be bolded because each new Important Term needs a bolding, like in school texts. Good catch! I missed it!

rachelnabors: Dan, I worry that the phrasing you suggest pops right in with "components" before introducing the concept of components. We were _just_ talking about views and how both mobile platforms use them for everything in their apps and then suddenly "platforms provide components! We call them..." Wait. What. Is. A. Component. I worry we're skipping ahead a bit if we frontload with that. This is a common problem I noticed with folks unfamiliar with either views or components: gotta build a bridge to link the two explicitly.

docs/intro-react-native-components.md
rachelnabors: @cpojer It's at the bottom as a sort of visual endcap.

Does it add value? I found it useful working with Luna and Kirsty for explaining how all the components relate to each other.

I'm happy to polish it (originally intended to, but then I realized I could spend 5 hours making a diagram no one found useful and wanted to see what reviewers thought first! I am sneaky like this.)

docs/intro-react.md
rachelnabors: Should be placeholder, thanks!

docs/introduction.md
gaearon: I don’t care strongly either way. Trust you to choose this!

docs/intro-react-native-components.md
gaearon: I trust your choice here

docs/introduction.md
rachelnabors: Sloppy!

docs/introduction.md
rachelnabors: What the... that looks like something is broken!

docs/intro-react.md
rachelnabors: Sharp eyes! I'd like to know how you highlight lines of code in the React docs, as it would be great to also highlight the `export` line!

docs/intro-react.md
rachelnabors: Sloppy! Sharp eyes!

docs/intro-react.md
rachelnabors: This was leftover from a previous iteration. Thank you for catching this! Could have been so confusing.

rachelnabors: I'm actually going to collapse the example. I'm not sure it's necessary to show the same code twice!

docs/intro-react.md
rachelnabors: This gave me an excellent excuse to make a reference to the three names Jellicle cats possess. I thank you.

docs/intro-react.md
rachelnabors: Well.... I wanted to make custom art for each of these... but first I wanted to make sure that all the proofreaders liked the samples ;)

So... can I draw some cats?

docs/intro-react.md
rachelnabors: Let me have a think! I also get tripped up on this one. Maybe I can see how others are shorthanding it to their students...

docs/intro-react.md
rachelnabors: Thanks! I made an engineer friend who came to visit read these docs and he tripped on this one!

docs/intro-react.md
rachelnabors: Handling Text Inputs is out of the scope of this PR, as it's slated to be scooped up into a minitutorial. 

Trying to think of other things that we can use state for. I want to keep UI cruft to a minimum—which is why I reached for inputText. And no to timers—too contrived, and personally I know I and other trip on timers. Let me ponder.

docs/introduction.md
rachelnabors: Good to know, will do!

docs/intro-react-native-components.md
rachelnabors: Went with Christoph's edits. Let's watch learners in the wild and see if this gels with them!

website/i18n/en.json
rachelnabors: React Fundamentals

docs/intro-react.md
```diff
@@ -0,0 +1,540 @@
...
+}
+```
+
+<block class="functional classical webNote androidNote" />
+
+`Image` has [many different props](image#props), including [`style`](image#style), which accepts a JS object of design and layout related property-value pairs.
+
+> Notice the double curly braces `{{ }}` surrounding `style`‘s width and height. In JSX, JavaScript values are referenced with `{}`. However, JS object are **_also_** denoted with curly braces: `{width: 193, height: 110}`. Therefor, to pass a JS object in JSX, you must wrap the object in **another pair** of curly braces: `{{width: 193, height: 110}}`
+
+Your own components can also use props. Props let you customize your components. For example, here each `<MyPet>` passes a different `name` for the `MyPet` function to render:
```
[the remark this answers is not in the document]
rachelnabors: Changed to "For example, here you pass each `<Cat>` a different `name` for `Cat` to render:"

docs/intro-react.md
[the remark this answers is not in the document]
rachelnabors: Changed to "Each `<Cat>` renders a unique element—which you can customize with props."

docs/introduction.md
rachelnabors: I'm going to leave the dev environment stuff in a callout under the interactive examples. One reference should be enough.

docs/introduction.md
rachelnabors: Heh, I concede your point, sir!

rachelnabors: "With React, you can make components using either classes or functions. Originally, class components were the only components that could have state. But since the introduction of React's Hooks API, you can add state and more to function components.

[Hooks were introduced in React Native 0.58.](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059), and because Hooks are the future-facing way to write your React components, we wrote this introduction using function component examples. Where useful, we also cover class components under a toggle like so:"

docs/intro-react-native-components.md
TheSavior: I don't think this needs to be capitalized since it doesn't feel like referring to the React Native view, especially because it starts with "In Android and iOS development". 

I'm good with "can contain", it feels clearer than "can nest" or "can composed". 👍

rachelnabors: Thanks for clearing that up, Eli! Also, great job on your intro materials I stole from ;)

docs/intro-react-native-components.md
rachelnabors: So I moved things around a bit so that the section was a little tighter and introduced the idea of "hey, you can make your own Native Components—but to get started, we're gonna work with this of premade Native Components that come with RN—the Core Components!"

docs/intro-react.md
rachelnabors: How about "While you can think of props as arguments you use to configure how components render, state is like a component’s personal data storage. Sate is useful for handling data that changes over time or that comes from user interaction."

docs/intro-react.md
rachelnabors: I think my response above https://github.com/facebook/react-native-website/pull/1535#discussion_r367682836 should clear this up?

docs/intro-react.md
rachelnabors: I've had test learners appreciate this here. But the new example might not need it!

docs/intro-react.md
rachelnabors: Well, this is because we go into depth on what Hooks are and how this one works and the other example is there for reference. I can try to trim down the function component section and perhaps expand a bit on the class component

docs/intro-react.md
rachelnabors: I showed you the Cat Cafe and milk demo. I'm going to go with that one—you press on a button to feed the cats at the cat cafe.

docs/running-on-device.md
```diff
@@ -5,7 +5,7 @@ title: Running On Device
 
 It's always a good idea to test your app on an actual device before releasing it to your users. This document will guide you through the necessary steps to run your React Native app on a device and to get it ready for production.
 
-If you used Expo CLI or Create React Native App to set up your project, you can preview your app on a device by scanning the QR code with the Expo app. In order to build and run your app on a device, you will need to eject and install the native code dependencies from the [Getting Started guide](getting-started.md).
+If you used Expo CLI or Create React Native App to set up your project, you can preview your app on a device by scanning the QR code with the Expo app—but order to build and run your app on a device, you will need to eject and install the native code dependencies from the [environment setup guide](environment-setup.md).
```
TheSavior: typo in your change at  "with the Expo app-but order to build and run"

rachelnabors:
All right folks. Have a look now: 

* https://deploy-preview-1535--react-native.netlify.com/docs/next/introduction
* https://deploy-preview-1535--react-native.netlify.com/docs/next/intro-react-native-components (cool new images)
* https://deploy-preview-1535--react-native.netlify.com/docs/next/intro-react

docs/intro-react.md
rachelnabors: I drew some cats.

STYLEGUIDE.md
```diff
@@ -101,6 +101,14 @@ When navigating operating system menus, the menu selection path should be **bold
 
 **Good** “Open **Xcode > Preferences...** and select the **Components** tab.” Path is clearer to pick out from the surrounding copy.
 
+## “Android” before “iOS”
```
cpojer: I'm so happy to see this codified in our style guide.

website/static/css/docs.css
```diff
@@ -102,18 +102,17 @@
...
+  margin: 2em 0;
 }
 
 table {
   border: 1px solid #b0b0b0;
   border-collapse: collapse;
   display: table;
   margin: 20px 0;
-  width: 80%;
+  /* width: 80%; */
```
cpojer: Can you remove this? We try not to land commented out code. We either need it, or don't – and we can go back to it via source control.

website/static/css/docs.css
```diff
@@ -156,9 +155,19 @@ table tr th {
   font-family: "Helvetica Neue", Arial, sans-serif;
   text-transform: uppercase;
 }
+/* Nixed so tables on pages like getting-started look good
```
cpojer: Same here, just remove the commented out code.

website/static/css/docs-prevnext.css
```diff
@@ -53,7 +53,7 @@
 
 .button.docs-next {
   margin-left: auto;
-  margin-right: -1.6rem;
+  /* margin-right: -1.6rem; */
```
cpojer: remove please

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,69 @@
...
+---
+
+React Native is an open source framework for building Android and iOS applications using [React](https://reactjs.org/) and the app platform’s native capabilities. With React Native, you use JavaScript to access your platform’s APIs as well as to describe the appearance and behavior of your UI using React components: bundles of reusable, nestable code. You can learn more about React in the next section. But first, let’s cover how components work in React Native.
+
+## Views and mobile development
+
+In Android and iOS development, a **view** is the basic building block of UI: a small rectangular element on the screen which can be used to display text, images, or respond to user input. Even the smallest visual elements of an app, like a line of text or a button, are kinds of views. Some kinds of views can contain other views. It’s views all the way down!
+
+<figure>
+  <img src="../assets/diagram_ios-android-views.svg" width="1000" alt="Diagram of Android and iOS app showing them both built on top of atomic elements called views.">
```
cpojer: Don't forget to revert those paths again before merging.

docs/intro-react-native-components.md
```diff
@@ -0,0 +1,69 @@
...
+    </ScrollView>
+  );
+}
+```
+
+---
+
+Because React Native uses the same API structure as React components, you’ll need to understand React component APIs to get started. The [next section](intro-react) makes for a quick introduction or refresher on the topic. However, if you’re already familiar with React, feel free to [skip ahead](handling-text-input).
+
+<img src="../assets/diagram_react-native-components.svg" width="1000" alt="A diagram showing React Native's Core Components are a subset of React Components that ship with React Native.">
```
cpojer: This one too

cpojer:
Btw. you shared "https://deploy-preview-1535--react-native.netlify.com/docs/next/introduction" but the actual link is "https://deploy-preview-1535--react-native.netlify.com/docs/next/getting-started". Not sure why that is, but let's make sure the deployed site won't have any broken links to start.

rachelnabors:
> Btw. you shared "https://deploy-preview-1535--react-native.netlify.com/docs/next/introduction" but the actual link is "https://deploy-preview-1535--react-native.netlify.com/docs/next/getting-started". Not sure why that is, but let's make sure the deployed site won't have any broken links to start.

Nope, that was me hand-writing links—never advisable! >.<

brentvatne:
🎉 🎉 👍 👍

PR #1536, opened 2020-01-01
docs: Remove advice to enable Fast Refresh for new apps

pdsouza:
RN 0.61 enables Fast Refresh by default for new apps. There is no need to instruct users to manually enable it in the Getting Started guide.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -556,8 +556,6 @@ Congratulations! You've successfully run and modified your first React Native ap
 
 <h2>Now what?</h2>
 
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Good point! Since this change is only to the `/docs` directory, it won't change the 0.61 docs located in `website/versioned_docs/0.61`, it will only change the next version of the docs.

Feel free to follow up with another PR changing them there!

pdsouza:
Thanks for the info and quick merge! Looks like @Naturalclar has taken care of updating the 0.61 docs in #1538.

PR #1544, opened 2020-01-06
Remove ToolbarAndroid from doc

Naturalclar:
This PR is part of https://github.com/facebook/react-native/issues/26591#issuecomment-570972727

Removed docs for ToolbarAndroid as it has been removed from the package.


connectdotz:
I was late for this PR, noticed this deprecation is handled differently than the other lean core components, such as the`[slider](https://reactnative.dev/docs/slider), which remains on the doc but with the deprecate message...  Is this different approach intended?

PR #1545, opened 2020-01-06
Fix links for website

vvscode:
Fix links for http://facebook.github.io/react-native/docs/geolocation.html page


vvscode:
Page with the fix https://deploy-preview-1545--react-native.netlify.com/docs/geolocation

vvscode:
@cpojer  Actually I see a lot of such cases in docs. Was the idea to process it on deploy?

PR #1552, opened 2020-01-14
Removed margin-right from docs-next button

vinayd001:
Removed margin-right from docs-next button because it goes little bit out of the screen while working on mobile and it's wasn't as far as on the right side as compare to the Getting Started button from it's left.


rachelnabors:
We caught this recently with the new intro docs, but thanks so much for catching it! Please submit any time!

PR #1556, opened 2020-01-21
Update slider.md

sudokzt:
When I read the deprecated message in this document, I tried to use [react-native-community/react-native-slider](https://github.com/react-native-community/react-native-slider, but is not corresponded to Expo. 

This doc has not the message. So I wasted a lot of time.


cpojer:
Thank you for submitting a PR. I'm worried that this actually makes the docs harder to understand. I'll take a note and will think about changing this later.

sudokzt:
@cpojer 
that make sense. thank you for checking.

PR #1560, opened 2020-01-23
Added testID property

mateosilguero:


cpojer:
Thank you for sending this PR! I checked the docs for TextInput and it says that it inherits all the props from View, which already lists `testID`. See https://github.com/facebook/react-native-website/blob/master/docs/textinput.md#props

PR #1566, opened 2020-01-29
Tweak the first example

gaearon:
The description is a bit incorrect. It's not commented out: `//` comments don't work inside JSX text content. So they actually do appear in the text. This is text, not a comment.


rachelnabors:
Thanks for catching this one, Dan!

PR #1567, opened 2020-01-29
Small fixes to React guide

gaearon:
Misc.

### DIFF
--- a/docs/intro-react.md
+++ b/docs/intro-react.md
@@ -143,18 +143,21 @@ export default function Cat() {
 }
 ```
 
-Any JavaScript will work between curly braces, including functions like `{fullName("Rum", Tum", "Tugger")}`:
+Any JavaScript expression will work between curly braces, including function calls like `{getFullName("Rum", Tum", "Tugger")}`:
 
 ```SnackPlayer name=Curly%20Braces
 import React from 'react';
 import { Text } from 'react-native';
 
 export default function Cat() {
-  function fullName(firstName, secondName, thirdName) {
+  function getFullName(firstName, secondName, thirdName) {
     return firstName + " " + secondName + " " + thirdName;
   }
+
   return (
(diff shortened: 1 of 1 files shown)


docs/intro-react.md
```diff
@@ -143,18 +143,21 @@ export default function Cat() {
 }
 ```
 
-Any JavaScript will work between curly braces, including functions like `{fullName("Rum", Tum", "Tugger")}`:
+Any JavaScript expression will work between curly braces, including function calls like `{getFullName("Rum", Tum", "Tugger")}`:
```
gaearon: The convention for function that return something is `get*`

docs/intro-react.md
```diff
@@ -143,18 +143,21 @@ export default function Cat() {
...
 ```SnackPlayer name=Curly%20Braces
 import React from 'react';
 import { Text } from 'react-native';
 
 export default function Cat() {
-  function fullName(firstName, secondName, thirdName) {
+  function getFullName(firstName, secondName, thirdName) {
     return firstName + " " + secondName + " " + thirdName;
   }
+
```
gaearon: Some spacing between function definitions and other code.

docs/intro-react.md
```diff
@@ -354,7 +357,7 @@ import React, {useState} from 'react';
 
 Then you declare the component’s state by calling `useState` inside its function. In this example, `useState` creates an `isHungry` state variable:
 
-```JS
+```jsx
```
gaearon: I think this should fix highlighting

docs/intro-react.md
rachelnabors: I almost feel like that would make a great

> note.

gaearon: This is more of a JS convention — can't really teach every one of them here! The reason for it is that unless you look at the definition, `fullName` actually appears to be a name itself. `getFullName` is more clearly a function.

PR #1574, opened 2020-01-31
Clarify Vibration iOS API

hramos:
Addresses https://github.com/facebook/react-native/issues/27920. Vibration on iOS only supports a plain `Vibration.vibrate()` call. 

Vibrating by pattern is not supported: https://github.com/facebook/react-native/blob/2f758c3abbef33a47cd3651e7213066f53fb373f/Libraries/Vibration/RCTVibration.mm#L39-L48 https://github.com/facebook/react-native/blob/2f758c3abbef33a47cd3651e7213066f53fb373f/Libraries/Vibration/RCTVibration.mm#L28-L31

Canceling vibration is not supported: https://github.com/facebook/react-native/blob/2f758c3abbef33a47cd3651e7213066f53fb373f/Libraries/Vibration/RCTVibration.mm#L39-L48

Intentionally used "webNote" for Android notes to force Android notes to show up by default. If no webNote is present, the notes block will default to empty as displaytabs() will hide everything that is not a webNote.

Test Plan:

```
yarn
yarn start
```

Verified Vibration API docs look good.

### DIFF
--- a/docs/vibration.md
+++ b/docs/vibration.md
@@ -3,57 +3,41 @@ id: vibration
 title: Vibration
 ---
 
-The Vibration API is exposed at `Vibration.vibrate()`. The vibration is synchronous so this method will return immediately.
+The Vibration API is exposed at `Vibration.vibrate()`. The vibration is synchronous so this method will return immediately. There will be no effect on devices that do not support vibration, such as the iOS Simulator or Android emulator.
 
-There will be no effect on devices that do not support vibration, such as the iOS Simulator or Android emulator.
+The `vibrate()` method can take a duration or pattern argument. On Android, if `pattern` is a number, it specifies the vibration duration in milliseconds. If `pattern` is an array, those odd indices are the vibration duration, while the even ones are the separation time. On iOS, this argument will have no effect as the `vibrate()` method invokes a simple `AudioServicesPlaySystemSound(kSystemSoundID_Vibrate)` call (e.g. a fixed time vibration).
 
-<div class="toggler">
-  <span>Developer Notes</span>
-  <span role="tablist" class="toggle-devNotes">
-    <button role="tab" class="button-webNote" onclick="displayTabs('devNotes', 'webNote')">Android</button>
-    <button role="tab" class="button-iosNote" onclick="displayTabs('devNotes', 'iosNote')">iOS</button>
-  </span>
-</div>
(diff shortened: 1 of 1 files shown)


rachelnabors:
Interesting use of the Notes! That behavior of "defaulting" is an issue with the JS I need to fix in a separate PR. 

What I'm wondering is if this is the right use for the Notes feature? Originally the Notes were intended to show information to folks with specific dev backgrounds: a person viewing iOS Notes would gain knowledge specific to their background ("In iOS, you might be familiar with this concept, but here it is this instead."), not as a callout to gotchas specific to different platforms (imagine if we needed to put a note in there for iOS and Android platform implementation gotchas—could a user miss an important bit of information like that?).

What do you think, Hector? Is there another way to share that information? Or is this (platform specific gotchas) a common pattern we should seek to address better in the docs design?

hramos:
I don't think this is the right use of Notes. I ended up coupling together actual Notes (the Android Manifest callout) with code Examples in the same "Notes block". I would like to have some similar control for code samples, though, that provides a platform (or language!) switcher. Basically, a Snack player without the app preview, since the code sample would not necessarily contain all the necessary code to build a screen.

I'll remove the Notes from this PR.

hramos:
I feel much better about this rewrite, as it looks cleaner to me.

Screen Shot 2020-02-04 at 12 59 53 PM

hramos:
I felt it was necessary to call out that this maps to `AudioServicesPlaySystemSound(kSystemSoundID_Vibrate)` on iOS, as that's a common way of triggering short vibrations on iOS natively, and it also implicitly lets readers that are more familiar with iOS know that we are not currently using the newer, more configurable APIs that make use of the Taptic Engine.

hramos:
Please don't merge this yet. While investigating https://github.com/facebook/react-native/issues/23404, I found that iOS does support vibration by pattern, it's just implemented in Vibration.js. I'll update this PR tomorrow.

hramos:
@rachelnabors the API docs should now reflect the actual behavior on iOS.

rachelnabors:
Alola! Having a dip into this today. I was thinking... we could one day do a platform-based switcher like the one we use for Function/Class Components... You could put just about anything under there: https://reactnative.dev/docs/next/intro-react#your-first-component

Would something like this work well for your platform-specific code and gotchas? Or would it make more sense to put all under headings?

PR #1578, opened 2020-02-04
Update docs for React Navigation 5

satya164:
The current documentation refers to React Navigation 4. We're about to publish new major version of React Navigation, so I have updated the documentation.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -21,7 +21,7 @@ First, you need to install them in your project:
 npm install @react-navigation/native @react-navigation/stack
 ```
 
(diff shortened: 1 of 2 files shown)


rachelnabors:
Oooh, what a noble update! Have you read our [contribution styleguide](https://github.com/facebook/react-native-website/blob/master/STYLEGUIDE.md)? Things like using "you" to refer to the reader instead of "we" (which refers to React Native as an entity and the people writing the docs!) Would be awesome if you could check this guide and do another pass!

satya164:
Oops! I'll take a look and update. Thank you!

Issue #1579, opened 2020-02-05
Component Docs Drive


rachelnabors:
**🍾🎉🎈Thank you everyone! We are mission complete!🍾🎉🎈**

We have a lot of component and API docs that need updating! The good news is that updating these docs can be a fun way to become more familiar with React Native’s core code while contributing to docs that help thousands of learners around the world! If this looks intimidating, we have you covered! There’s a handy guide below that should help you out :)

## What needs updating?

* Update the API to reflect the props and methods reflected in core
* Ensure example code is 1) in a Snack player and 2) uses function components and Hooks:
    * Add example code if it is missing
    * Update example code to use function components and React Hooks
    * Put example code in an embedded Snack player

## How to contribute?

1. First, select a component or API from the table below. 
2. Choose a single update to make:
    1. Update the API
    2. Add missing example code
    3. Convert existing example code to use function components and React Hooks
    4. Embed any existing example code in a Snack player
3. Comment below on what component/API you want to take on and which fix; I'll update the matrix with your name
4. Submit your PR tagged with this issue

When your PR is merged, we’ll convert your name into a checkmark, indicating completion!

### Update the API to reflect the props and methods reflected in core

This part requires a little investigating! Check [react-native/Libraries/](https://github.com/facebook/react-native/tree/9caf99162dcd1ee73c571c494ca3115540f45f69/Libraries)YourComponent and** **[react-native/Libraries/Components/](https://github.com/facebook/react-native/tree/9caf99162dcd1ee73c571c494ca3115540f45f69/Libraries/Components)YourComponent for your component or API’s list of props. Sometimes they can be hidden, like the [LayoutProps API](https://facebook.github.io/react-native/docs/layout-props) is in [react-native/Libraries/StyleSheet/StyleSheetTypes.js](https://github.com/facebook/react-native/blob/8553e1acc4195479190971cc7a3ffaa0ed37a5e0/Libraries/StyleSheet/StyleSheetTypes.js). 

> **Pro tip:** If you get stuck, search for “YourComponent.js” or search the “YourComponent” with a prop or method already listed.

### Add missing example code

If there is no example code, add a new example using [function components](https://reactjs.org/docs/hooks-state.html#hooks-and-function-components) and [Hooks](https://reactjs.org/docs/hooks-intro.html). Embed your code with a Snack player like so:

```` 
```SnackPlayer name=Component/API%20Example

// Example code goes here

``` 
````

### Convert existing example code to use function components and Hooks

[...]

PR #1493, opened 2019-12-03
deprecation indicator and deprecation notes standardization

Simek:
By deprecation notes standardization I have meant following changes:
* using current NPM package name
* using similar phrasing
* positioning deprecation note always on the top of the page

The rest of changes are basically a concept, which you might not like. 

I was aiming to clarify deprecation and discourage developers to use those parts of RN so I have added  🚧emoji to the title of pages marked as deprecated Component or API. (Of course I'm open to change selected emoji, this could be anything else.)

Screenshot 2019-12-03 at 14 59 41

I can work further on this PR or on any other solution that help in the case mentioned above.

Issue #1045, opened 2019-06-19
Proposal: Don't completely remove pages for Lean Core removed components


rickhanlonii:
## Overview

We currently have a few open and merged PRs which are removing the documentation pages for components removed in the Lean Core effort.

I see a few issues with this:
- **Confusion**: Users that have been checking the React Native site for these docs will suddenly not be able to find them
- **Discovery**: Users that are on an old version of React Native will not know to check the docs of the version they're on that has the component
- **Loss of SEO**: these pages are the top result when searching for these components

## Proposal
To fix, I propose that we
- Remove these docs from the sidebars
- Replace the docs with shell pages that explain where the component how lives and a link to docs of the last React Native version that includes the component 

```
ImageEditor has been moved to https://github.com/react-native-community/react-native-image-editor

For the last React Native version, see: https://facebook.github.io/react-native/docs/0.58/imageeditor
```

This way, they won't clutter the sidebar and users looking for this component organically or from Google will be able to still discover it and be routed to the correct home.

ecreeth:
I think it would be good to create a section called **Deprecated Components** where it is explained why it was decided to move these components and where they can find them.

or a sub-category

rickhanlonii:
That may be useful, but we still would want to create individual pages for the old URLs so when people land there they can go to the right place

PR #1583, opened 2020-02-05
DatePickerIOS - Added Snack Examples 🍿

DaniAkash:
Belongs to  facebook/react-native-website#1579

Created a nice little snack for people to view DatePickerIOS in action using both Function & Class Component examples.

### DIFF
--- a/docs/datepickerios.md
+++ b/docs/datepickerios.md
@@ -24,7 +24,7 @@ Use `DatePickerIOS` to render a date/time picker (selector) on iOS. This is a co
 
 ```SnackPlayer name=DatePickerIOS&supportedPlatforms=ios
 import React, {useState} from 'react';
-import {DatePickerIOS, View, StyleSheet, Text, Platform} from 'react-native';
+import {DatePickerIOS, View, StyleSheet} from 'react-native';
 
 export default function App () {
 
@@ -52,7 +52,7 @@ const styles = StyleSheet.create({
 
 ```SnackPlayer name=DatePickerIOS&supportedPlatforms=ios
 import React, {Component} from 'react';
-import {DatePickerIOS, View, StyleSheet, Text, Platform} from 'react-native';
+import {DatePickerIOS, View, StyleSheet} from 'react-native';
 
 export default class App extends Component {
   constructor(props) {
@@ -66,6 +66,7 @@ export default class App extends Component {
     this.setState({chosenDate: newDate});
   }
 
+  render() {
     return (
       <View style={styles.container}>
         <DatePickerIOS
--- a/docs/datepickerios.md
+++ b/docs/datepickerios.md
@@ -22,21 +22,13 @@ Use `DatePickerIOS` to render a date/time picker (selector) on iOS. This is a co
 
 <block class="functional syntax" />
 
-```SnackPlayer name=DatePickerIOS
+```SnackPlayer name=DatePickerIOS&supportedPlatforms=ios
 import React, {useState} from 'react';
 import {DatePickerIOS, View, StyleSheet, Text, Platform} from 'react-native';
 
 export default function App () {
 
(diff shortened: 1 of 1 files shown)


docs/datepickerios.md
```diff
@@ -9,9 +9,50 @@ Use `DatePickerIOS` to render a date/time picker (selector) on iOS. This is a co
...
+      Class Component Example
+    </li>
+  </ul>
+</div>
+
+<block class="functional syntax" />
+
+```SnackPlayer name=DatePickerIOS&supportedPlatforms=ios
+import React, {useState} from 'react';
+import {DatePickerIOS, View, StyleSheet, Text, Platform} from 'react-native';
```
Simek: `Platform` and `Text` imports are unnecessary after af16849a7ffc97bb402c117867385cd2084d09a0 commit.

docs/datepickerios.md
```diff
@@ -9,9 +9,50 @@ Use `DatePickerIOS` to render a date/time picker (selector) on iOS. This is a co
...
+  },
+});
+```
+
+<block class="classical syntax" />
+
+```SnackPlayer name=DatePickerIOS&supportedPlatforms=ios
 import React, {Component} from 'react';
-import {DatePickerIOS, View, StyleSheet} from 'react-native';
+import {DatePickerIOS, View, StyleSheet, Text, Platform} from 'react-native';
```
Simek: `Platform` and `Text` imports are unnecessary after af16849a7ffc97bb402c117867385cd2084d09a0 commit.

Simek:
I'm not sure if adding an example for the the deprecated components is a good idea. 

They will be sooner or later completely removed form the codebase and not providing examples for them could be a good way to discourage usage of those components.

docs/datepickerios.md
DaniAkash: Have removed these imports

docs/datepickerios.md
DaniAkash: have removed them

DaniAkash:
> I'm not sure if adding an example for the the deprecated component is a good idea.
> 
> They will be sooner or later completely removed form the codebase and not providing examples for them could be a good way to discourage usage of those components.

@Simek I'll leave the decision to the maintainers 👍

Simek:
It seems that Rachel got similar concerns to my: https://github.com/facebook/react-native-website/issues/1579#issuecomment-582448869

Also after thinking about this a bit more I would like to start discussion about removing all examples form the deprecated components after #1493 will be merged.

rachelnabors:
> Also after thinking about this a bit more I would like to start discussion about removing **all** examples from the deprecated components after #1493 will be merged and before next major version (0.61 docs and below will remain untouched).

Actually we want to remove all the deprecated components from the sidebar for 0.62. Want to do it for me?

rickhanlonii:
Keep in mind https://github.com/facebook/react-native-website/issues/1045 while removing pages - if people have bookmarks or page ranked search results we should guide them to the new location for awhile.

Simek:
@rickhanlonii I think that @rachelnabor want only to remove  entries for the deprecated components from the Sidebar and leave pages intact under versioned `0.62` urls.

@rachelnabors I can help with that, no problem. It should help if you describe exactly what is desirable effect of that change. Currently I'm assuming that you want me to to generate versioned docs for `0.62` (they do not exist ATM) and remove those pages ONLY form the Sidebar in there. Is that right?

rachelnabors:
@rickhanlonii So @Simek has me right here—would be great to just remove the links from the sidebar from 62 onward. Editing only **sidebars.json** should change this in **/docs/next/** and all versions going forward—which is what we want! I'd submit a PR for that very thing, @Simek :)

rickhanlonii:
Yup that works.

PR #1587, opened 2020-02-05
update example code to function component

elderferreiras:
#1579


rachelnabors:
Looks like another PR was merged first—but thank you for your work all the same! Please check #1579 for another component that needs love!

PR #1592, opened 2020-02-05
Button: Update code snippet and example

dsznajder:
Issue: #1579 

Seems like example is already using function component :)


rachelnabors:
Thanks for the update ;)

PR #1599, opened 2020-02-05
📋 Clipboard - added snack examples [WIP] 👷🏽‍♂️👨🏽‍💻🕵🏽‍♂️

DaniAkash:

Created snacks with both functional and class components. However, unable to get the snack working on the expo web player. Clipboard is implemented in react-native-web, hence could use a little help here...

Current issues preventing this PR from merging into the master:
- [ ] Unable to use clipboard on web snack 
  - Fix: Disabling web option in the snack (https://github.com/react-native-community/react-native-clipboard/issues/5) 
  - update: The above fix isn't the required fix ☹️

### DIFF
--- a/docs/clipboard.md
+++ b/docs/clipboard.md
@@ -7,6 +7,111 @@ title: Clipboard
 
 ---
 
+### Example
+
+<div class="toggler">
+  <ul role="tablist" class="toggle-syntax">
+    <li id="functional" class="button-functional" aria-selected="false" role="tab" tabindex="0" aria-controls="functionaltab" onclick="displayTabs('syntax', 'functional')">
+      Function Component Example
+    </li>
+    <li id="classical" class="button-classical" aria-selected="false" role="tab" tabindex="0" aria-controls="classicaltab" onclick="displayTabs('syntax', 'classical')">
+      Class Component Example
+    </li>
+  </ul>
+</div>
+
+<block class="functional syntax" />
+
+```SnackPlayer name=Clipboard
+import React, {useState} from 'react';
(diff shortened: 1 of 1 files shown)


rachelnabors:
I suspect it has to do with react native web: https://github.com/react-native-community/react-native-clipboard/issues/5

It's possible to configure it so the web version doesn't show in the Snack... `&supportedPlatforms=ios,android`

rachelnabors:
Turns out React Native Web says it _does_ support the clipboard: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Clipboard/index.js

Can you investigate?

DaniAkash:
> Turns out React Native Web says it _does_ support the clipboard: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Clipboard/index.js
> 
> Can you investigate?

I tried running it in my local expo-web setup and it was working fine, for some reason it is not working within the embedded player. Does the player share its clipboard with the system?

rachelnabors:
Let me ask @brentvatne!

rachelnabors:
Someone else submitted a Clipboard example. Let's drop it for now as it looks like the API is being deprecated. Thank you for your hard work!

brentvatne:
in the web example, `getString()` will not work because the implementation it depends on is as follows:

https://github.com/necolas/react-native-web/blob/a7ab961d95de892e4b41c96ea945aa344fd49e50/packages/react-native-web/src/exports/Clipboard/index.js#L23-L25

`setString` should work, and it does in my tests using the same example code: https://snack.expo.io/@notbrent/clipboard-example

on ios and android things work as expected :)

PR #1600, opened 2020-02-05
⚫️ Dark Mode: Add Appearance module documentation 

hramos:
⬛️ Documents the Appearance module that was added back in August 2019 in https://github.com/facebook/react-native/commit/63fa3f21c5ab308def450bffb22054241a8842ef#diff-66ff56fe20c3cf9819920211fb38c9bc, and which will be part of the upcoming 0.62 release.

🌑 The Appearance API has been used by Instagram in production since October 2019 when they launched  [Dark theme](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme) and [Dark Mode](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/dark-mode/) support in their Android and iOS apps.

Screen Shot 2020-02-05 at 2 50 07 PM

Screen Shot 2020-02-05 at 2 50 11 PM

🌚 Trivial: removed unnecessary space in Dimensions doc.

## Test Plan:

```
yarn
cd website
yarn start
```

🖤 Prettier applied some changes as part of the pre-commit hook, which I am including in this PR.


hramos:
@rachelnabors good catch! I just updated my other PR, https://github.com/facebook/react-native-website/pull/1574, with a Snack example.

Looking at this Appearance PR, I think the Snack can be added later because this API is added in 0.62, which as of this writing has not yet reached stable and therefore it's not supported in Expo or Expo's Snack.

Do you think we can merge this as is, and I can make sure to add the Snack when it's supported? It will likely be based on https://github.com/facebook/react-native/blob/master/RNTester/js/examples/Appearance/AppearanceExample.js, which is linked to from the docs.

hramos:
Hope it's ok if I merge this! I want to make sure it's documented in 0.62.

PR #1601, opened 2020-02-06
Dimensions 📐 - updated Docs 📑 & new snack examples ✨

DaniAkash:

Inspiration - To address this [stackoverflow question](https://stackoverflow.com/questions/44978804/whats-the-difference-between-window-and-screen-in-the-dimensions-api)

Contains the snack with both class & functional component examples and also a bit of update to the documentation regarding screen, window & event listener.

Pending items in this PR:
- [ ] `useDimensions` hook
- [ ] Info regarding scaling
- [ ] Verify if rounding is necessary for using Dimensions

### DIFF
--- a/docs/dimensions.md
+++ b/docs/dimensions.md
@@ -2,16 +2,129 @@
 id: dimensions
 title: Dimensions
 ---
-```jsx
-import {Dimensions } from "react-native";
-```
-You can get device width and height using below :
 
-Get device screen width and height :
 ```jsx
-const screenWidth = Math.round(Dimensions.get('window').width);
-const screenHeight = Math.round(Dimensions.get('window').height);
+import {Dimensions} from 'react-native';
 ```
+
+You can get the application window's width and height using below code:
+
+```jsx
+const windowWidth = Math.round(Dimensions.get('window').width);
+const windowHeight = Math.round(Dimensions.get('window').height);
+```
+
+If you are targeting foldable devices or devices which can change the screen size or app window size, you can use the event listener available in the Dimensions module as shown in the below example.
+
+### Example
+
+<div class="toggler">
+  <ul role="tablist" class="toggle-syntax">
+    <li id="functional" class="button-functional" aria-selected="false" role="tab" tabindex="0" aria-controls="functionaltab" onclick="displayTabs('syntax', 'functional')">
+      Function Component Example
+    </li>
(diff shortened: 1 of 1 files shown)


docs/dimensions.md
```diff
@@ -2,16 +2,129 @@
...
+    alignItems: "center"
+  }
+});
 ```
+
+<block class="endBlock syntax" />
+
+### React Native Hooks
+
+You can also try [useDimensions](https://github.com/react-native-community/react-native-hooks#usedimensions) hook from [React native hooks](https://github.com/react-native-community/react-native-hooks) library which makes handling screen/window size changes much simpler.
```
jeremy-deutsch: RN actually ships its own `useWindowDimensions` hook now:
https://github.com/facebook/react-native/blob/8d57691a606ce81e6d861f93ccc11875088f58b6/Libraries/Utilities/useWindowDimensions.js

The docs should probably describe that hook here instead

DaniAkash: Thanks, I didn't knew!

docs/dimensions.md
```diff
@@ -42,9 +157,9 @@ Example: `var {height, width} = Dimensions.get('window');`
 
 **Parameters:**
 
-| Name      | Type     | Required | Description                                                                                   |
-| ------    | ------   | -------- | ----------------------------------------------------------------------------------------------|
-| dim       | string   | Yes      | Name of dimension as defined when calling `set`. @returns {Object?} Value for the dimension.  | 
+| Name | Type   | Required | Description                                                                                  |
+| ---- | ------ | -------- | -------------------------------------------------------------------------------------------- |
+| dim  | string | Yes      | Name of dimension as defined when calling `set`. @returns {Object?} Value for the dimension. |
```
jeremy-deutsch:
```suggestion
| dim  | string | Yes      | Name of dimension as defined when calling `set`. |

Returns an object with four number fields: `width`, `height`, `scale`, and `fontScale`.
```

`scale` and `numberScale` should maybe be documented better here, but honestly, I don't really know much about them!

DaniAkash: Indeed, these areas could use some more detail. I'll try them out and update the docs

Simek:
In the first example documentation is suggesting rounding `Dimensions` values before usage. This part was added in #1274. I don't know the exact reason behind this but I would assume that in some cases/on some devices this value could return float. 

Currently proposed examples do not utilize `Math.round`. I'm not sure which approach to handling those values is better but I think that someone should decide about that for the consistency and after that some parts of code should be updated.

DaniAkash:
Closing it in favour of https://github.com/facebook/react-native-website/pull/1671

PR #1602, opened 2020-02-06
docs: add Image Style Props examples

ronnyiscoding:
Issue #1579 Add Image Style Props examples.


website/i18n/en.json
```diff
@@ -341,9 +341,6 @@
       "toastandroid": {
         "title": "ToastAndroid"
       },
-      "toolbarandroid": {
```
Simek: Am I assuming correctly that this change was added to the commit by mistake?

ronnyiscoding: Yes, I will remove it

Simek: So it seems that someone forget to add a change generated after building app to their PR which landed on `master`. Regardless that this change should not be a part of this PR in my opinion.

ronnyiscoding: PR update https://github.com/facebook/react-native-website/pull/1602/commits/8d87b86c9943022a26b5e9bba6bead7c173f8c6b changes restored

rachelnabors:
@ronnyiscoding My last concern: have you updated the API?

PR #1604, opened 2020-02-06
added clipboard example

brunokiafuka:
#1579,

Added example using hooks to clipboard API.

### DIFF
--- a/docs/clipboard.md
+++ b/docs/clipboard.md
@@ -11,36 +11,38 @@ title: Clipboard
 ## Example
 
 ```SnackPlayer name=Clipboard%20API%20Example
-import React, { useState } from "react";
-import { SafeAreaView, View, Text, TouchableOpacity, Clipboard } from "react-native";
+
+import React, { useState } from "react"
+import { SafeAreaView, View, Text, TouchableOpacity, Clipboard } from "react-native"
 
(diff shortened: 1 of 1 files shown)


docs/clipboard.md
```diff
@@ -7,6 +7,44 @@ title: Clipboard
 
 ---
 
+
+## Example
+
+```SnackPlayer name=Clipboard%20API%20Example
```
Simek: I think that `Clipboard` would be a sufficient name 😉

brunokiafuka:
Hi @rachelnabors and @Simek, I hope you are all doing well.
what would be the next steps based on #1608.

Simek:
@brunokiafuka Hi and thank you for your contribution. Currently we need to wait for maintainers decision, please check out my comment for the more details: https://github.com/facebook/react-native-website/issues/1579#issuecomment-583735028

brunokiafuka:
@rachelnabors the clipboard API get method only works on android and ios, on web, it can set the item to the clipboard but returns an empty string yet if you try to paste the copied content elsewhere it will work. 

> Clipboard gives you an interface for setting and getting content from Clipboard on both Android and iOS. 
👆🏾 API intro line.

PR #1605, opened 2020-02-06
docs: Modal updated API and examples

Darking360:
# Modal new examples and API updated check

## Issue

https://github.com/facebook/react-native-website/issues/1579

# Checks

- [x] Check updated API for Modal (compared against properties in React Native core)
- [x] Add examples for functional and Hooks based components.
- [x] Embed any existing example code in a Snack player

## Messages

- Adding new code blocks as shown by @rachelnabors
- Adding better styling for example
- Both function and class example added

### DIFF
--- a/docs/modal.md
+++ b/docs/modal.md
@@ -10,29 +10,38 @@ The Modal component is a basic way to present content above an enclosing view.
 <block class="functional syntax" />
 
 ```SnackPlayer name=rn-modal-function-example
-import React, {Component, useState} from 'react';
-import {Modal, Text, TouchableHighlight, View, Alert, StyleSheet} from 'react-native';
+import React, { Component, useState } from "react";
+import {
+  Modal,
+  Text,
+  TouchableHighlight,
+  View,
+  Alert,
+  StyleSheet
+} from "react-native";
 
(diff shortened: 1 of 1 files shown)


Darking360:
@rachelnabors I opened an issue inside Snack Web to check why it's not working for web, and someone commented an error from the minified version, maybe a `terser` bug, the issue is here: https://github.com/expo/snack-web/issues/34

I was checking on the code implementation for `RCTModalHostView` from React Native core but I got lost 😅 I'd say to wait until someone from Expo answers the issue maybe? 🤔

rachelnabors:
Gonna merge and CC @brentvatne for future reference!

rachelnabors:
Thank you for your contribution :)
mlp_ur-the-best

tcdavis:
I believe that Modal is not supported in RNWeb currently. https://github.com/necolas/react-native-web/issues/1020
You may want to restrict the supported platforms to just Android and iOS :/

brentvatne:
@rachelnabors - the above is true, see https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Modal/index.js

PR #1608, opened 2020-02-07
docs: add deprecation notice for clipboard

Naturalclar:
Adding deprecation notice for the `Clipboard` module, as it is now moved to the [community](https://github.com/react-native-community/react-native-clipboard)

Ref: https://github.com/facebook/react-native/blob/master/index.js#L301-L309

### DIFF
--- a/docs/clipboard.md
+++ b/docs/clipboard.md
@@ -1,7 +1,9 @@
 ---
 id: clipboard
-title: Clipboard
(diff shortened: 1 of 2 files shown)


Simek:
You should also update `website/i18n/en.json` file and add deprecation emoji to the `title` field. Strings from this file are used in the sidebar.

PR #1613, opened 2020-02-09
🔷🔹 Transforms 🔸🔶 - Added Snack Examples ▫️◻️⬜️

DaniAkash:

Contains the Snack Examples and some details for using Transform style properties. 

However, the skew property on Android isn't working, so I'll mark it as a pending item.

Pending items in this PR:
- [ ] Skew transform not working on Android ☹️ https://github.com/facebook/react-native/issues/27649

### DIFF
--- a/docs/transforms.md
+++ b/docs/transforms.md
@@ -3,6 +3,229 @@ id: transforms
 title: Transforms
 ---
 
+Transforms are style properties that will help you modify the appearance and position of your components using 2D or 3D transformations. However, once you apply transforms, the layouts remain the same around the transformed component hence it might overlap with the nearby components. You can apply margin to the transformed component or the nearby components to prevent such overlaps. The following example increases the margin of the transformed component whenever it overlaps with the nearby components after a transform.
+
+## Example
+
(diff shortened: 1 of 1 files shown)


DaniAkash:
> So exhaustive! Thank you so _so_ much!! A few thoughts:
> 
>     1. I don't _normally_ ask to put styles inline (quite the opposite!), but here I think it might be handy to inline just the transform styles on the views they impact.
> 
>     2. What about moving the labels _inside_ the transformed objects? That would help associate the styles and explanations a little more closely!

Having styles inline will help here. I'll edit that!

But I'm kinda worried that putting the label inside the transforming boxes for skew & rotate transforms might make the text unreadable...

rachelnabors:
@DaniAkash Hm... Give it a try? See how it looks!

DaniAkash:
> @DaniAkash Hm... Give it a try? See how it looks!

I did and it looks kewl ✨

image

rachelnabors:
I'm giving you bonus points for those cats. Take these diamonds!
harley-diamonds

PR #1615, opened 2020-02-09
TextInput: Added new docs for PasswordRules Prop

espipj:
This prop which was added on [here](https://github.com/facebook/react-native/pull/25407) wasn't documented in the docs nor in the API. Related with #1579

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -756,6 +756,17 @@ Possible values for `textContentType` are:
 
 ---
 
+### `passwordRules`
+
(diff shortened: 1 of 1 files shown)


Naturalclar:
there was a typo in the `passwordRules` prop, which will be fixed in this PR https://github.com/facebook/react-native/pull/27999.
The document should also reflect the change

espipj:
Hey @Naturalclar it seems that your PR is pointing to react-native core repo but this PR is making changes in the docs website itself. I'll update the prop name though, thanks

PR #1619, opened 2020-02-10
added Keyboard API example using function components & hooks

ahmdtalat:
I added a `function` component example with hooks beside the `class` component example


kaioduarte:
Hi @ahmdtalat, just checked your snacks and they are not working because it's missing `export default` in both examples :sweat_smile: (check Netlify url above). I'd suggest also to add styles to them, it was hard to find the `TextInput`.

ahmdtalat:
@kaioduarte thanks 
@rachelnabors I've added snacks to the examples, but there is a problem, on Android, it works fine but on IOS there is a weird behavior and I don't why?

**problem**
when I click on the TextInput, the keyboard shows and hides automatically with the alerts fired.

PR #1623, opened 2020-02-11
added functional component examples with hooks

tsawan:
Hi, I've added functional component examples with hooks for Text component. #1579

### DIFF
--- a/docs/text.md
+++ b/docs/text.md
@@ -9,7 +9,61 @@ A React component for displaying text.
 
 In the following example, the nested title and body text will inherit the `fontFamily` from `styles.baseText`, but the title provides its own additional styles. The title and body will stack on top of each other on account of the literal newlines:
 
-```SnackPlayer name=Text
+<div class="toggler">
+  <ul role="tablist" class="toggle-syntax">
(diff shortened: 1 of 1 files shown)


docs/text.md
```diff
@@ -9,7 +9,61 @@ A React component for displaying text.
 
 In the following example, the nested title and body text will inherit the `fontFamily` from `styles.baseText`, but the title provides its own additional styles. The title and body will stack on top of each other on account of the literal newlines:
```
lcelso: @Simek @rachelnabors @tsawan 

Do you think that here we can have examples of how to use each props?

rachelnabors: Re:
> Do you think that here we can have examples of how to use each props?

I'd love to see that in the main example! Is it possible, @tsawan? Perhaps in a follow up submit?

tsawan:
sorry @rachelnabors I was away for a while.
I am not clear what do you mean by "show off each property in action". Please elaborate a bit about properties and I will try to look into it.

I also remember running prettier but I've noticed you prettifying snack example to change it to double quotes. Do you have a different prettier config from the one in repo.?

PR #1626, opened 2020-02-11
Update Exampe stylesheet

lcelso:
I have a suggestion to make, both `absoluteFill` and` absoluteFillObject` do the same thing. Do you still think it necessary to keep this in the documentation?

Can't we force any of them to become obsolete?
Does it make sense to have two methods that do the same thing?

#1579


docs/stylesheet.md
```diff
@@ -1,18 +1,31 @@
 ---
-id: stylesheet
 title: StyleSheet
+original_id: stylesheet
```
ecreeth: The page cannot be referenced. I think this should be `id: stylesheet` and move it up

lcelso: Done

PR #1627, opened 2020-02-12
Update Example state.md with Hooks

luism3861:
i updated only the example state to hooks!!

### DIFF
--- a/docs/state.md
+++ b/docs/state.md
@@ -13,14 +13,14 @@ For example, let's say we want to make text that blinks all the time. The text i
 import React, { useState, useEffect } from 'react';
 import { Text, View } from 'react-native';
 
-function Blink(props) {
-  const [isShowingText,setIsShowingText] = useState(true);
+const Blink = (props) => {
+  const [isShowingText, setIsShowingText] = useState(true);
 
    useEffect(() => {
-     // Toggle the state every second
      const toggle = setInterval(() => {
-     setIsShowingText(!isShowingText);
+       setIsShowingText(!isShowingText);
      }, 1000);
+
      return () => clearInterval(toggle);
   })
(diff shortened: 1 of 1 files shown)


docs/state.md
```diff
@@ -10,36 +10,30 @@ In general, you should initialize `state` in the constructor, and then call `set
 For example, let's say we want to make text that blinks all the time. The text itself gets set once when the blinking component gets created, so the text itself is a `prop`. The "whether the text is currently on or off" changes over time, so that should be kept in `state`.
 
 ```SnackPlayer name=State
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
```
lcelso: My suggestion

```suggestion
import React, { useState, useEffect } from 'react';
import { Text, View } from 'react-native';

const Blink = (props) => {
  const [isShowingText, setIsShowingText] = useState(true);

   useEffect(() => {
     const toggle = setInterval(() => {
       setIsShowingText(!isShowingText);
     }, 1000);
     
     return () => clearInterval(toggle);
  })

  if (!isShowingText) {
    return null;
  }

  return <Text>{props.text}</Text>;
}

export default function BlinkApp() {
  return (
    <View style={{marginTop: 50}}>
      <Blink text='I love to blink' />
      <Blink text='Yes blinking is so great' />
      <Blink text='Why did they ever take this out of HTML' />
      <Blink text='Look at me look at me look at me' />
    </View>
  );
}
```

luism3861: i change with your suggestion thank you!!

luism3861: i change your suggestion thank you!!! :)

luism3861:
thanks rachel!!!

PR #1629, opened 2020-02-14
Refactor AccessibilityInfo Example into React Hooks

mdabrowka:
Issue: #1579

- Updates jsx example with React Hooks
- Adds Snack example with React Hooks

Please note: I’ve tested with SnackPlayer and the version of React Native it uses underneath does not include the methods `isReduceMotionEnabled` and `isScreenReaderEnabled` from the AccessibilityInfo API. For that reason, I used `fetch` in the Snack example.

### DIFF
--- a/docs/accessibilityinfo.md
+++ b/docs/accessibilityinfo.md
@@ -76,6 +76,75 @@ const styles = StyleSheet.create({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


mdabrowka:
Apologies for the messed up deploy-preview -I have now fixed the code- could the preview perhaps be re-run?

PR #1633, opened 2020-02-15
Add security documentation

kadikraman:
This PR adds some general security documentation that would be great to know for every React Native engineer.

I considered creating a new top level sidebar item with Security and having the 3 subcategories as sections, however there would then be no place to put the introduction, so I ended up keeping it all on the same page.


kadikraman:
Thanks @mitulsavani - I addressed your comments 🙏

docs/security.md
```diff
@@ -0,0 +1,115 @@
+---
+id: security
+title: Security
+---
+
+Security is an often overlooked aspect when building software. Similarly to how we’ve yet to invent a completely impenetrable lock (bank vaults do, after all, still get broken into), it is also almost impossible to build software to be completely impenetrable. However, the probability of falling victim to a malicious attack or being exposed for a security vulnerability is inversely proportional to the effort you’re willing to put in to protecting your application against any such eventuality. Although an ordinary padlock is pickable, it is still much harder to get past than a cabinet hook!
```
yohix: Fix Grammar

```suggestion
Security is an often overlooked aspect when building software. Similarly to how we’ve yet to invent a completely impenetrable lock (bank vaults do, after all, still get broken into), it is also almost impossible to build software to be completely impenetrable. However, the probability of falling victim to a malicious attack or being exposed to a security vulnerability is inversely proportional to the effort you’re willing to put into protecting your application against any such eventuality. Although an ordinary padlock is pickable, it is still much harder to get past than a cabinet hook!
```

charpeni:
Great job, @kadikraman!

Now that 0.62 is getting closer, do we want to include this in it? @rachelnabors @kadikraman.

rachelnabors:
Yes! The text is good. I want to add more illustrations this week. But I can always backport them to 0.62!

PR #1637, opened 2020-02-17
fix grammar - add 'will'

mr-islam:
Hi,

Just wanted to point out a tiny word that was missing in the docs that messed up the grammar.

The page on [Share](https://facebook.github.io/react-native/docs/share) currently states:

> In Android, Returns a Promise which always be resolved with action being `Share.sharedAction`.

This PR adds the word _will_ to the above sentence, leading to this improved sentence:

> In Android, Returns a Promise which will always be resolved with action being `Share.sharedAction`.

It's a small issue but having perfect docs is always nice :)


rachelnabors:
I agree! Good grammar also makes it easier to use automatic translation services. Thank you for your vigilance!

PR #1641, opened 2020-02-18
Alert documentation and minor fixes

PaigeKH:
I updated the Alert documentation page so its formatting matched the rest of the documentation pages, and fixed a little bit of grammar.


PaigeKH:
Removed the commit with extra changes, this only has changes to the Alert page now.

PR #1642, opened 2020-02-19
update docs for suggest Theme Provider

soroushm:
suggest a good new lib


rachelnabors:
We try not to lead people to individual libraries in the docs and stick strictly to documenting the APIs and Core Components! Thank you!

PR #1645, opened 2020-02-19
use const

CyrusZei:
since we are using const all over the place, but not here


CyrusZei:
ahh, nice :D Thank you and keep up the good work that you guys are doing. 

Have a wonderful day

PR #1649, opened 2020-02-20
rewrite fetch example using hooks

Simek:
Refs #1579, #1272.

Same as with earlier PR, this page is not a part of "Component Docs Drive" but I think it will be a good supplementary improvement to those changes. 

@rachelnabors Question is if class component example should be rewritten and retained (using tabs)  or is it okay to just replace it?


rachelnabors:
Re: class example, ah yes, you know not a bad idea to tuck that under the tabs interface! Sorry, got a bit excited is all!

Simek:
No problem 😉 I'll prepare separate PR with rewritten component class example and tabs later today.

PR #1666, opened 2020-02-25
updated example to use functional components and hooks

tsawan:
Hi, I've added functional component examples with hooks for AppState API. #1579


rachelnabors:
@tsawan Do you know if the props and methods are up to date as well?

tsawan:
@rachelnabors Sorry I have no idea if the props and methods are up to date.

rachelnabors:
It would be a good next issue for you to tackle!

PR #1668, opened 2020-02-25
docs: update typescript template for 0.61

Naturalclar:
Currently, accessing the document for `typescript` https://reactnative.dev/docs/typescript leads to page for React Native v0.60, which specifies the template version to be v6.2.0, for RN 0.60.

To use the typescript-template with the latest version 0.61.5, user would want to use the current latest version of `react-native-typescript-template`.

The document for typescript that does not have a specified template version only exists in the `@next` version of the document, which I believe is more suited for the rc versions?

Once the React Native 0.62 gets released, the template version for 0.61 will need to be specified (I'm assuming it's going to be 6.3.x), but until then, the document for 0.61 should not specify any version for the typescript template.

### DIFF
--- a/website/versioned_docs/version-0.61/typescript.md
+++ b/website/versioned_docs/version-0.61/typescript.md
@@ -1,0 +1,224 @@
+---
+id: version-0.61-typescript
+title: Using TypeScript with React Native
+original_id: typescript
+---
+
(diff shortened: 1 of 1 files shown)


Simek:
@Naturalclar You did not commit changes generated by Docuosaurus while running site:

Screenshot 2020-02-27 at 09 47 51

@rachelnabors Is there any way we can config deploy bot to fail in such situation?

PR #1674, opened 2020-02-27
update Share doc with functional component example

amandeepmittal:
Here is an update for the #1579 regarding the Share component doc.

The class component already exists. I am adding a functional component example too as well as add the support for Snack player.

### DIFF
--- a/docs/share.md
+++ b/docs/share.md
@@ -22,7 +22,7 @@ title: Share
 import React from 'react';
 import { Share, View, Button } from 'react-native';
 
-const ShareExample = () => {
+export default ShareExample = () => {
   const onShare = async () => {
     try {
       const result = await Share.share({
@@ -48,8 +48,6 @@ const ShareExample = () => {
     </View>
   );
 };
-
-export default ShareExample;
 ```
 
 <block class="classical syntax" />
--- a/docs/share.md
+++ b/docs/share.md
@@ -3,6 +3,85 @@ id: share
 title: Share
 ---
 
+## Example
+
+<div class="toggler">
+  <ul role="tablist" class="toggle-syntax">
+    <li id="functional" class="button-functional" aria-selected="false" role="tab" tabindex="0" aria-controls="functionaltab" onclick="displayTabs('syntax', 'functional')">
+      Function Component Example
+    </li>
+    <li id="classical" class="button-classical" aria-selected="false" role="tab" tabindex="0" aria-controls="classicaltab" onclick="displayTabs('syntax', 'classical')">
+      Class Component Example
+    </li>
+  </ul>
+</div>
+
+<block class="functional syntax" />
+
+```SnackPlayer name=Function%20Component%20Example
+import React, { Component } from "react";
+import { Share, Button } from "react-native";
+const ShareExample = () => {
+  onShare = async () => {
+    try {
+      const result = await Share.share({
+        message:
+          "React Native | A framework for building native apps using React"
+      });
+      if (result.action === Share.sharedAction) {
+        if (result.activityType) {
+          // shared with activity type of result.activityType
+        } else {
+          // shared
+        }
+      } else if (result.action === Share.dismissedAction) {
+        // dismissed
+      }
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


docs/share.md
```diff
@@ -3,6 +3,85 @@ id: share
...
+    <li id="classical" class="button-classical" aria-selected="false" role="tab" tabindex="0" aria-controls="classicaltab" onclick="displayTabs('syntax', 'classical')">
+      Class Component Example
+    </li>
+  </ul>
+</div>
+
+<block class="functional syntax" />
+
+```SnackPlayer name=Function%20Component%20Example
+import React, { Component } from "react";
```
Simek: Unnecessary `Component` import.

docs/share.md
```diff
@@ -3,6 +3,85 @@ id: share
...
+import { Share, Button } from "react-native";
+const ShareExample = () => {
```
Simek: I think that `export default ShareExample ...` will be better for the consistency between examples.

amandeepmittal:
Hey @Simek thanks for the review. I made an update to the doc and also (as per Code Tips) am adding `&supportedPlatforms=ios,android` to each Snack Player to disable the support for `web`. Is this okay, let me know, otherwise I will make changes.

amandeepmittal:
Hey @Simek thanks for the review. I made an update to the doc and also (as per Code Tips) am adding `&supportedPlatforms=ios,android` to each Snack Player to disable the support for `web`. Is this okay, let me know, otherwise I will make changes.

Simek:
@amandeepmittal It's the right way, thank you!

I have also left one more comment which should be addressed before merge.

docs/share.md
```diff
@@ -3,6 +3,86 @@ id: share
...
+  </ul>
+</div>
+
+<block class="functional syntax" />
+
+```SnackPlayer name=Function%20Component%20Example&supportedPlatforms=ios,android
+import React from "react";
+import { Share, Button } from "react-native";
+
+export default const ShareExample = () => {
```
Simek: This syntax is invalid and produce error. I have deliberate posted the right way. Please test you examples before committing.

Annotation 2020-02-27 162156

amandeepmittal:
Sorry about that @Simek.

I have changed the example a bit not the core functionality. I have added a `View` with spacing such that someone running the example using Expo Snack on a device with notch will get the desired results. Let me know if this okay?

Also, I have tested both the examples in the Snack player.

docs/share.md
```diff
@@ -3,6 +3,97 @@ id: share
...
+      } else if (result.action === Share.dismissedAction) {
+        // dismissed
+      }
+    } catch (error) {
+      alert(error.message);
+    }
+  };
+  return (
+    <View style={{ marginTop: 50 }}>
+      <Button onPress={() => onShare()} title="Share" />
```
Simek: This part can be simplified to `onPress={onShare}`.

Simek:
@amandeepmittal Other than comment above LGTM, but I would prefer to have `export default` at the top of the code as I mentioned earlier. This matches style of most of new examples, reduce possibility of issues while copying and saves few lines.

docs/share.md
amandeepmittal: Done.

amandeepmittal:
@Simek thank you.

If I use the `export default` at the top in the functional component example, I am getting the following error in the Snack. So either I have to use the keyword `function` or have to eliminate the name of the function and it could be like `export default () => {...}`.

Let me know what would be the optimal preference?

Screenshot 2020-02-28 at 5 31 00 PM

Simek:
@amandeepmittal It's just: `export default ShareExample = () => {`, without `const`, as my first review comment stated.

amandeepmittal:
Thanks @Simek I am sure have taken a lot of your time for this one.

Simek:
@amandeepmittal I hope this process will help you with the future PRs. 

And on more thing, `export default` position in class component needs to be reverted back to the top, as in original example, because now we are again not consistent like at beginning of the review.

PR #1684, opened 2020-02-29
Change width and height.md  example with functional component

luism3861:
@rachelnabors  i change this little example with functional component


rachelnabors:
Thanks for pitching in!

harley_blow-kiss

PR #1685, opened 2020-03-01
🔙 Updated BackHandler Docs with snack examples 🤾🏽‍♂️

DaniAkash:

Have added snack examples for BackHandler and also added references to React Navigation & React Native hooks.

### DIFF
--- a/docs/backhandler.md
+++ b/docs/backhandler.md
@@ -3,19 +3,16 @@ id: backhandler
 title: BackHandler
 ---
 
-Detect hardware button presses for back navigation.
+The Backhandler API detects hardware button presses for back navigation, lets you register event listeners for the system's back action, and lets you control how your application responds. It is Android-only.
 
-Backhandler lets you register event listeners for the system's back action and helps you control how your application should respond.
+The event subscriptions are called in reverse order (i.e. the last registered subscription is called first). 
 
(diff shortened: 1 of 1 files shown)


docs/backhandler.md
```diff
@@ -3,62 +3,166 @@ id: backhandler
...
+      ]);
+      return true;
+    };
+
+    const backHandler = BackHandler.addEventListener(
+      "hardwareBackPress",
+      backAction
+    );
+
+    return () => backHandler.remove();
```
rachelnabors: Question: should this example use `Backhandler.removeEventListener()`?

DaniAkash: `removeEventListener` is a static method that requires passing the reference of the event handler to clear the event listener. 
However, `addEventListener` already returns a `NativeEventSubscription`
Screenshot 2020-03-04 at 4 31 12 AM
which can be cleared by calling the remove method which feels more straightforward 
Screenshot 2020-03-04 at 4 31 24 AM

If we need to explain the usage of `removeEventListener` it should be a separate example

rachelnabors: I feel like that needs covering as well. Ideally the example should include all of the API—or have text explaining where and how to use the bits that aren't covered.

DaniAkash: Seems like I need to come up with an explanation & an example on this 🤔

rachelnabors: Would you please, @DaniAkash? Would really complete the docs!

DaniAkash: Added the explanation!

PR #1686, opened 2020-03-01
TouchableWithoutFeedback: example code in a Snack player

kakukakug:
#1579

TouchableWithoutFeedback：
I changed  example code to use a Snack player.

### DIFF
--- a/docs/touchablewithoutfeedback.md
+++ b/docs/touchablewithoutfeedback.md
@@ -37,15 +37,14 @@ export default function TouchableWithoutFeedbackExample() {
 
   return (
     <View style={styles.container}>
+      <View style={styles.countContainer}>
+        <Text style={styles.countText}>Count: {count}</Text>
+      </View>
       <TouchableWithoutFeedback onPress={onPress}>
         <View style={styles.button}>
           <Text> Touch Here </Text>
         </View>
       </TouchableWithoutFeedback>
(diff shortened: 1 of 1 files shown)


docs/touchablewithoutfeedback.md
```diff
@@ -23,6 +21,56 @@ function MyComponent(props) {
...
+  return (
+    <View style={styles.container}>
+      <TouchableWithoutFeedback onPress={onPress}>
+        <View style={styles.button}>
+          <Text> Touch Here </Text>
+        </View>
+      </TouchableWithoutFeedback>
+
+      <View style={styles.countContainer}>
+        <Text style={styles.countText}>{count !== 0 ? count : null}</Text>
```
Simek: Just a question: What is wrong with rendering `0`?  😃

It will be nice if you can move counter above button and add "Count:" label like in [TouchableOpacity](https://reactnative.dev/docs/next/touchableopacity) Snack example.

kakukakug: Thank you for the review!
That's easier to understand, I'll fix it.

kakukakug: thanks！
Fixed it

kakukakug:
Thanks for the review! Please check and pull it because we fixed it

PR #1687, opened 2020-03-01
add Snack example to PixelRatio page

Simek:
Refs #1579.

This PR adds Snack example to the Pixel Ratio page.  This example is very simple and I have deliberate chosen text representation for the image/resource size calculation, mainly for the clarity what this API can do and which information it provides. But the phrasing might be the first thing to tweak here ; )

Preview:
Screenshot 2020-03-01 at 20 50 10


Simek:
Example devices list has been updated to contain only currently supported Apple devices.

PR #1691, opened 2020-03-03
💃🏽 Animated & PanResponder 👨🏽‍🎤 Snack Examples 👯‍♀️

DaniAkash:

Contains New Function + Class Component examples for the
- Animated API
- PanResponder API
- Animations Guide

Willing to create more examples ﹣ if cool ideas are suggested 😎 💡 🤝

### DIFF
--- a/docs/animated.md
+++ b/docs/animated.md
@@ -9,7 +9,7 @@ The most basic workflow for creating an animation is to create an `Animated.Valu
 
 > The animated value should not be modified directly like Javascript variables hence it is usually kept as a state variable in class components. Functional components with hooks can use `useRef` hook which returns a mutable ref object whose `current` property is initialized as the given argument and persists throughout the component lifecycle.
 
-The following example contains a simple view which will fade in & fade out based on the animated value `fadeAnim`
+The following example contains a `View` which will fade in & fade out based on the animated value `fadeAnim`
 
 <div class="toggler">
   <ul role="tablist" class="toggle-syntax">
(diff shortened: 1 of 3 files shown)


rachelnabors:
@DaniAkash did you check the Animated and PanResponder APIs to see if they need updating as well?

DaniAkash:
> @DaniAkash did you check the Animated and PanResponder APIs to see if they need updating as well?

Looking into the source code. Will update the PR in 2 days

DaniAkash:
@rachelnabors I looked into the `PanResponder` API & discovered something big! there is a ﹣ 
- [`getInteractionHandle`](https://github.com/facebook/react-native/blob/master/Libraries/Interaction/PanResponder.js#L389)
- And [`clearInteractionHandle`](https://github.com/facebook/react-native/blob/31980094107ed37f8de70972dbcc319cc9a26339/Libraries/Interaction/PanResponder.js#L561) method. 

So does that mean pan handlers are like event listeners that should be cleared on unmount?
If that's the case I'll create a new PR addressing it 👍

rachelnabors:
I would imagine so! @DaniAkash can you handle?

DaniAkash:
It's gonna take some time. I'll raise the PR!

DaniAkash:
> I would imagine so! @DaniAkash can you handle?

@rachelnabors the panresponder events are configured to be handled internally so even if the event handler methods are there, they are not exposed to the users. 

Which also means the API is up-to-date and doesn't need another PR

Issue #1573, opened 2020-01-31
Small enhancement to the animated doc


kelset:
Basically, in the Animated doc is not super clear that you have a callback option for `start()`; the only section mentioning [it is here](https://facebook.github.io/react-native/docs/animated.html#working-with-animations) as far as I can tell.

I think it would be super helpful/clarifying to have a clearer code sample. The closest thing to it is in a reply on a GH issue of the main repo:

`this.animateValue.spring({}).start(function onComplete() {})`
[(src)](https://github.com/facebook/react-native/issues/3212)

Which I think is also outdated, because according to the link above in the docs should be something like

`this.animateValue.spring({}).start((finished) => {})`

(at least, I think - ideally it would be better to go and check source code)

---

BTW I only found out about this via a [StackOverflow issue](https://stackoverflow.com/questions/38053071/react-native-animated-complete-event) so yeah it seems that it's quite relevant.

I'll assign to myself and try to tackle asap, just can't now but don't want to lose track of this.

(also, it's probably a really good first issue for someone trying to do their first contribution so I'll add that tag too)

gedeagas:
hi @kelset , are you still planing to tackle this? if not, I would love to help.

kelset:
Hey @gedeagas - no I haven't had time yet, feel free to do a PR for it if you want :)

rachelnabors:
It would be great if `start` was listed under Animated's methods, too! Please tag this with #1579 to get credit for completing the API docs, @gedeagas!

gedeagas:
@kelset Awesome, i will be taking this then.

@rachelnabors thank you for the information, will do.

Simek:
The issue reported here has been fixed by the merge of PR created by @gedeagas.

PR #1702, opened 2020-03-06
WIP: Update Animated.md to add proper docs about start(), stop(), and reset()

gedeagas:
#1579 

According to #1573 the Animated doc is not super clear that you have a callback option for start(). 
So the purpose of this PR is to add the proper docs and code example about callback option for start().

- [x] `start()`
- [x] `stop()`
- [x]  `reset()`

### DIFF
--- a/docs/animated.md
+++ b/docs/animated.md
@@ -556,6 +556,26 @@ Animated.timing({}).start(({finished}) => {
 });
 ```
 
+---
+
+### `stop()`
+
+```jsx
+static stop()
+```
+
+Stops any running animation.
+
+---
+
+### `reset()`
+
+```jsx
+static reset()
+```
+
+Stops any running animation and resets the value to its original.
+
 ## Properties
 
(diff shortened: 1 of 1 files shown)


gedeagas:
Hi @rachelnabors, i still quite unsure where to put `start()` in the list of methods. Also because we are adding start on to the list of methods isn't it better to add `stop()` and `reset()` also?. Thanks.

rachelnabors:
Hi, @gedeagas! Thanks so much for taking this on! And you're right: `stop()`, `start()`, and `reset()` should all be in the methods section (alphabetically listed, naturally). It would be great if you could check the source code to see if there are any other props, methods, or events not listed. Seems like a big deal, right?

gedeagas:
Will do, i will double check the source code again if there are any other props, methods, or events not listed on `stop()`, `start()`, and reset(). Thank you @rachelnabors.

PR #1714, opened 2020-03-10
update 2 examples usinglistview with functional components

luism3861:
@rachelnabors  i update this examples with functional components :)!!


Simek:
Could you also use arrow function since you touched this examples?

luism3861:
> Could you also use arrow function since you touched this examples?

thanks i change your suggestion :)!!

luism3861:
> Please use arrow functions as per @Simek!

i change with arrow functions @rachelnabors  :))

PR #1718, opened 2020-03-10
update Transforms page and simplify examples

Simek:
This PR updates the `Transforms` API page. The following changes has been made:
* examples has been simplified and adjusted to fit better narrow Snack player
* intro text has been adjusted (also example reference has been removed)
* code block has been added to the `transform()` examples
* type table has been adjusted for the readability

I have also removed the type table below deprecated methods because it content was confusing and invalid (deprecated methods were accepting `number` or `string` depending on performed transformation).

### DIFF
--- a/docs/transforms.md
+++ b/docs/transforms.md
@@ -3,7 +3,7 @@ id: transforms
 title: Transforms
 ---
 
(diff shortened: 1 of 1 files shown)


rachelnabors:
So one of the reasons I inlined those styles was to put them close to the element people wanted to experiment with. Originally they were all in the separate CSS file, but that seemed a bit to abstracted to me. Of course, the resulting code is more messy—but it keeps the transforms on the Element they transform! What do you think? Is there a better way?

Simek:
I think there is no better way, you can specify it in the `StyleSheet` or inline in component styles. 

Only different formatting could help with readability in this example, so I have manual reformatted the code. Only bad news is that after running code thought Prettier with default settings it will always format code back to the previous state.

If you think that it's not a good solution I can revert the most part of the examples and leave only docs changes.

PR #1719, opened 2020-03-10
Clarify what a component and element are

rachelnabors:
We weren't being explicit about this. Studying more React on my own and with others, I think it makes sense to add a little clarification that `Cat` is the component and `<Text>` is the React element—that represents Views. Would've loved to put this in the prior section on Native Components and Core Components, but honestly, we really needed to explain that RN isn't a compiler and what Views are in relation to Native Components. Hopefully this little bit helps drive this home even more :)

### DIFF
--- a/docs/intro-react.md
+++ b/docs/intro-react.md
@@ -56,15 +56,15 @@ Your component starts as a function:
 function Cat() {}
 ```
 
-Whatever a function component returns is rendered as a **React element.** `Cat` will render a `<Text>` element:
+You can think of components as blueprints. Whatever a function component returns is rendered as a **React element.** React elements let you describe what you want to see on the screen.
+
+Here the `Cat` component will render a `<Text>` element:
 
 ```jsx
 function Cat() {
   return <Text>Hello, I am your cat!</Text>;
 }
 ```
-
-Here, `Cat` is the component, a blueprint. What it returns, `<Text>`, is a React element, an object that describes the Views that the compnent represents.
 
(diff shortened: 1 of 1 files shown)


docs/intro-react.md
```diff
@@ -56,14 +56,16 @@ Your component starts as a function:
...
-Whatever a function component returns is rendered as a React element. `Cat` will render a `<Text>` element:
+Whatever a function component returns is rendered as a **React element.** `Cat` will render a `<Text>` element:
 
 ```jsx
 function Cat() {
   return <Text>Hello, I am your cat!</Text>;
 }
 ```
 
+Here, `Cat` is the component, a blueprint. What it returns, `<Text>`, is a React element, an object that describes the Views that the compnent represents.
```
gaearon: >Here, `Cat` is the component, a blueprint.

I like this.

>What it returns, `<Text>`, is a React element, an object that describes the Views that the compnent represents.

I think this sounds a bit too abstract — there's multiple levels of indirection ("element, an object that describes views that component represents"). If you don't know these terms already I think it's too dense.

Here's how I would word it:

>Here, `Cat` is the component, a blueprint. What it returns, `<Text>`, is a React element. React elements let you describe what you want to see on the screen.

rachelnabors: Hmmm... I want to reinforce what was said in the previous section: that Native Components are backed by Views. Where we might say on the Web

> a React element describes the DOM nodes the component represents 

in React Native mightn't we say 

> a React element describes the Views the component represents

pedantic-raptor

Issue #1331, opened 2019-10-09
Missing typescript type definition for StyleSheet.compose()


FDiskas:
https://facebook.github.io/react-native/docs/stylesheet#compose

PR #1723, opened 2020-03-10
add basic types to the StyleSheet methods

Simek:
Refs #1331 

This PR adds basic, missing types to the `SteelSheet.compose()` and `SteelSheet.flatten()` methods.

I have also put information about experimental nature of `setStyleAttributePreprocessor()` in a note box.

Source: [facebook/react-native/Libraries/StyleSheet/StyleSheet.js](https://github.com/facebook/react-native/blob/4e71a30969d74073309d0350be55cadb84ae43ff/Libraries/StyleSheet/StyleSheet.js)


Simek:
This PR received another update with following changes, which seems needed:
* restore `create()` method reference deleted during docs refactor
* sort method and properties alphabetically
* remove unused code, simplify examples and run them through Prettier
* update examples UI to be more coherent with  the website

@rachelnabors I also think that `StylesSheet` page should be moved from the `Core Components` to the `APIs` section of the sidebar. What do you think about that change?

Simek:
@rachelnabors What about sidebar (last paragraph from my last comment), should we change the position?

rachelnabors:
That makes sense to me

Simek:
@rachelnabors StyleSheet moved : )

Issue #1724, opened 2020-03-10
React Native Showcase Link Broken


sazzadsazib:
Hello, I was going through react native docs, in this [Showcase](https://reactnative.dev/docs/more-resources#examples) section it has a broken link !

Naturalclar:
@sazzadsazib nice catch! the correct link seems to be `https://reactnative.dev/showcase/`
Would you like to send a PR to fix this?

sazzadsazib:
Alright !

PR #1727, opened 2020-03-11
Update showcase link of More Resource

sazzadsazib:
Updated the link **showcase** in **More Resource** section. 

fix for #1724


sazzadsazib:
> mlp_party
> Thanks for spotting this!

Always welcome 😸

PR #1737, opened 2020-03-12
docs: add information about toolbar android

Naturalclar:
Add information about the community ToolbarAndroid package https://github.com/react-native-community/toolbar-android


rachelnabors:
Did you mean to add this to a versioned do?

Naturalclar:
@rachelnabors yes, the addition to the versioned doc is intentional.
unlike the other modules that are deprecated (but still exists in the core), ToolbarAndroid no longer exists in the react-native repo, and the document no longer exist in the `next` version. :+1:
ref: https://github.com/facebook/react-native/issues/26591#issuecomment-570972727

PR #1748, opened 2020-03-15
Fixing Mardown syntax

cHaLkdusT:
# Fix
Fixed **docs/removing-default-permissions.md** file since contains invalid Markdown syntax.

# Preview
Screen Shot 2020-03-15 at 21 24 46

## NOTE
Ran `$ yarn prettier` before committing that's why other files were updated too.

### DIFF
--- a/docs/datepickerandroid.md
+++ b/docs/datepickerandroid.md
@@ -3,7 +3,7 @@ id: datepickerandroid
 title: 🚧 DatePickerAndroid
 ---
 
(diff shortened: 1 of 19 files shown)


cHaLkdusT:
Hi @rachelnabors, I have now reverted the other files so that this pull request includes only related file. 

Is it okay if I create another PR for the other files so that it is consistent with Prettier config?

PR #1751, opened 2020-03-15
Linking: Add Snack examples, function components and hooks

danilobrinu:
In Reference to #1579

This PR adds Snack examples, function components and hooks for the Linking API.


danilobrinu:
@rachelnabors can you review this PR ?

Also I get that error from the Snack example, so I don't know why ? Can you check that ?

rachelnabors:
Thanks so much for working on this @danilobrinu! Before I can accept, a few things:

1) Can you confirm that you have checked the API is up to date in this PR?
2) What error in what snack example? Which platform?

danilobrinu:
@rachelnabors 

1. Yes, the API is updated, I checked the source code before write the examples. 

2. You can try the Snack example for send intents.

danilobrinu:
@rachelnabors thanks to you for your amazing work. Big hugs.

PR #1752, opened 2020-03-16
🎨 Format SnackPlayer code snippets

cHaLkdusT:
Make all SnackPlayer code snippets consistent with our Prettier config

### DIFF
--- a/docs/accessibilityinfo.md
+++ b/docs/accessibilityinfo.md
@@ -21,9 +21,8 @@ Sometimes it's useful to know whether or not the device has a screen reader that
 <block class="functional syntax" />
 
 ```SnackPlayer name=AccessibilityInfo%20Function%20Component%20Example
-
-import React, { useState, useEffect } from "react";
-import { AccessibilityInfo, View, Text, StyleSheet } from "react-native";
+import React, {useState, useEffect} from 'react';
+import {AccessibilityInfo, View, Text, StyleSheet} from 'react-native';
 
 export default function App() {
   const [reduceMotionEnabled, setReduceMotionEnabled] = useState(false);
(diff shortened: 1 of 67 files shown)


rachelnabors:
Could you see about updating the config so imports are not multi-lined? I hate seeing a third of the Snack window consumed by importing View, Input, Text, etc..

Simek:
IMO this PR should be in hold until Component Docs Drive (#1579) is finished.

cHaLkdusT:
Hi @rachelnabors, I've checked Prettier config, unfortunately, there's no option to make imports not multi-lined. There's an on-going discussion here: https://github.com/prettier/prettier/issues/5995

Would you like revert files with multi-lined imports instead?

docs/direct-manipulation.md
```diff
@@ -62,23 +62,23 @@ Composite components are not backed by a native view, so you cannot call `setNat
...
+};
 
-export default App = () => {
-    return (
-      <TouchableOpacity>
-        <MyButton label="Press me!" />
-      </TouchableOpacity>
-    )
-}
+export default (App = () => {
```
Simek: Curly braces after `export default` are unnecessary (and a bit confusing). It will be nice to omit them if config allows to.

docs/accessibilityinfo.md
```diff
@@ -21,58 +21,57 @@ Sometimes it's useful to know whether or not the device has a screen reader that
 <block class="functional syntax" />
 
 ```SnackPlayer name=AccessibilityInfo%20Function%20Component%20Example
-
-import React, { useState, useEffect } from "react";
-import { AccessibilityInfo, View, Text, StyleSheet } from "react-native";
+import React, {useState, useEffect} from 'react';
```
Simek: So long as I'm think of this I'm not sure if disabling [bracket spacing](https://prettier.io/docs/en/options.html#bracket-spacing) is a good change. I think that having those spaces improves overall readability of the code (especially when function uses object as a parameter).

docs/direct-manipulation.md
rachelnabors: I'm not seeing where we can add that: https://prettier.io/docs/en/options.html

Simek: It's look like a ESLint/Prettier issue with this particular construction. Expo Snack editor also adds those brackets on autoformat. I'm not sure why though since `export` is an expression/directive not a function.

rachelnabors:
Sorry to close this... I updated Prettier to be a little more readable based on the conversation here and am landing redundant updates. But THANK YOU for bringing this conversation to light! Sometimes discussions themselves are worth the PR :)

PR #1759, opened 2020-03-18
update flexbox examples with functional components

luism3861:
@rachelnabors  i change this little example with functional components with arrow functions!!


luism3861:
> Remove all `margin: 20` styles because these can cause confusion with Flexbox topic. See for example the `justifyContent: 'space-between'`
> Screen Shot 2020-03-18 at 19 58 51
> 
> See: **Write for more than a React audience** at STYLEGUIDE.md

thanks i remove margintop styles in all examples thank you @cHaLkdusT

PR #1765, opened 2020-03-18
Add instructions if something goes bad with installing cocoapods

crramirez:

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -220,6 +220,12 @@ Using the default Ruby install will require you to use `sudo` when installing ge
 sudo gem install cocoapods
 ```
 
+If you face errors about a header file not found try to install this package:
+
+/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg
+
+And try again
+
 For more information, please visit [CocoaPods Getting Started guide](https://guides.cocoapods.org/using/getting-started.html).
 
 <block class="native linux android" />


Simek:
I think that this tip need to provide a bit more context for the users. 

From the quick web skimm I have found that this issue affects only macOS `10.15` systems after XCode 10 release ([mention in release notes](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes#3035624)) due to CLI changes.

crramirez:
I have macOS 10.14 Mojave and happened. Here is the link to the track where I got the information: https://github.com/castwide/vscode-solargraph/issues/78#issuecomment-535210783

Simek:
@crramirez Yup, there was a typo in my comment about version, but I have in mind that this seems like specific OS version issue which occurs only with a certain range of XCode version. 

Can you add some more information and clarify who and in which circumstances should try this fix? The description should help people less familiar with the issue. You can also add the link that you mentioned here or any other useful link for the reader.

crramirez:
OK good, I'll work on it.

Simek:
After reading comments in the issue you have linked here it seems the solution heavily depends on the macOS and XCode version combination.
* for 10.14 and XCode 11 seems that https://github.com/castwide/vscode-solargraph/issues/78#issuecomment-552675511 helped
* for 10.15 and XCode 11 seems that only ```export SDKROOT=$(xcrun --show-sdk-path)``` is needed

It could be hard to not confuse the reader in this situation.

Simek:
It will be nice if you could update [CocoaPods Troubleshooting](https://guides.cocoapods.org/using/troubleshooting#installing-cocoapods) page  ([CocoaPods/guides.cocoapods.org/source/using/troubleshooting.html.md](https://github.com/CocoaPods/guides.cocoapods.org/blob/eecd642bf487f45d3db1d64dd9cc86ae1267b370/source/using/troubleshooting.html.md)), list all the cases there and then we can just use Troubleshooting link on the React Native website.

rachelnabors:
@crramirez Let us know if you can get back to this! I'll close it out Friday if not. Thank you!

PR #1771, opened 2020-03-20
Prettier config update

rachelnabors:
In #1752 we discussed what an ideal Prettier config would look like: more space, no trailing commas. Applied everywhere


docs/easing.md
```diff
@@ -359,7 +359,7 @@ A useful tool to visualize cubic bezier curves can be found at http://cubic-bezi
 ### `in()`
 
 ```jsx
-static in(easing);
+static in easing;
```
Simek: This is a Prettier issue, can we [ignore](https://prettier.io/docs/en/ignore.html#markdown) this line?

lcelso:
I loved it.

Simek:
@cHaLkdusT Prettier rules are extension not context based. This means that until JS code is inside `md` file all rules will apply to this code. 

`*.js` rule only apply to JS files such as `core/Footer.js`.

TheSavior:
What is the reason for the website code snippets having a different prettier config than the core codebase? I could understand some things like additional spacing making things easier to read, but I’m not sure about the trailing comma change

docs/easing.md
rachelnabors: This seems like a pretty big deal. How many other places is it doing something/going to do something similar?

Simek: I was bit surprised but it seems that only this line is affected. 

I assume that the reason is that `in` in JS is a `keyword` and you usually cannot use it as a name for stand-alone variables or functions. Here it's fine because it's `static` method.

rachelnabors:
I'm going to close this and retry after the dust of 0.62 settles.

PR #1776, opened 2020-03-23
Update to pixelratio.md - getFontScale() returns value on iOS

devben:
iOS does return a user value using getFontScale

### DIFF
--- a/docs/pixelratio.md
+++ b/docs/pixelratio.md
@@ -134,9 +134,9 @@ static getFontScale(): number
 Returns the scaling factor for font sizes. This is the ratio that is used to calculate the absolute font size, so any elements that heavily depend on that should use this to do calculations.
 
 * on Android value reflects the user preference set in **Settings > Display > Font size** 
-* on iOS value reflects the user preference set in **Settings > Accessibilty > Display & Test Size > Larger Text**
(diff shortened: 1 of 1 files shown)


Simek:
@devben I have checked and indeed `getFontScale()` seems to working properly on iOS ([despite in code documentation](https://github.com/facebook/react-native/blob/master/Libraries/Utilities/PixelRatio.js#L95)). 

But it returns values based on `"Text Size"` in `"Display & Brightness"` menu. 

Checking `"Large Accessibility Sizes"` setting in `"Accessibility"` -> `"Display & Text Size"` menu only extends the maximum font size limit for the `"Text Size"` setting.

It will be nice if you could also update description in the mentioned above code.

devben:
@Simek sure I can update. Should we perhaps include both?

 **Accessibility>Display & Text Size>Larger Text** section does also show the slider with the "Large Accessibility Sizes" toggle simply increasing the range, this or using the **Accessibility Inspector** are the only methods available to developers testing on the iOS simulator. 

IMG_17BFE5D91D6E-1
https://support.apple.com/en-us/HT202828

Simek:
@devben Sure, I think that mentioning both menus is the best solution. 

Note that after setting `"Large Accessibility Sizes"` in `"Accessibility"` slider in the `"Display & Brightness" -> "Text Size"` also increase it's range. It's look like same view is used for both Settings entries but with additional Switch in on the `"Accessibility"` page.

PR #1789, opened 2020-03-27
adds og:image to every page

threepointone:
This PR adds an `og:image` meta tag to every page, and reuses react's `og:image` for the same. This means links shared on twitter will correctly have a react logo as an attached image.

I can't verify that this actually works till the PR lands, but the generated html looked fine to me. 
image


rachelnabors:
Not a bad idea! I'd like it if individual posts could specify their own og images—like the Security and Getting Started guides have some illustrations that would rock. But that's a Docusaurus level change, I think~

threepointone:
Yeah I dunno how to do it per page, eg- in the react docs we’d override react helmet something or the other. In any case it would be a future PR, this should be good for now.

threepointone:
So it looks like older pages may still be cached with twitter, and may or may not be refresh in the near future. Some rarer pages already show the image correctly tho. eg: In this image - 
image
the first is a link to https://reactnative.dev/ and you can see there *isn't* an image, and the second to https://reactnative.dev/docs/platform-specific-code, which has the correct image. I'm assuming the older ones will fix themselves as the caches flush.

charpeni:

Building with commit 3d87506b4163372a57069d8e77c9e9f6eeecf072

https://app.netlify.com/sites/react-native/deploys/5e7decfcf6a91e0008195b64

PR #1801, opened 2020-03-30
missing Animated.Value.setValue method

kaiwen:
missing `Animated.Value.setValue` method, newbie don't know how to reset the value, they may try `v = new Animate.Value(0)` in `start`'s complete handler, but it does not work.

### DIFF
--- a/docs/animated.md
+++ b/docs/animated.md
@@ -580,7 +580,7 @@ Stops any running animation and resets the value to its original.
 
 ### `Value`
 
-Standard value class for driving animations. Typically initialized with `new Animated.Value(0);`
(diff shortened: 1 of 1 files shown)


Simek:
What do you think about https://reactnative.dev/docs/animatedvalue ? Is this page up to date and helpful? If yes, we can link that page in text and even add it to the sidebar.

kaiwen:
> What do you think about https://reactnative.dev/docs/animatedvalue ? Is this page up to date and helpful? If yes, we can link that page in text and even add it to the sidebar.

Oh, It's wonderful! It's what we want, thank you very much!

rachelnabors:
Thanks so much, both of you! @simek, your PR #1807 makes this PR redundant. What a great example of problem solving together!

PR #1806, opened 2020-03-31
Update getting-started.md

ffpetrovic:


rachelnabors:
We're specifically not using "native" to describe applications anymore. There are iOS apps, Android apps, Windows apps, and so forth—but none of their developers call themselves or what they work on "native."

Issue #426, opened 2018-06-26
Missing documentation for Animated.ValueXY


wh1t3cAt1k:
Encountered today, but https://github.com/facebook/react-native/issues/16967 was closed due to being stale.

https://facebook.github.io/react-native/docs/animated#valuexy does not describe the signature of the `ValueXY` constructor.

For example, it is not clear what is the difference between creating a `ValueXY` with an explicit set of coordinates vs. no coordinates at all.

Simek:
Hi @wh1t3cAt1k, thank you for pointing that! Unfortunately Animated value pages has been removed from the sidebar by the mistake. Page was you was looking for was available under the https://reactnative.dev/docs/animatedvaluexy URL but not from the menu on the left. I have updated the [`next`](https://reactnative.dev/docs/next/animatedvaluexy) version and added the pages back to the sidebar. This docs will become the current one with `0.63` release.

PR #1807, opened 2020-03-31
update animated values pages, add them to the sidebar

Simek:
Refs #1801, #426

For unknown for me reason `Animated.Value` and `Animated.ValueXY` pages has been lost without a sidebar entries for the few releases. After a quick glance at the code those APIs docs seems up to date.

This PR updates both pages to roughly match current standards, `Animated.ValueXY` class component example has been replaced with functional example in Snack, both of them has been added to the sidebar and a mentions has been added to the `Animated` page.

I can also port those changes to the `0.62` branch if you think it would be helpful.

### Preview
Annotation 2020-03-31 190840

### DIFF
--- a/docs/animated.md
+++ b/docs/animated.md
@@ -582,11 +582,15 @@ Stops any running animation and resets the value to its original.
 
 Standard value class for driving animations. Typically initialized with `new Animated.Value(0);`
 
+You can read more about `Animated.Value` API on the separate [page](animatedvalue).
+
 ---
 
 ### `ValueXY`
 
 2D value class for driving 2D animations, such as pan gestures.
+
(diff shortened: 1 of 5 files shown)


rachelnabors:
/Hey, @simek! Wow, this is a pretty big oversight! I didn't know we had these pages! Makes me wonder why we don't reference them anywhere?

I'm wondering, this would be the first time that we'd have a "child" listed in the sidebar (Value and ValueXY being properties of Animated). I wonder if they should live directly on Animated's page? Otherwise, this sets a new precedent/is out of line with how the other pages are arranged.

What do you think?

charpeni:

Building with commit 756a586364480947ab821716f05559045d9405ec

https://app.netlify.com/sites/react-native/deploys/5e835ea98fa77800082eef61

PR #1809, opened 2020-04-01
Fix outdated links in `getting-started.md` page

jyblue:
Motivation:
* It seems that some of external links to `Watchman` are outdated.

Modification:
* Fix broken links

Result:
* The page is updated with the correct links.

PR #1811, opened 2020-04-02
Update the Watchman installation guide link

sterjoski:


Simek:
It looks like this fix is also a part of more broad `Getting Started` links fixing PR: #1809

rachelnabors:
Thanks for submitting! We are fixed in #1809.

PR #1813, opened 2020-04-02
PlatformColor API

kikisaints:
This PR adds the api documentation for the [PlatformColor (https://github.com/facebook/react-native/pull/27908) function.


website/sidebars.json
```diff
@@ -127,7 +127,7 @@
       {
         "type": "subcategory",
         "label": "Hooks",
-        "ids": ["usecolorscheme", "usewindowdimensions"]
+        "ids": ["platformcolor", "usecolorscheme", "usewindowdimensions"]
```
Simek, line 130: What's the reason behind listing `PlatformColor` in `Hooks` section not `APIs`?

kikisaints: I was under the impression that's where it belongs because it's a function and operates similarly to usecolorscheme - if that's incorrect however, where do you think it should go?

Simek: I was not digging into this too deep, but if what you said is correct and code structure of `PlatformColor` is basically a hook then it violates the [Rules of Hooks](https://reactjs.org/docs/hooks-custom.html) because all hooks must start with ["use"](https://reactjs.org/docs/hooks-faq.html#what-exactly-do-the-lint-rules-enforce). If it's true this should be addressed in RN repo.

If it's not a hook, page should be listed in `APIs` section as I wrote earlier.

kikisaints: Hmm, @tom-un - should this have been "usePlatformColor"? If not I'll move it out of hooks.

Simek: Alsto it seems that despite `PlatformColor` is in code base no entry in [CHANGELOG](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md) mention about releasing it. Are you sure it was included in `0.62` (I'm asking because you also have updated `0.62` docs)?

If it is a part of `0.62` release changelog should be updated.

acoates-ms: @kikisaints, its an API not a hook.

docs/platformcolor.md
```diff
@@ -0,0 +1,48 @@
...
+    ...Platform.select({
+      ios: {color: PlatformColor('labelColor')},
+      android: {color: PlatformColor('?attr/colorControlNormal')},
+      default: {color: 'black'},
+    }),
+  },
+});
+```
+
+The string value provided to the `PlatformColor` function must match and agree with the same string as it exists on the native platform the app is being run on. This means to avoid runtime errors the function should be wrapped in a platform check, either through a `Platform.OS === ‘platform’` or a `Platform.Select()`.
```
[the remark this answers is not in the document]
kikisaints: Done! thanks for catching this.

website/sidebars.json
kikisaints: Thanks for checking. I moved it out of the versioned docs and out of the hooks section.

TheSavior:
Are there going to be separate pages for DynamicColorIOS and ColorAndroid? I’m curious if these should all be separate and/or if there should be a page for Colors in general (named, rgb, hex, and these new functions)

kikisaints:
@TheSavior I don't know much about DynamicColorIOS or ColorAndroid, but I know for Windows we have a separate section on "Colors". I find it's nice to have that separate for easy finding+readability.

TheSavior:
DynamicColorIOS and ColorAndroid are all part of the generic "PlatformColor" api that @tom-un added that is new in 0.63, so we should make sure we have documentation for those. 

Ah, it looks like we have the same colors page: https://reactnative.dev/docs/colors

I think it would make sense to include these new types of colors on that page.

rachelnabors:
Yes, adding them to the colors pages makes sense. @kikisaints would you mind? Good idea, @TheSavior!

docs/platformcolor.md
```diff
@@ -0,0 +1,48 @@
+---
+id: platformcolor
+title: PlatformColor
+---
+
+```
+PlatformColor(string)
```
tom-un, line 7: I see in other API references that Flow notation is being used.   The full Flow notation for `PlatformColor` is: PlatformColor(...names: Array<string>): [Object object]

The PlatformColor function takes an array of strings.   The array must have at least one value.   At runtime, if the first string value does not resolve to a usable color, then the next string value is used and so on.   If none of the values resolve to a color then a RedBox error is raised.   This is useful if an app wishes to use a new platform color that is available in later OS releases but not available in earlier OS releases: the latest platform color name is specified first in the array with fallback color names specified after.

tom-un:
> DynamicColorIOS and ColorAndroid are all part of the generic "PlatformColor" api that @tom-un added that is new in 0.63, so we should make sure we have documentation for those.
> 
> Ah, it looks like we have the same colors page: https://reactnative.dev/docs/colors
> 
> I think it would make sense to include these new types of colors on that page.

Agree we should have a `DynamicColorIOS` reference page as well.   

Regarding `ColorAndroid`: it was added more as a validation tool than as something useful to a developer.   When making the original [PR](https://github.com/facebook/react-native/pull/27908) we felt it was valuable and useful to have working platform specific methods for the two platforms in core to test that the pattern worked in app code (PlatformColorExample.js in RNTester) and that the Flow validation worked, etc.    Practically `PlatformColor()` is more useful to a developer on Android than `ColorAndroid()`.    Now that the construct has served its purpose, I propose I make another PR to remove `ColorAndroid`.  @TheSavior : what do you think?

TheSavior:
Ah, I didn't realize that distinction. PlatformColor handles all the cases that ColorAndroid does? In that case, yeah, removing ColorAndroid is reasonable.

tom-un:
> Ah, I didn't realize that distinction. PlatformColor handles all the cases that ColorAndroid does? In that case, yeah, removing ColorAndroid is reasonable.

Ya, on the native side its identical.   I'll make a PR to remove it.   Thanks!

docs/platformcolor.md
tom-un: As I re-read this, I think it makes more sense to say in English: 'The PlatformColor function takes one or more string value arguments'.   The way I wrote it above implies one passes an array literal, but one does not.   The Flow notation is somewhat confusing.   Also, it looks like I didn't properly markdown escape things.   It should be: `PlatformColor(...names: Array<string>): [Object object]`.

kikisaints:
@TheSavior and @rachelnabors just to be sure I'm understanding what's needed here, you would like me to add to this PR:

- PlatformColor as it is under the API's section
- DynamicColorIOS under in the colors page under the Docs section

is this correct? Want to make sure I'm not leaving out anything here.

TheSavior:
I'm not the docs expert, so I defer to @rachelnabors, but this is my thinking:

I think both of the APIs should live in the Colors page as these APIs only exist as a way to specify colors. If we do want standalone API pages *as well*, then I could see them just pointing to the appropriate section of the colors page.

rachelnabors:
@kikisaints you have it correctly!

@TheSavior The Colors page is an odd case: it's essentially a reference chart. As such, it makes sense for it to mention the API pages and for the API pages to reference it. It keeps those pages nice and parseable by humans. The problem with the page is that there aren't any others like it in the docs aside from maybe the "shared props" pages. 

If the docs were auto-generated, the color APIs would still have their own pages. So the way they are is fine. But perhaps there is a more appropriate place for the color reference page. And perhaps the Docs section needs a real page on the ways to handle colors!

TheSavior:
I think the colors page would make more sense if it was linked to throughout all prop definitions for Color: image

I agree that the APIs would show up if we were generating everything from code, so having standalone pages makes sense too

rachelnabors:
Great idea, @TheSavior, and these [are linked to the color pages](https://reactnative.dev/docs/text-style-props#color) already, but the styling for some reason doesn't indicate a prop is a link :)

kikisaints:
> Great idea, @TheSavior, and these [are linked to the color pages](https://reactnative.dev/docs/text-style-props#color) already, but the styling for some reason doesn't indicate a prop is a link :)

I was able to away with link indication on our RNW site by doing the following within the tables:
`<ins>[color](https://reactnative.dev/docs/color)</ins>`

just adding html `<ins/>` tags.

Simek:
> I was able to away with link indication on our RNW site by doing the following within the tables:
> `<ins>[color](https://reactnative.dev/docs/color)</ins>`
> 
> just adding html `<ins/>` tags.

This can be fixed also using CSS, I can drop the PR for that, but as far as I know, sooner than later reference structure and appearance will be completely redesigned.

docs/colors.md
```diff
@@ -31,6 +31,18 @@ React Native supports also colors as an `int` values (in RGB color mode):
 
 > **_Note:_** This might appear similar to the Android [Color](https://developer.android.com/reference/android/graphics/Color) ints representation but on Android values are stored in SRGB color mode (0xaarrggbb).
 
+## DynamicColorIOS
+
+```jsx
+static DynamicColorIOS({light: [Object object], dark: [Object object]}): [Object object]
+```
+
+The `DynamicColorIOS` function is a platform color type specific to iOS. `DynamicColorIOS` takes a single argument as an object with two keys: light and dark. The values of light and dark themselves must be values returned by the `PlatformColor` function.
```
TheSavior, line 40: I'm not sure it is true that you have to use `PlatformColor`: https://github.com/facebook/react-native/blob/master/RNTester/js/examples/PlatformColor/PlatformColorExample.js#L239

I think you can specify any color type listed in this file in the `light` and `dark` key: hex, named, rgb, `PlatformColor`, etc. You could probably pass a `DynamicColorIOS` even, but I have no idea what that would do. 😅

docs/platformcolor.md
TheSavior: I don't think it is useful to show the return type of the function. It should be completely opaque, all that matters is that it is a value that can be used as a color, anywhere a color can be used.

For reference, [`func.call`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) uses this syntax: `func.call([thisArg[, arg1, arg2, ...argN]])` which for us would be `PlatformColor(color1, [color2, ...colorN])`

docs/colors.md
```diff
@@ -31,6 +31,18 @@ React Native supports also colors as an `int` values (in RGB color mode):
 
 > **_Note:_** This might appear similar to the Android [Color](https://developer.android.com/reference/android/graphics/Color) ints representation but on Android values are stored in SRGB color mode (0xaarrggbb).
 
+## DynamicColorIOS
```
TheSavior, line 34: This could probably use some examples similar to what you have for PlatformColor

website/i18n/en.json
```diff
@@ -257,6 +257,9 @@
       "platform-specific-code": {
         "title": "Platform Specific Code"
       },
+      "platformcolor": {
```
TheSavior, line 260: I think we should be consistent between PlatformColor and DynamicColorIOS. They are both public APIs from React Native, and they can both be used everywhere any other color can. So if one has a standalone page, they should both have a standalone page. If one is in the colors.md file, they both should be.

charpeni:

Building with commit 24aeaf109fd1d41f334d0845c7e6326ca564b08b

https://app.netlify.com/sites/react-native/deploys/5e8e0675723ba400063f4ca6

docs/colors.md
tom-un: You're right: they can be any of the traditional string values or an object returned by PlatformColor.   Sorry Kiki, I gave you wrong information offline.    You *cannot* pass a DynamicColorIOS object as the light or dark value to another DynamicColorIOS -- that is enforced at runtime and by the ESLint rule.   Maybe the language can just end after "... with two keys: light and dark."   If we want to be pedantic we could say they values can be a string per https://reactnative.dev/docs/colors or values returned by PlatformColor.

jfrolich:
Are there any plans to have a `DynamicColor` API, so that there is a single API for dynamic colors that also works on android?

docs/colors.md
kikisaints: removed the comment about platformcolor being the only thing it can take - and removed the static before the dynamiccolorios code block at the top.

docs/colors.md
kikisaints: I'll sync with tom to try and get an example piece in.

website/i18n/en.json
kikisaints: I'm a bit confused. I got the impression Rachel wanted dynamiccolorios in the colors.md and the platformcolor separate.

@rachelnabors - Should I pull out the dynamiccolorios and put it in APIs as a separate section, or combine both into colors.md?

tom-un:
> Are there any plans to have a `DynamicColor` API, so that there is a single API for dynamic colors that also works on android?

No, there are no plans for a platform agnostic DynamicColor API.    

The intention of PlatformColor() is that platform specific strings are passed to the function and it returns an opaque type that is handled by react-native in ways specific to each platform.   On iOS, the string argument is an iOS system color name and under the covers the result is a UIColor object.    On Android, PlatformColor() string arguments are resource paths and under the covers the results are RGB colors as Integers.   Platform specific functions like DynamicColorIOS are another way to create platform specific colors where a single string argument is insufficient.   Like PlatformColor on iOS, DynamicColorIOS ends up creating a UIColor under the covers.    On iOS, UIColor have the ability to display different actual colors and effects depending on their context within the app and system appearance settings.   On Android, colors are Integers.    Other platforms such as Windows and macOS also will support PlatformColor as well as platform specific functions that require more than one argument.   

In short, the intention of PlatformColor() and the other color creation functions suffixed with platform names (-IOS, -Android, etc.) are intended to expose the richness of each platform's color creation abilities but *not* to create cross-platform or platform agnostic color styling abilities.

docs/colors.md
```diff
@@ -31,6 +31,25 @@ React Native supports also colors as an `int` values (in RGB color mode):
...
+
+```jsx
+DynamicColorIOS({light: [Object object], dark: [Object object]}): [Object object]
```
TheSavior, line 37: `[Object object]` probably isn't super great for this since it could be "red", or "rgb(128,255,255)". How else could we make this signature seem like it supports any color for light and dark?

Also, just like PlatformColor, the return type isn't needed

website/i18n/en.json
TheSavior: I'm curious what Rachel will say here, but I'm not sure why `DynamicColorIOS` would be treated differently than `PlatformColor` since they are the essentially the same.

docs/colors.md
tom-un: Perhaps merely 
```
DynamicColorIOS({light: color, dark: color})
```
?

tom-un:
> > Ah, I didn't realize that distinction. PlatformColor handles all the cases that ColorAndroid does? In that case, yeah, removing ColorAndroid is reasonable.
> 
> Ya, on the native side its identical. I'll make a PR to remove it. Thanks!

ColorAndroid removal PR out: https://github.com/facebook/react-native/pull/28577

jfrolich:
@tom-un: Thanks for taking the time of explaining this! I can see the trade-off here, and this is probably the best call. I do think a higher-level API to support dark-mode would be nice, as it's something that is a thing on most platforms right now. 

With this lower level API we need to trigger a rerender on android and we don't on iOS. While I do think there would be a nice way to do this in user-land with hooks, the nicest API of defining a color that will automatically change without the developer explicitly detecting a change between dark/light mode and rendering would be hard to implement in userland. This would be much faster as it doesn't need the whole reconciliation of the app (also it would work with more imperative API's like `react-native-navigation`). I also think probably Android is going to support dynamic colors like iOS in the near future (have a native API for it). 

Anyway I think going for the lower level API first is definitely the right approach, and it would already be a big win!

rachelnabors:
I've had a chat with @kikisaints, and it sounds like it would be great if `DynamicColorIOS` could one day live on its own page and be referenced from the Colors page. But for now, it's a bit beyond the scope of this PR. I'll file an issue, though, so Kiki can move on to the more _pressing_ issue at hand of `Pressable`! Thanks everyone!

TheSavior:
I think it is fine to follow up with making DynamicColorIOS a separate page and consistent with PlatformColor as a follow up. However, we do need `PlatformColor` to be mentioned on the `colors` page, even if it is a link to the standalone page.

PR #1777, opened 2020-03-24
Correct command to init with TS

SConaway:

PR #1819, opened 2020-04-04
Update getting-started.md

voslartomas:
Fix `--template` param.


Simek:
Refs #1777, we really need to sort this out, heh 😉

voslartomas:
I think we can close this issues, because referenced issues is already closed. So I guess it's been solved.

PR #1822, opened 2020-04-06
Update Typescript template name in getting-started.md

agazso:
It seems that there was a change in the `react-native init` command when using a template argument, because it fails to me if I try the way it's written on the website.

```
$ npx react-native init projectname --template react-native-template-typescript
# ...
# omitting irrelevent lines
info Fetching template react-native-template-react-native-template-typescript...
yarn add v1.19.2
[1/4] 🔍  Resolving packages...
warning jest > jest-cli > jest-config > jest-environment-jsdom > jsdom > left-pad@1.3.0: use String.prototype.padStart()
warning jest > jest-cli > jest-config > jest-environment-jsdom > jsdom > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
error An unexpected error occurred: "https://registry.yarnpkg.com/react-native-template-react-native-template-typescript: Not found".
info If you think this is a bug, please open a bug report with the information provided in "projectname/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
yarn remove v1.19.2
[1/2] 🗑  Removing module react-native-template-react-native-template-typescript...
error This module isn't specified in a package.json file.
info Visit https://yarnpkg.com/en/docs/cli/remove for documentation about this command.
warn Failed to clean up template temp files in node_modules/react-native-template-react-native-template-typescript. This is not a critical error, you can work on your app.
(node:31885) UnhandledPromiseRejectionWarning: Error: Command failed: yarn add react-native-template-react-native-template-typescript
```

If you try with template name as `typescript` then it works:
```
$ npx react-native init projectname --template typescript 
```

PR #1828, opened 2020-04-08
fix incorrect template example in getting started

murat-mehmet:


murat-mehmet:
Also there is another bug with the `--template` switch that it won't be applied to the project. It just creates a template folder. This is not right place anyway.

Simek:
Please check out PR #1777 disscussion and https://github.com/react-native-community/react-native-template-typescript#note-on-the-legacy-cli usage note.

Duplicate of #1822.

murat-mehmet:
Yep, thanks.

Issue #333, opened 2018-05-02
Shouldn't this be called Android/Apple TV?


moar55:
I am referring to this readme: `react-native-website/docs/building-for-apple-tv.md`

isa3bel:
Can you be more specific in the problem?

Issue #1279, opened 2019-09-20
TV documentation fixes needed


brodybits:
__updated:__

On **Building For TV Devices** web page:

* ~~an internal Android tab seems to have no information~~
* it looks like some info _still_ needs to be updated now that Apple TV support was moved into the fork
* ~~The internal page URL and title are not consistent (see below)~~

Out of tree page:

* clarify or revert the changes from PR #1250 since the tvOS support is a fork and not a separate, out-of-tree module (see react-native-community/react-native-tvos#11 & react-native-community/react-native-tvos#14), my bad

Open question:

I am not certain whether or not Apple TV (tvOS) and Android TV will be supported in the same fork in the future. react-native-community/react-native-tvos#14 indicates that this may be the case, maybe the fork should be renamed if it will cover both tvOS and Android TV.

see also: #333

/cc @dlowder-salesforce @cpojer @charpeni

rickhanlonii:
cc @rachelnabors, the [Building For TV Devices](https://reactnative.dev/docs/building-for-apple-tv) is out of date - starting with 0.62 users will need to begin using https://github.com/react-native-community/react-native-tvos to develop TV devices.

brodybits:
This issue still does not seem to be resolved.

Simek:
> This issue still does not seem to be resolved.

Hi @brodybits, can you elaborate on that? What is still missing in the TV devices docs?

brodybits:
The info in the tvOS tab still seems to have information like this:

> - _Native layer_: React Native Xcode projects all now have Apple TV build targets, with names ending in the string '-tvOS'.

> - _react-native init_: New React Native projects created with `react-native init` will have Apple TV target automatically created in their XCode projects.

The out-of-tree page seems to still have some incorrect info, which should be corrected in PR #2308.

Simek:
> The info in the tvOS tab still seems to have information like this:

True, but before those bullets you can see a notice that part of documentation is deprecated. Do you think it will be better to remove the whole Android TV content? It is not what we usually do, for example `Geolocoation` has been removed from the core in [`0.60`](https://reactnative.dev/blog/2019/07/03/version-60), but the docs content is still in there - https://reactnative.dev/docs/geolocation.

PR #1859, opened 2020-04-16
add deprecation note to the Apple TV (tvOS) guide

Simek:
This PR adds deprecation note to the `Building For TV Devices` guide according to `0.62` [release blog](https://reactnative.dev/blog/#moving-apple-tv-to-react-native-tvos) post, refs #1279.

### DIFF
--- a/docs/building-for-apple-tv.md
+++ b/docs/building-for-apple-tv.md
@@ -3,6 +3,8 @@ id: building-for-apple-tv
 title: Building For TV Devices
 ---
 
+> **Deprecated.** Use [react-native-tvos](https://github.com/react-native-community/react-native-tvos) instead. For the details please check the [0.62 release blog post](https://reactnative.dev/blog/#moving-apple-tv-to-react-native-tvos).
+
 TV devices support has been implemented with the intention of making existing React Native applications work on Apple TV and Android TV, with few or no changes needed in the JavaScript code for the applications.
 
 <div class="toggler">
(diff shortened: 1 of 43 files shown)


charpeni:

Building with commit cd9b66fee6863c5bebcf48afd8864a0b973614b4

https://app.netlify.com/sites/react-native/deploys/5e98bf7f7d029f0006a85e29

rachelnabors:
Great catch! I was thinking: we should remove this from the sidebar as well _or_ we should retool the page as a jumping off point for those other community projects. What do you think?

Simek:
From the information I have acquired only `tvOS` platform has been deprecated and building for Android TV instead of phones requires small amount of changes in the code so, as is it for now, it stays within core.

This also means that the whole page is NOT deprecated, so at least for now, page should remain in the sidebar.

I have also renamed the page in current and next docs from `building-for-apple-tv` to more general and more accurate `building-for-tv` name, this change also fixes #333.

rachelnabors:
How about adding a redirect from the old building-for-tv to the new as well?

Simek:
@rachelnabors Unfortunately Docusaurus do not have an ability to generate redirects. Only thing that I could do I revert `building-for-apple-tv` files and replace the content with note with link. Example: 
* https://reactnative.dev/docs/navigator

rachelnabors:
Check the redirects in **website/static/docs** as per our chat!

PR #1862, opened 2020-04-20
addition of Pressable to apis

kikisaints:
PR to add the pressable component to the API's section of the doc site.

There are some callouts:

@TheSavior - Props list was pulled from [this merge](https://github.com/facebook/react-native/commit/3212f7dfe82d187e27f1410c8c3cb1d9fb9f5094) into React Native master, please help me ensure they are correct and the "required" field is filled out correctly.
 > e.g. I know onPress is required for button, but it doesn't seem to be for Pressable?

@rachelnabors Snack example does not currently exist, because as previously discussed, Expo is not on the latest version yet. So static example will have to do for now, please help me ensure this looks sufficient.


docs/pressable.md
```diff
@@ -0,0 +1,175 @@
...
+
+Enables the Android ripple effect and configures its color.
+
+| Type       | Required | Platform |
+| ---------- | -------- | -------- |
+| ColorValue | No       | Android  |
+
+### `children`
+
+Either children or a render prop that receives a boolean reflecting whether the component is currently pressed.
```
TheSavior, line 91: Maybe instead of "render prop" it should say "function"

docs/pressable.md
```diff
@@ -0,0 +1,175 @@
...
+| ------- | -------- | -------- |
+| boolean | No       | Android  |
+
+### `android_rippleColor`
+
+Enables the Android ripple effect and configures its color.
+
+| Type       | Required | Platform |
+| ---------- | -------- | -------- |
+| ColorValue | No       | Android  |
```
TheSavior, line 87: should this link to the colors page?

docs/pressable.md
```diff
@@ -0,0 +1,175 @@
...
+  let textLog = '';
+  if (timesPressed > 1) {
+    textLog = timesPressed + 'x onPress';
+  } else if (timesPressed > 0) {
+    textLog = 'onPress';
+  }
+
+  return (
+    <View>
+      <Pressable
```
TheSavior, line 37: I think the example should change which style it uses based on press state too, since that is one of the key parts of Pressable

docs/pressable.md
yungsters: Also, it receives an object with the press state, of which `pressed` is a boolean. Right?

docs/pressable.md
rachelnabors: Agreed. Is this possible?

kikisaints: I can do it with inline conditionals - @TheSavior were you thinking of a method to do this in particular or is any approach fine for this case?

TheSavior: Inline conditional is what I had in mind! 🙂

docs/pressable.md
```diff
@@ -0,0 +1,175 @@
...
+```jsx
+<Pressable onPress={onPressFunction}>
+  <Text>I'm pressable!</Text>
+</Pressable>
+```
+
+## Example
+
+```
+import * as React from 'react';
```
Simek, line 17: It would be nice to avoid `*` in import, can you change this line to `import React, { useState } from 'react';`?

docs/pressable.md
```diff
@@ -0,0 +1,175 @@
...
+import {
+  Pressable,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
+
+export default function App() {
```
Simek, line 25: Can you also use arrow function here (just for the consistency with the rest of example)?

docs/pressable.md
```diff
@@ -0,0 +1,184 @@
...
+```
+import React, { useState } from 'react';
+import {
+  Pressable,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
+
+export default function App() => {
```
Simek, line 25: `export default App = () => {` is the correct syntax. Current code produces `Unexpected token` error.

kikisaints: thank you! I fixed it.

docs/pressable.md
```diff
@@ -0,0 +1,184 @@
...
+
+  return (
+    <View>
+      <Pressable
+          onPress={() => {
+            setTimesPressed(current => current + 1);
+          }}>
+          {({pressed}) => (
+            <Text
+              style={[styles.text, pressed ? styles.textPress : styles.textRest]}>
```
TheSavior, line 43: This isn't quite what I meant, I meant using the `style` prop on Pressable, so that it shows that it takes a function just like `children` do. 

I was thinking of having it use style like this example from RNTester: https://github.com/facebook/react-native/blob/2c473e1a38c35957fe80b6c334a1983c034c2bbc/RNTester/js/examples/Pressable/PressableExample.js#L333-L336

docs/pressable.md
```diff
@@ -0,0 +1,177 @@
...
+        onPress={() => {
+          setTimesPressed((current) => current + 1);
+        }}
+        style={({pressed}) => [
+          {
+            backgroundColor: pressed ? 'rgb(210, 230, 255)' : 'white',
+          },
+          styles.wrapperCustom,
+        ]}>
+        <Text style={styles.text}>Press Me</Text>
```
TheSavior, line 42: Sorry, I was thinking that this example would show both the style and children function API. 

```suggestion
        {({pressed}) => (<Text style={styles.text}>{pressed ? 'Pressed!' : 'Press Me'}</Text>)}

```

kikisaints:
@rachelnabors pinging this! Let me know if you can merge 😅

rachelnabors:
Hey Kiki!

Just got back from organizing Women of React Conf. Great job on ferrying Pressable forward, and thank you for your patience!

I have just a few things:

— Suggest it as an alternative to the Touchables by linking to it from each the Touchable components pages
- What is "RectOrSize"?
- I think it's worth light paragraph at the top describing some of the Pressable can do: the `onPressIn` -> `onPress` -> `onPressOut` cycle, that can delay that, and what `hitSlop` and `pressReactOffset` are—and why you might use them. This will be brand new to some folks out there, and they'll be super pumped to understand the appeal of this exciting new component!

kikisaints:
> Hey Kiki!
> 
> Just got back from organizing Women of React Conf. Great job on ferrying Pressable forward, and thank you for your patience!
> 
> I have just a few things:
> 
> * Suggest it as an alternative to the Touchables by linking to it from each the Touchable components pages
> * What is "RectOrSize"?
> * I think it's worth light paragraph at the top describing some of the Pressable can do: the `onPressIn` -> `onPress` -> `onPressOut` cycle, that can delay that, and what `hitSlop` and `pressReactOffset` are—and why you might use them. This will be brand new to some folks out there, and they'll be super pumped to understand the appeal of this exciting new component!

Sure! Sorry for the delay, I can make bullet 1 happen.
@TheSavior could you help explain to me some of the stuff Rachel is asking for in bullet 3?

TheSavior:
@kikisaints for hitSlop and pressRect, check out this ascii art documentation in the code:

https://github.com/facebook/react-native/blob/16ea9ba8133a5340ed6751ec7d49bf03a0d4c5ea/Libraries/Pressability/Pressability.js#L326-L345

All of the touchables have always had these too, but they may not have been documented well either.

TheSavior:
I'm curious how @rachelnabors feels about including that ascii art in the docs. The words on that ascii art in code are intended for developers working on Pressability. I don't think we should be mentioning Pressability in the docs as it isn't exposed and is purely an implementation detail. 

I do think this particular example is the epitome of "a picture is worth 1000 words" though. I defer to Rachel on the execution of how to share the understanding of those props to docs readers.

kikisaints:
> I'm curious how @rachelnabors feels about including that ascii art in the docs. The words on that ascii art in code are intended for developers working on Pressability. I don't think we should be mentioning Pressability in the docs as it isn't exposed and is purely an implementation detail.
> 
> I do think this particular example is the epitome of "a picture is worth 1000 words" though. I defer to Rachel on the execution of how to share the understanding of those props to docs readers.

I agree it's a little weird, but honestly feels like it illustrates the point clearer than I could with just words. I'm up for adding in a graph instead, but that will delay this PR since I won't have the time to dedicate to a graph until late in the week.

rachelnabors:
I am happy to illustrate the `hitSlop/Rect` areas myself and doubt that the illustration needs to be quite so in depth as the ASCII art.

As for `RectOrSize`, it takes a numeric or an object like so:

```
{
  bottom: ?number,
  left: ?number,
  right: ?number,
  top: ?number,
}
```

We need to explain this format on the page, but the docs as a reference are not normalized for displaying complex objects yet. @simek and I have been working on it, though, so we can probably figure out a good way forward together.

rachelnabors:
I added some illustrations and updated the text that didn't make sense in light of the source code—please correct me if I'm wrong!

TheSavior:
How can I see the live demo of your change? I don't see a link with your commit. Hmm

Simek:
@TheSavior Link for the preview deploy is the same as earlier, site is updated with each commit:
* https://deploy-preview-1862--react-native.netlify.app/docs/next/pressable

rachelnabors:
True story, ended up with a weird conflict of IDE, hub tool, and remotes and accidentally merged this PR. Let me see if I can unmake this mess :)

PR #1872, opened 2020-04-24
Added  cleanup method to interaction manager hook example

miketamis:


Simek:
Can you provide more information about the change?

miketamis:
Yeap, I have added a cleanup method for ` InteractionManager.runAfterInteractions(() => onShown());` because otherwise if runAfterInteractions is fired after its unmounted it will cause an error.

PR #1874, opened 2020-04-25
delete unused component  in Props Examples

luism3861:
this pr only remove component import in both examples.


Simek:
Can you also convert `LotsOfGreetings` to the arrow function?

luism3861:
> Can you also convert `LotsOfGreetings` to the arrow function?

thanks i change your suggestion!!

PR #1875, opened 2020-04-25
update  height and width examples with arrow functions

luism3861:
this PR i update both examples with arrow functions!

### DIFF
--- a/docs/height-and-width.md
+++ b/docs/height-and-width.md
@@ -10,10 +10,10 @@ A component's height and width determine its size on the screen.
 The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are unitless, and represent density-independent pixels.
 
 ```SnackPlayer name=Height%20and%20Width
-import React, { Component } from 'react';
+import React from 'react';
 import { View } from 'react-native';
 
-export default function FixedDimensionsBasics() {
+const FixedDimensionsBasics = () => {
     return (
       <View>
         <View style={{width: 50, height: 50, backgroundColor: 'powderblue'}} />
@@ -22,6 +22,8 @@ export default function FixedDimensionsBasics() {
       </View>
     );
 }
+
+export default FixedDimensionsBasics;
 ```
 
 Setting dimensions this way is common for components that should always render at exactly the same size, regardless of screen dimensions.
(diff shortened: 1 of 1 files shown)


docs/height-and-width.md
```diff
@@ -10,10 +10,10 @@ A component's height and width determine its size on the screen.
 The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are unitless, and represent density-independent pixels.
 
 ```SnackPlayer name=Height%20and%20Width
-import React, { Component } from 'react';
+import React from 'react';
 import { View } from 'react-native';
 
-export default function FixedDimensionsBasics() {
+const FixedDimensionsBasics = () => {
```
Simek, line 16: You can use export with arrow function in the same line: `export default FixedDimensionsBasics = () => {`, same with example below.

MattGoldwater: @Simek I don't think this is correct. I'm getting an error when I use that code and with the ScrollView example here [ScrollView](https://reactnative.dev/docs/using-a-scrollview)

StackOverflow seems to back this up: [https://stackoverflow.com/questions/36458697/is-it-possible-to-export-arrow-functions-in-es6-7](https://stackoverflow.com/questions/36458697/is-it-possible-to-export-arrow-functions-in-es6-7) though one comment said the syntax you suggested worked. Maybe it could vary by browser? I'm using Chrome 81.

MattGoldwater: Actually, the FlatList example from [UsingListView](https://reactnative.dev/docs/using-a-listview) works for me.

Simek: @MattGoldwater No errors for me Chrome 81 on macOS and Win. It is [correct code](http://www.ecma-international.org/ecma-262/6.0/#table-42) and should be valid.

There are discussion about exports is JS community for years without a decisive conclusion. We were having this discussion here few times also, the latest could be found in the #1579 mega-issue. There are different approaches, but at the end is matter of the preference. 

ES6+ specification is very flexible according to `export`s. I think that for example [this comment](https://github.com/airbnb/javascript/issues/1275#issuecomment-274367523) states a good point why inline exports should be preferred. There are some small caveats base on what transpiler you use, in case of Babel you can test and see the transposed code in the [REPL](https://babeljs.io/en/repl). There are slight difference between the code produced by each export method, but overall all of them are valid.

luism3861: i change both example and it works !!
thanks for your suggestion @Simek :)

PR #1884, opened 2020-04-28
Update stylesheet.md

reinismu:
This is very confusing to new people that `StyleSheet.create` doesn't state performance. I think you can achieve better design patters without using it. Only reason I see myself using it is if there are any performance benefits, but it seems not be the case.

### DIFF
--- a/docs/stylesheet.md
+++ b/docs/stylesheet.md
@@ -5,6 +5,8 @@ title: StyleSheet
 
 A StyleSheet is an abstraction similar to CSS StyleSheets
 
+> **NOTE:** As it is now there are no performance benefits of using `StyleSheet` instead of plain objects. 
+
 ```SnackPlayer name=StyleSheet
 import React from "react";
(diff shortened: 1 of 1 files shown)


rachelnabors:
Is there a performance hit for using Stylesheet?

reinismu:
Question is "Is there a performance hit for **not** using Stylesheet?"

There should be description on why thing as `Stylesheet` exists. If it is legacy thing then users need to know it. 

I opened this PR to start a discussion as I couldn't find any other source of info for this.

For example this https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheet.js#L218 is it true? Somewhere else I see that it is not true anymore. We should have this kind of information in docs.

rachelnabors:
Thanks for thinking of us, but it turns out this is just not in our purview at this moment. We may revisit in the future, but not today. Thanks for bringing it up!

PR #1890, opened 2020-04-30
update TextInput.md examples with arrow functions

luism3861:
This Pr I update both examples with arrow functions!!


rachelnabors:
Alas! We are switching away from arrow functions for the examples at this time due to some issues with Prettier and ESLint

PR #1895, opened 2020-05-01
update Flatlist examples with arrow functions

luism3861:
this pr only update both examples with arrow functions

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -40,7 +40,7 @@ const DATA = [
   },
 ];
 
-function Item({ title }) {
+const Item = ({ title }) => {
   return (
     <View style={styles.item}>
       <Text style={styles.title}>{title}</Text>
@@ -48,7 +48,7 @@ function Item({ title }) {
   );
 }
 
-export default function App() {
+export default App = () => {
   return (
     <SafeAreaView style={styles.container}>
       <FlatList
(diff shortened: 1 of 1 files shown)


rachelnabors:
We're actually going to standardize to `function()` syntax because it seems more widely used in training materials as well as some issues with ESLint and Prettier.

luism3861:
> We're actually going to standardize to `function()` syntax because it seems more widely used in training materials as well as some issues with ESLint and Prettier.

ok rachel can i take some examples and convert with functions ?

luism3861:
> > We're actually going to standardize to `function()` syntax because it seems more widely used in training materials as well as some issues with ESLint and Prettier.
> 
> ok rachel can i take some examples and convert with functions ?

@rachelnabors

rachelnabors:
That would rock!! Yes yes, please! 

To be clear, it's just the function definitions themselves. I'll let @Simek explain more :)

luism3861:
> That would rock!! Yes yes, please!
> 
> To be clear, it's just the function definitions themselves. I'll let @Simek explain more :)

thanks :) !!!

luism3861:
@Simek  could you lend me explain about how to convert  examples with functions?

Simek:
@luism3861 At the beginning I want to thank you for all the contributions you have made lately, good stuff!

According to the arrow functions situation: We are working on the bigger code standardization issue (something like Component Docs Drive) so the whole community got the opportunity to contribute. Please look for it in the next few days!

luism3861:
> @luism3861 At the beginning I want to thank you for all the contributions you have made lately!
> 
> According to the arrow functions situation: We are working on the bigger code standardization issue (something like Component Docs Drive) so the whole community got the opportunity to contribute. Please look for it in the next few days!

thank you @Simek  i will check issues for this new opportunity to contribute :D !!

PR #1900, opened 2020-05-04
Running all pages through prettier

rachelnabors:
Had to be done!

### DIFF
--- a/docs/easing.md
+++ b/docs/easing.md
@@ -87,8 +87,8 @@ export default App = () => {
         sections={SECTIONS}
         keyExtractor={(item) => item.title}
         renderItem={({ item }) => (
-          <TouchableOpacity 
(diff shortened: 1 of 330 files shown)


docs/easing.md
```diff
@@ -359,7 +359,7 @@ A useful tool to visualize cubic bezier curves can be found at http://cubic-bezi
 ### `in()`
 
 ```jsx
-static in(easing);
+static in easing;
```
Simek, line 362: This is still a Prettier issue, can we [ignore](https://prettier.io/docs/en/ignore.html#markdown) this line? 😃

rachelnabors: Just this one line in this one file? Used these: https://prettier.io/docs/en/ignore.html

Anything else we should ignore?

Simek: It seems like the only one place.

PR #1902, opened 2020-05-05
Add workflow page "Using Libraries"

brentvatne:
- It's a safe bet to say that every React Native project ever shipped contains some kind of third party JavaScript library, and so we should help users to understand this and to learn how they can integrate libraries into their projects. This came up in a discussion with several folks interested in https://reactnative.directory/
- In my mind, this is the first step towards creating a first-class section on the topic, similar to the [flutter.dev Packages & Plugins section](https://flutter.dev/docs/development/packages-and-plugins/using-packages) (shout-out to Flutter folks, nice work).
- I included a backport to 0.62 docs because this isn't tied to a specific React Native version and would be immediately useful. Feedback is probably best kept to just docs/libraries.md as the versioned copy is identical.
- This was really hard to write, feedback much appreciated.

### DIFF
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -24,7 +24,7 @@ To install a library in your project, navigate to your project directory in your
 npm install react-native-webview
 ```
 
-The library that we have just installed includes native code, and we need to link to our app before we use it.
+The library that we installed includes native code, and we need to link to our app before we use it.
 
 ## Linking Native Code on iOS
 
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -1,0 +1,77 @@
+---
+id: libraries
+title: Using Libraries
+author: Brent Vatne
+authorURL: https://twitter.com/notbrent
+description: This guide introduces React Native developers to finding, installing, and using third-party libraries in their apps.
+---
+
+React Native provides a set of built-in [Core Components and APIs](./components-and-apis) ready to use in your app. You're not limited to the components and APIs bundled with React Native. React Native has a community of thousands of developers. If the Core Components and APIs don't have what you are looking for, you may be able to find and install a library from the community to add the functionality to your app.
+
+## Selecting a Package Manager
+
+React Native libraries are typically installed from the [npm registry](https://www.npmjs.com/) using a Node.js package manager such as [npm CLI](https://docs.npmjs.com/cli/npm) or [Yarn Classic](https://classic.yarnpkg.com/).
+
+If you have Node.js installed on your computer then you already have the npm CLI installed. Some developers prefer to use Yarn Classic for slightly faster install times and additional advanced features like Workspaces. Both tools work great with React Native. We will assume npm for the rest of this guide for simplicity of explanation.
+
+> 💡 The terms "library" and "package" are used interchangably in the JavaScript community.
+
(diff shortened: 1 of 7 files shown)


Simek:
The CI fails are related to the changes introduced in this PR. Please update your code and fix this issues out.

docs/libraries.md
```diff
@@ -0,0 +1,77 @@
...
+author: Brent Vatne
+authorURL: https://twitter.com/notbrent
+description: This guide introduces React Native developers to finding, installing, and using third-party libraries in their apps.
+---
+
+React Native provides a set of built-in [Core Components and APIs](./components-and-apis) ready to use in your app. You're not limited to the components and APIs bundled with React Native. React Native has a community of thousands of developers. If the Core Components and APIs don't have what you are looking for, you may be able to find and install a library from the community to add the functionality to your app.
+
+## Selecting a Package Manager
+
+React Native libraries are typically installed from the [npm registry](https://www.npmjs.com/) using a Node.js package manager such as [npm CLI](https://docs.npmjs.com/cli/npm) or [Yarn Classic](https://classic.yarnpkg.com/).
```
Simek, line 13: "Yarn Classic" link is not working (too many redirects). It looks like `https://classic.yarnpkg.com/lang/en/` is the correct URL.

BTW. Are there any specific reasons why the v1 is recommended?

ide: Yarn v2 installs packages without `node_modules`, which isn't compatible with some prior code that made assumptions about `node_modules`, as you might expect. https://github.com/yarnpkg/berry/issues/766

brentvatne: @Simek - that's strange - what browser are you using? might also be a locale issue, I used the bare https://classic.yarnpkg.com/ url so it can redirect to the appropriate language, eg: for me https://classic.yarnpkg.com/lang/en/ - but we could just use the english language link if that's causing issues for people.

docs/libraries.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Installing a Library
+
+To install a library in your project, navigate to your project directory in your terminal and run `npm install <name-of-the-library>`. Let's try this with `react-native-webview`:
+
+```bash
+npm install react-native-webview
+```
+
+The library that we installed includes native code, and we need to link to our app before we use it.
```
brentvatne, line 27: i wish we had an easy way for developers to determine "does this library include native code and require me to rebuild the native project?"

docs/libraries.md
Simek: @ide Thank you for the explanation and the link. But to be 100% correct, according to the issue you provided and the last post which summarized the whole discussion, V2 supports installation of `node_modules` but it requires the additional configuration file (`.yarnrc.yml`). So it's reasonable until the whole situation settles and the V2 stables to recommend using the V1.

@brentvatne Clearing the local cache solved the "Too many redirects" issue, but the redirect is still there. Since language select is available from the top menu and until RN website doesn't support i18n I think that it's safer to use the full link to English version to avoid similar issues brought by the community in the future.

brentvatne: thanks for checking @Simek! I'll update to point to https://classic.yarnpkg.com/en/ then

PR #1906, opened 2020-05-05
add ability to highlight new pages in the sidebar

Simek:
This PR introduces ability to mark pages in the main (left) sidebar as "New". This was a nice-to-have feature for a some time, but Docusaurs do not offer a build-in solution or feature for adding such marks and produced HTML do not have any common used anchors. 

To solve this issue I decided to use `href` attribute and "contain" [selector](https://www.w3schools.com/cssref/sel_attr_contain.asp) to match the desired pages. 

The downside of this solution is that the CSS should be updated at least on each version cut.

I have tested those changes with the latest browsers on Win (Chrome, FF, Edge), macOS (Chrome, Safari) and iOS (Chrome, Safari). Few CSS properties has been added to ensure that the design is similar everywhere.

## Preview
Annotation 2020-05-05 235649

### DIFF
--- a/website/static/css/docs.css
+++ b/website/static/css/docs.css
@@ -229,3 +229,27 @@ figcaption {
   color: #134484;
   border-bottom-color: #134484;
 }
+
+/*
+  Mark pages as "new" using href attribute:
+  - to show badge in all versions use "PAGEID"
+  - to restrict badge only to the current/default version use "docs/PAGEID"
+  - to restrict badge only to the next version use "next/PAGEID"
+*/
+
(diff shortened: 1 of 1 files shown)


Simek:
I don't have any preference according to the color used here (currently it is red with little bit of shift towards orange - [#ff5722](https://www.colorhexa.com/ff5722)). 

Feel free to update the PR and change the HEX as you wish, every color works for me.

Simek:
From the preview/mockup the light green color has been selected:
new-badge

rachelnabors:
Let's go with green!

PR #1909, opened 2020-05-07
fix: add missing Appearance module APIs

thymikee:
## Summary

Added missing `addChangeListener` and `removeChangeListener` APIs present in Appearance module: https://github.com/facebook/react-native/blob/0.62-stable/Libraries/Utilities/Appearance.js
Updated docs for 0.62 and current.


rachelnabors:
tuxedo-thanks
Thanks so much for keeping us up to date!

PR #1912, opened 2020-05-08
edited documentation for macOS and android

christinavoudouris:
Hey, first off, thank you for the thorough documentation! I just noticed some redundant, slightly confusing instructions for Android and MacOS. I thought they should match the instructions for iOS.


Simek:
Hi @christinavoudouris, thank you for spotting this and working on the improvement! 

I have also changed the sections arrangement a bit (`yarn` notice is at the top now), I have updated commands to utilize `npx` everywhere and I have tweaked wording a bit in the Metro sentece to avoid confusion/redundancy.

PR #1913, opened 2020-05-08
remove RCTLocationObserver.m link in native modules ios

luism3861:
this PR remove RCTLocationObserver.m, when you click this link appears 404, about version 0.58 it works but 59 or greater it doesn't work.


luism3861:
> Good catch! According to the [CHANGELOG](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#removed-2) `Geolocation` was removed in RN `0.60`.
> 
> Can you also update the current, `0.61` and `0.60` docs?
> 
> It would be nice if we can provide alternative example from the RN repo for those versions, but if there are no suitable one I think the removal is acceptable.

i try to find native-modules-ios.md in versioned_docs  0.60, 0.61 and it doesn't exist,  i think @Simek removal this file in  RN 0.62 is acceptable.

Simek:
If there are no files for exact version you need to create `0.60` versioned page copy and remove the note there (so the old version will only load for docs below `0.59` below). Also the [current version](https://github.com/facebook/react-native-website/blob/master/docs/native-modules-ios.md) of the page need an update.

luism3861:
> If there are no files for exact version you need to create `0.60` versioned page copy and remove the note there (so the old version will only load for docs below `0.59` below). Also the [current version](https://github.com/facebook/react-native-website/blob/master/docs/native-modules-ios.md) of the page need an update.
> 
> You can find more about docs versioning [here](https://docusaurus.io/docs/en/versioning).

Hi @Simek  i created exact version and copy native_modules_ios.md versioned `0.60` and `0.61` and i added this file in versiones_docs and delete this link in both versions, and new version can you check this commit thanks!!

luism3861:
> > If there are no files for exact version you need to create `0.60` versioned page copy and remove the note there (so the old version will only load for docs below `0.59` below). Also the [current version](https://github.com/facebook/react-native-website/blob/master/docs/native-modules-ios.md) of the page need an update.
> > You can find more about docs versioning [here](https://docusaurus.io/docs/en/versioning).
> 
> Hi @Simek i created exact version and copy native_modules_ios.md versioned `0.60` and `0.61` and i added this file in versiones_docs and delete this link in both versions, and new version can you check  last  commit thanks!!

rachelnabors:
kylo_thumbsup
Thanks for the update!

PR #1915, opened 2020-05-08
Need go back to project root before run testing

fffx:
Since the previous step is:
```
$ cd android
$ ./gradlew bundleRelease
```
so next step we need to go back to the project root folder

### DIFF
--- a/docs/signed-apk-android.md
+++ b/docs/signed-apk-android.md
@@ -93,7 +93,7 @@ _Note: In order for Google Play to accept AAB format the App Signing by Google P
 
 ### Testing the release build of your app
 
-Before uploading the release build to the Play Store, make sure you test it thoroughly. First uninstall any previous version of the app you already have installed. Install it on the device using(in project root):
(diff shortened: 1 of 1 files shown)


fffx:
@Simek @rachelnabors  Thanks, it's updated

docs/signed-apk-android.md
```diff
@@ -93,7 +93,7 @@ _Note: In order for Google Play to accept AAB format the App Signing by Google P
 
 ### Testing the release build of your app
 
-Before uploading the release build to the Play Store, make sure you test it thoroughly. First uninstall any previous version of the app you already have installed. Install it on the device using:
+Before uploading the release build to the Play Store, make sure you test it thoroughly. First uninstall any previous version of the app you already have installed. Install it on the device using(in project root):
```
Simek, line 96: Space is missing between last word and round bracket.

What do you think about putting this information into the sentence? 
* "Install it on the device using the following command in the project root:"

fffx: Any language advice is accepted, as I am not an English speaker

Simek: I'm not the one too, thank you for updating this PR during review!

PR #1916, opened 2020-05-10
React Native packager -> Metro bundler

EvanBacon:
Users coming from a web background will make a clearer connection to how React works when running natively if they can draw the comparison between the bundlers. 

cite: React docs: https://reactjs.org/docs/create-a-new-react-app.html#creating-a-toolchain-from-scratch


cpojer:
Question: should we use "Metro bundler" only once in the bundler specific page and then later refer to it as "Metro"? We don't say "Jest testing framework" or "React Native UI library" in the docs either.

EvanBacon:
@cpojer I think it'd be better to refer to it as the "bundler" because this is more inline with other web tools. Because most bundlers do not currently have first-class support for React iOS, or React Android runtimes, using "Metro bundler" helps convey the idea that at the moment React, when run natively, requires Metro to be bundled properly.

I think if we continue to bring awareness to the internal stack of React Native it'll make be easier to just use tools by their name.

PR #1919, opened 2020-05-11
Adding stepped instructions for running RN application

rachelnabors:

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -474,18 +474,33 @@ Click "Next" then "Finish" to create your AVD. At this point you should be able 
 
 <h2>Running your React Native application</h2>
 
-> If you use the Yarn package manager, you can use `yarn` instead of `npx` when running React Native commands inside an existing project.
+<h3>Step 1: Start Metro</h3>
 
-Run `npx react-native start` inside your React Native project folder:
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -507,18 +522,33 @@ The above command will automatically run your app on the iOS Simulator by defaul
...
-Run `npx react-native start` inside your React Native project folder:
+To start Metro, run `npx react-native start` inside your React Native project folder:
 
 ```sh
-cd AwesomeProject
 npx react-native start
 ```
 
-The command above starts Metro Bundler, which you can read more about [here](https://facebook.github.io/metro/).
+`react-native start` starts Metro Bundler.
```
Simek, line 535: Can we use `npx` in this command also? Just for the sake of consistency 😄

rachelnabors: I wanted to keep it free of commands for those using yarn... Too obscure?

Simek: I'm not a `npx` advocate, I think that sticking to one common approach is the best. In this case just all other commands on the page are using `npx`, so it looks a bit odd when this is not, but it's not a reason to hold this PR at all.

PR #1922, opened 2020-05-12
update CoreComponents function Examples

luism3861:
This PR Update Core Components with new standardize functions, some example in core components like `KeyboardAvoidingView` it has this standardize!!


rachelnabors:
shera_entrapta

Thank you so much for helping us make these docs rock!

PR #1930, opened 2020-05-13
update design docs examples with new standardize

luism3861:
This PR update all design examples with new Standardize!!


rachelnabors:
shera_entrapta

Thank you so much for helping us make these docs rock!

PR #1933, opened 2020-05-14
addition of the pressable API (con't)

kikisaints:
Reopening this due to accidental closure on the previous pr of #1862


rachelnabors:
TODO: We're going to need to document the Event and React.node types with these docs.

TheSavior:
Apparently you can only comment inline on the code on changed lines :(

> 2. [`onPress`](#onpress) is called when a single press gesture is triggered, 130 milliseconds from `onPressIn`.

I worry that this includes implementation details that can change as part of our "public api" by being in the docs. Nobody should rely on 130ms and that could be changed by us at any time. Same for the other delays in the docs. 

Should platform specific props like `android_disableSound` show up at the bottom of the list instead of the top? It's new that we are prefixing the platform specific props so we probably haven't run into this before, but sticking android only props first is probably not what people are looking for. 

> Either children or a function that receives a boolean reflecting whether the component is currently pressed.
| Type       | Required |
| ---------- | -------- |
| React.Node | No       |

The type of React.Node in the table isn't quite right since it can also take a function. Perhaps it should be something like this for both `children` and `style`?

`React.Node | (PressState) => React.Node`


----------
Also, since we can't embed a Snack for the example since Snack won't have this version, could it have an animated gif instead?

TheSavior:
Also, I think the image across the top that shows the order of events might not be quite right. press and longPress will never both fire for the same event. onPress fires after longPress, but before onPress. I feel like articulating these complexities and ordering might be out of scope for the documentation and make it more complicated than not including it. What do you think?

Also I highly recommend checking out the examples in RNTester to validate the content here. See this recording of the press events example that showed me that the image isn't quite right.

android_Phone_Screen_Recording_2020-05-14_at_12 18 40

docs/pressable.md
```diff
@@ -136,41 +138,41 @@ Duration (in milliseconds) from `onPress` before `onLongPress` is called.
...
 ### `hitSlop`
 
 Sets additional distance outside of element in which a press can be detected.
 
-| Type       | Required |
-| ---------- | -------- |
-| RectOrSize | No       |
+| Type       | Representation                                                                         | Required |
+| ---------- | -------------------------------------------------------------------------------------- | -------- |
+| RectOrSize | number <hr/> object `{ bottom: ?number, left: ?number, right: ?number, top: ?number }` | No       |
```
TheSavior, line 151: Each key is optional. As worded right now it makes it seem like every key is required, but takes `null | undefined | number`. 

I think it should be:
```
number <hr/> object `{ bottom?: number, left?: number, right?: number, top?: number }`
```

Or if we are actually using Flow syntax like we have in the codebase, technically both sides need question marks because it *does* let a key be defined and set to `null | undefined` as well.
 
```
number <hr/> object `{ bottom?: ?number, left?: ?number, right?: ?number, top?: ?number }`
```

Simek: That's a good remark, thank you! I like the second version because it's true to the code. But from the discussions and polls it seems that quite many developers are not familiar with the Flow syntax. Unfortunately (due to MD files parsing rules) it's not possible to put code blocks inside tables which restrain the possibilities to format code there.

This is the small preview which shows what can be achieved using some more CSS, but I'm not sure if it isn't too much information (there are two representations for key type for the comparison):
Annotation 2020-05-15 144730

Is it a better solution? What's your opinion?

rachelnabors:
Thanks for all your feedback, @Eli, and for clarifying how the interactions work. I actually had trouble running RNTester until last night, thus had to work with source and tests to make this illustration. I do feel the illustration helps explain the sequence of events. Changing the diagram to reflect that is not impossible :)

I agree, the 130ms can go as people might set timers by it. But the 500ms should probably stay as it's the default.

As for `RectOrSize`, perhaps it should live on its own Object Type page as well, given that many learners are unfamiliar with both flow notation and TS notation.

docs/pressable.md
rachelnabors: Not a fan of either of these, personally. Think we should go with an Object Type page, [ala MDN](https://developer.cdn.mozilla.net/en-US/docs/Web/API/Web_Animations_API/Keyframe_Formats). Better to be explicit, and we're going to have to do this with other objects on the page as well, so...

rachelnabors: Screenshot 2020-05-15 at 15 08 14
This is how that could look. Making a separate PR!

rachelnabors:
@TheSavior Almost forgot! Re Android-at-the top: all props are alphabetized now. It would be weird to put iOS and Android out of order. Perhaps we could shift them to a "Platform Props" section further down the page. Thoughts, @Simek?

docs/react-node.md
```diff
@@ -0,0 +1,13 @@
+---
+id: react-node
```
TheSavior, line 2: I feel like if this is documented it should be in the React repo and not the React Native repo. This is anything that React allows as a child. 

Is there an existing page that React.Node can link to?

Simek:
I just have pushed the initial version of the `PressEvent`. All the descriptions are based on the things I was able to deduce from the code and playing with live events. Probably not all of the them will be 100% accurate. Keys with missing description: `changedTouches`, `target` and `touches`. Hope someone with better knowledge about events and event system can correct me and fill the gaps in.

Also from the test I have made it looks for me that `force` should be marked as an option field/key in the `CoreEventTypes` definition: https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js#L91

docs/pressable.md
```diff
@@ -96,81 +99,83 @@ const styles = StyleSheet.create({
...
 ### `hitSlop`
 
 Sets additional distance outside of element in which a press can be detected.
 
-| Type       | Required |
-| ---------- | -------- |
-| RectOrSize | No       |
+| Type                                 | Required |
+| ------------------------------------ | -------- |
+| [`RectOrSize`](rectorsize) or number | No       |
```
TheSavior, line 154: `or number` is redundant. That is what RectOrSize means :)

docs/pressable.md
```diff
@@ -15,14 +15,17 @@ Pressable is a Core Component wrapper that can detect various stages of press in
...
-3. [`onLongPress`](#onlongpress) is called only if the press gesture is activated beyond 500ms from `onPressIn` or the time set with [`delayLongPress`](#delaylongpress).
-4. [`onPressOut`](#onpressout) is called when the press gesture is deactivated.
+- [`onPressIn`](#onpressin) is called when a press is activated.
+- [`onPressOut`](#onpressout) is called when the press gesture is deactivated.
 
-  <img src="/docs/assets/d_pressable_pressing.svg" width="1000" alt="Diagram of the onPress events in sequence.">
+After pressing [`onPressIn`], one of two things will happen:
 
-Fingers are not the most precise instruments, and it's not uncommon for users to "fat finger" an interface—to activate the wrong thing or miss the activation area. To help, `Pressable` has an optional `HitRect` you can use to define how far a touch can register away from the the wrapped element. Presses can start anywhere within a `HitRect`.
+1. The person will remove their finger, triggering [`onPressOut`] followed by [`onPress`](#onpress).
```
TheSavior, line 23: Should onPressOut here be a link as well? There are a couple of other ones missing links as well

docs/pressable.md
rachelnabors: How would you word the `RectOrSize` page though? It doesn't quite work right as a reference page like that...

docs/pressable.md
rachelnabors: You only have to link on the first use of a term.

docs/pressable.md
TheSavior: Maybe make it a Rect page and say `Rect or number` here?

docs/react-node.md
rachelnabors: I also got hung up on this! But talked with Dan about it, and we are following his advice verbatim with these pages. (React node is more of a typing construct than something React exports, after all.)

rachelnabors:
shera_feels-powerful

And that's a wrap, folks! Thanks @Simek for working on the label, attribute and design updates! Thanks to @kikisaints for writing the doc! And thanks to @TheSavior for providing technical direction and feedback! (And a special thanks to @yungsters for helping me figure out the sequence of events to make that illustration 100% accurate!)

Great work, everyone! This sets a new standard for how we document components and APIs :)

PR #1934, opened 2020-05-14
Adding illustrations to security doc

rachelnabors:
I had wanted to add these illustrations with the original contribution.


rachelnabors:
I didn't have time to do the whitelist bouncer :)

PR #1936, opened 2020-05-15
Adding the start of Object Types reference

rachelnabors:
We have many shared complex object types that bloat component and API references. It's time to start cutting some up into handy to reference pages, starting with those in service of `Pressable`


rachelnabors:
Love the new heading!

shera_hordak_hardwork

TheSavior:
RectOrSize is a `Rect` or a `Size`. I think this page only documented the Rect aspect of it. Instead of passing an object you could also just specify a single number that would be used for all dimensions.

PR #1951, opened 2020-05-22
Add light background to the search bar

bell-steven:
This PR changes the background of the search bar from dark to light.

There have been quite a few instances recently where I really struggled to find the search bar on the site, as there is not a whole lot of contrast between the search bar and the rest of the header.

Current search bar: 
image

Proposed search bar:
image


rachelnabors:
I think this makes sense. Thanks for contributing!
mlp_fistbump

PR #1953, opened 2020-05-26
Update security.md

ZenPylon:
Hello!  Found a minor typo

> later iOS versions (iOS 11) where the instituted a first-come-first-served principle

I changed "the instituted" to "they instituted".


rachelnabors:
klk_nice-catch
Excellent! Thanks for catching this!

PR #1966, opened 2020-06-02
update descriptions iOS simulator in running-on-simulator-ios.md

luism3861:
this PR update iOS simulators descriptions, i think` iPhone 5s` it is old  so i changed for `iPhone SE`.


luism3861:
> LGTM, thank you!
> 
> @luism3861 You might also want to update the example in the `cli`:
> 
> * https://github.com/react-native-community/cli/blob/master/packages/platform-ios/src/commands/runIOS/index.ts#L528

oh thanks  @Simek  i update in cli !!

Issue #1230, opened 2019-09-04
Cannot deep link to CLI sections


jmeistrich:
It's not possible to deep link into a section on the React Native CLI Quickstart side of the page since those elements are hidden by default.

For example:
https://facebook.github.io/react-native/docs/getting-started#node-watchman-jdk

isa3bel:
will you elaborate on the issue more? I am down to try to fix this!

jmeistrich:
Sure! The Getting Started page starts with the "Expo CLI Quickstart" section selected, so the elements under "React Native CLI Quickstart" are hidden. So it's not possible to deep link into a section of the CLI Quickstart, which we were trying to do in the doctor command of the CLI. @rachelnabors asked me to create this issue as it might be fixable when redesigning the page.

PR #1968, opened 2020-06-03
enviroment setup: remove CLI tabs, allow RN CLI linking

Simek:
Refs: #968, #1230

This PR introduces the following changes to the "Setting up the development environment" page that might be a bit a bit controversional:
* CLI toggler removed (Expo & RN) and the two sections are visible
* linkable CLI section titles added, so users can share url directly to the RN one
* few Emojis used across the page to emphasise or catch user attention to the few key sentences/places
* RN OSes tabs received a custom design just to improve their visibility
* Android before iOS in the target OSes

I understand that this is not the ideal solution for the mentioned above problems but at least I hope that the changes should help a bit. Tabs are nice a feature, great for the class/functional examples, good for small text blocks toggling but not so great in this case. I also want to avoid creating more issues for the Docusaurus V2 migrate so I don't want to expand custom JS files at this moment.

## Preview
Annotation 2020-06-03 014518

Annotation 2020-06-03 154303


rachelnabors:
I would want @brentvatne's eyes on this page before making such a major change. I feel the new page may lead to more confusion, not less, and prefer the old one.

Personally, I'd like all the togglers respect a hash in the URL for content like this or in local storage for preferences like platform/class toggles. That would solve problems on many pages and could probably be turned into a module for Docusaurus 2.

brentvatne:
I do like some of the changes here but I agree with @rachelnabors that we lose a bit of clarity. I think this comes from listing everything on one page, and without a way to hide the content that is not relevant to you.

Things I like:
- the formatting of the intro paragraph along with the emoji looks a lot more friendly and easier to parse due to better spacing.
- the development OS and target OS selectors are also a nice improvement - although could go either way with the toggle style changes, I really like the emoji to clearly call attention to them and make it obvious at a glance what you are looking at.

> Personally, I'd like all the togglers respect a hash in the URL for content like this or in local storage for preferences like platform/class toggles. That would solve problems on many pages and could probably be turned into a module for Docusaurus 2.

I agree! This would be a nice touch.

cpojer:
What's the status of this PR?

Simek:
@cpojer If I remember correctly only small tweaks from this PR were ready to merge, but main change - CLI and Expo tabs removal - was not satisfying enough. 

I think that this PR can be close for now, but it would be nice to address this after V2 migration (which will bring few more options  to improve this page to the table).

Issue #1427, opened 2019-10-23
Navigator page: anchor tags lead to 404 pages, Edit link leads to github 404


quetzaluz:
To reproduce:
- Visit https://facebook.github.io/react-native/docs/0.43/navigator.html
- Click any of the method links, for example onWillFocus
- You arrive at a 404 page at that route: https://facebook.github.io/react-native/docs/0.43/navigator.md#onwillfocus
<img width="904" alt="Screen Shot 2019-10-23 at 4 08 20 PM" src="https://user-images.githubusercontent.com/3665912/67440841-60957400-f5af-11e9-81ca-c92149b0b9e9.png">

Currently observed only on the navigator docs page.

quetzaluz:
Noting that this is also occuring for https://facebook.github.io/react-native/docs/0.43/navigatorios

Affected pages also lead to github 404 pages when you click edit, ex https://github.com/facebook/react-native-website/blob/master/docs/navigatorios.md

issue seems to be when these pages do not exist in the root docs directory

samslow:
This problem occurs in more places than you think.
I've had the same problem with the flatlist.
What these problems have in common is that they are linked to '.md'

I'll fix it.

rachelnabors:
Thanks for the heads up!

I think I see what the issue is: these links use `<componentname>.md#<anchor>` in the link name instead of `<componentname>#<anchor>` or even `<anchor>`. The solution is to scour the docs for `.md#` and replace those with `#`. It would make a fine first time contribution. Any takers?

mdb1710:
> Thanks for the heads up!
> 
> I think I see what the issue is: these links use `<componentname>.md#<anchor>` in the link name instead of `<componentname>#<anchor>` or even `<anchor>`. The solution is to scour the docs for `.md#` and replace those with `#`. It would make a fine first time contribution. Any takers?

I would be happy to tackle this for you.

rachelnabors:
I'll assign it to you then. Good luck, and if you get stuck, DM me!

yozaam:
Hey, I see that it is assigned but since he/she is inactive, I would love to do this

mdb1710:
> Hey, I see that it is assigned but since he/she is inactive, I would love to do this
> Sorry if it is a noob question,
> The changes need to be made in every file in /website/versioned_docs or in /docs as well?
> Or only in docs and docs is compiled into versioned_docs?
> https://github.com/facebook/react-native-website#versioned-docs
> 
[...]

PR #1970, opened 2020-06-03
Getting Rid of In-Content Links

jpbulman:
Fixing #1427 where some pages of older versions have broken links from the inline links to ```methods``` and ```props```


Simek:
`NavigatorIOS` also has been removed from the core but a bit later with version [0.58](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#the-slimmening-is-happening) so I have added the similar notice about that fact like on the `Navigator` page but to this component.

A little surprising was the fact that there was versioned page for the 0.58 docs for the `NavigatorIOS` but with the usual content when mentioned above commit clearly has been present in the 0.58 release.

PR #1973, opened 2020-06-05
small tweaks for the BLM banner

Simek:
This small PR fixes a display issue with the sticky navs and BLM banner. 

Also the background color of the banner has be adjusted to stand out a bit more.

#### Before
Annotation 2020-06-05 155218

#### After
Annotation 2020-06-05 155032

### DIFF
--- a/website/static/css/header.css
+++ b/website/static/css/header.css
@@ -200,12 +200,11 @@ input#search_input_react:focus {
 
 /* Overrides for the announcement banner */
 .announcement {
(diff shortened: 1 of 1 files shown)


rickhanlonii:
@Simek thanks for the fix! Fwiw, and this is just personal preference, but I liked how the banner and header blended together with the same color instead of being separate. That was intentional and IMO it looks better. See also: reactjs.org

PR #1988, opened 2020-06-10
Update getting-started.md

anwargul0x:
Configure the ANDROID_HOME environment variable
Section doesn't mentions how to open or navigate to  $HOME/.bash_profile or $HOME/.bashrc files from Terminal.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -294,6 +294,16 @@ The React Native tools require some environment variables to be set up in order 
 
 <block class="native mac linux android" />
 
+Open the Terminal and open your `$HOME/.bash_profile` or `$HOME/.bashrc` using `GNU nano`.
+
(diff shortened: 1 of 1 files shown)


Simek:
hi @anwargul0x, usually we are avoiding adding such detailed steps. 

There are many ways to edit those files (not only through command line, the preference is heavy user dependent). Also the guide could not be valid for every distro or OS version (this block is visible for Linux and macOS).

anwargul0x:
@Simek  thank you for making me realize that how stupid I was to think that there should be a step for total beginners to look how they can edit it. I mean not everyone owns Mac but sometimes they buy it for the purpose of development either learning or work. 

Instead of just closing it, wouldn't it better if you guys  edited for other OS in a generalize way.

Point learned: React Native docs has been garbage and will remain garbage.

PR #1989, opened 2020-06-11
fix: update installation instructions

developer239:
- macOS Catalina now uses Zsh instead of Bash

It took me a while to figure this out when I was installing RN on a fresh OS. 🙂 You can find more information here: https://thenextweb.com/dd/2019/06/04/why-does-macos-catalina-use-zsh-instead-of-bash-licensing/

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -294,7 +294,7 @@ The React Native tools require some environment variables to be set up in order 
 
 <block class="native mac linux android" />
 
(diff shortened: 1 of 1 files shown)


Simek:
hi @developer239, thank you for pointing this out. 

Since `zsh` can also be used on the earlier versions of macOS and on the Linux too it would be nice if you can switch the content in the brackets to something more general like: "if you are using `zsh` then...".

Also please include the include both `zsh` [profile](https://support.apple.com/en-ca/HT208050) files: `~/.zprofile` and `~/.zshrc`

developer239:
@Simek ✅

Simek:
@developer239 Thank you so much! 😄

PR #1992, opened 2020-06-11
Update state.md

sunnylqm:
Since the example is using hooks, it does not make sense to mention constructor and setState

### DIFF
--- a/docs/state.md
+++ b/docs/state.md
@@ -5,8 +5,6 @@ title: State
 
 There are two types of data that control a component: `props` and `state`. `props` are set by the parent and they are fixed throughout the lifetime of a component. For data that is going to change, we have to use `state`.
 
-In general, you should initialize `state` in the constructor, and then call `setState` when you want to change it.
-
 For example, let's say we want to make text that blinks all the time. The text itself gets set once when the blinking component gets created, so the text itself is a `prop`. The "whether the text is currently on or off" changes over time, so that should be kept in `state`.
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @sunnylqm! I'm not sure if removing the sentence is the right way to fix this issue. I rather would like to see the addition of class component example and tabs. Then the note/sentence could be placed inside the class component tab to not confuse those who are using the functional approach.

Also there are more references to the class component, for example:
> When setState is called, BlinkApp will re-render its Component. By calling setState within the Timer, the component will re-render every time the Timer ticks.

I think that @rachelnabors should decided what's the best approach there(since I'm not familiar with this page and it's purpose).

BTW. How you have landed on this page? I'm asking since it's out of the sidebar right now. Was it the App template link, search result or maybe something else?

sunnylqm:
So is this doc going to be removed? I am running a translation mirror site for this so I will going through all the docs once a while.

Simek:
@sunnylqm I don't know what's the purpose or what is the future of this page. 

> BTW. How you have landed on this page? I'm asking since it's out of the sidebar right now. Was it the App template link, search result or maybe something else?
Can you answer the question above?

And to be honest you could wait with the issue closure until @rachelnabors response 😉

sunnylqm:
I simply read all the docs in the repo and do some typo fix sometimes. 
> I am running a translation mirror site for this so I will go through all the docs once in a while.

PR #1998, opened 2020-06-15
update second example in Flatlist docs

luism3861:
this PR update `Flatlist` example i think select one button and change background color it's  a good example, not like older example to select one button it never turn off background color.

flatlistt

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -123,35 +123,34 @@ const App = () => {
     setSelected(item);
   }
 
-  const renderItemSelect = (item) => {
-    return item.id === selected.id ? (
-       <TouchableOpacity
-      onPress={() => onSelect(item)}
-      style={[
-        styles.item,
-        { backgroundColor: '#6e3b6e' },
-      ]}
-    >
-      <Text style={styles.title}>{item.title}</Text>
-    </TouchableOpacity>
-    ) : (
-        <TouchableOpacity
-      onPress={() => onSelect(item)}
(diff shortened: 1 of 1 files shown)


docs/flatlist.md
```diff
@@ -116,47 +116,42 @@ const DATA = [
...
+    setSelected(item);
+  }
+
+  const renderItemSelect = (item) => {
+    return item.id === selected.id ? (
+       <TouchableOpacity
+      onPress={() => onSelect(item)}
       style={[
         styles.item,
-        { backgroundColor: selected ? '#6e3b6e' : '#f9c2ff' },
```
Simek, line 125: Please use ternary operator for the `backgroundColor` like in the earlier version of example.

docs/flatlist.md
```diff
@@ -116,47 +116,42 @@ const DATA = [
...
+    </TouchableOpacity>
+    )
+  }
 
   return (
     <SafeAreaView style={styles.container}>
       <FlatList
         data={DATA}
-        renderItem={renderItem}
+        renderItem={({ item }) => renderItemSelect(item)}
```
Simek, line 154: I would be nice if you can change the method signature so the `renderItem={renderItemSelect}` syntax can be used.

luism3861:
> Please also correct the white spaces 😉
> 
> But generally I like the way you have rewritten the example, good work!

thanks for your comments, i change your suggestion @Simek  thanks!!!:)

PR #2002, opened 2020-06-19
fix links in images.md

luism3861:
this PR fix links in `Static Non-Image Resources` 'metro links', in `version-0.62` it does't exists images.md so i used docusaurus docs to add ` images.md` in version-0.62 and fix this links.!

when you click this links appears error 404 so i fix that.


luism3861:
> The correct way to fix this issue will be correcting the links in the version which is load for `0.62` docs, rather than creating new page.

thanks @Simek !! i found this file in `version-0.5`

PR #2019, opened 2020-07-02
Fixed the Function Component Example

devofthings:
We experienced that the documentation was faulty so after we at @enpit found a fix, we wanted to share it with you. 
The old trigger `appState.current.match(/inactive|background/) && nextAppState === "active"` would never work because the useState hook doesn't update the state fast enough for the function to work. The solution is to use useRef instead and optional (to display the inactive status under iOS) also use a `useState` for cosmetics.

### DIFF
--- a/docs/appstate.md
+++ b/docs/appstate.md
@@ -22,9 +22,6 @@ For more information, see [Apple's documentation](https://developer.apple.com/do
 
 To see the current state, you can check `AppState.currentState`, which will be kept up-to-date. However, `currentState` will be null at launch while `AppState` retrieves it over the bridge.
 
-You can remove line 7 & 24 in the functional component example if you don't want to see the AppState update from active to inactive (iOS).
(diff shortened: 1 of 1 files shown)


devofthings:
@Simek I removed the block statement and added a descriptive text.

docs/appstate.md
```diff
@@ -22,6 +22,9 @@ For more information, see [Apple's documentation](https://developer.apple.com/do
 
 To see the current state, you can check `AppState.currentState`, which will be kept up-to-date. However, `currentState` will be null at launch while `AppState` retrieves it over the bridge.
 
+You can remove line 7 & 24 in the functional component example if you don't want to see the AppState update from active to inactive (iOS).
```
Simek, line 25: Ohh, since this comment is only related to the functional component please move it inside the `functional syntax` block.

PR #2023, opened 2020-07-06
Remove duplicate documentation property for statusBarTranslucent on Modal

levibuzolic:
This property is already documented further up this page, in addition to this instance wasn't defined as being Android-only.


levibuzolic:
@Simek done, thanks for the suggestion :+1:

PR #2025, opened 2020-07-06
add Prettier check to the CircleCI lint job

Simek:
This PR adds the Prettier check of latest docs (`next`) and blog posts to the CircleCI `lint` job. This should be run with every PR to help contributors follow the style guidelines and formatting. I decided to skip the versioned docs just to spare some time (changes there are not that often and whole check takes ~15-20x CPU time on my machine).


charpeni:
Thanks for looking into that!

As a bit of context, we didn't include it into the lint job as most people were editing the docs directly from GitHub, and most of those pull requests were blocked or would have been blocked by this check. For people that cloned the repo and installed dependencies, we have a husky pre-commit hook to run prettier.

So, what we were doing was to run prettier manually before cutting a new release or sometimes.

If we're okay that some PRs (a lot) may need a manual prettier fix from us (maintainers), I'm okay with that! But I think it may be more convivial to run it ourselves instead of forcing the contributors to do so. 😬 

Could we setup a cronjob to automatically fix prettier on master? Or rely on a GitHub action that will automatically apply prettier on pull requests? 🤔

Simek:
> As a bit of context, we didn't include it into the lint job as most people were editing the docs directly from GitHub, and most of those pull requests were blocked or would have been blocked by this check. For people that cloned the repo and installed dependencies, we have a husky pre-commit hook to run prettier.

* As you can see this PR has this check turned on and there an no Prettier issues at all (https://app.circleci.com/pipelines/github/facebook/react-native-website/2476/workflows/9851ba08-8bb1-45bb-96e9-6ca8c26ef52d/jobs/12826). It would be ideal to maintain the code quality like that and this job should help with that.

* Running git hooks on push can be easily disable when working and committing from the local copy (you can't assume that those files were checked).

> So, what we were doing was to run prettier manually before cutting a new release or sometimes.

* IMO This is not a good practice or approach.

> If we're okay that some PRs (a lot) may need a manual prettier fix from us (maintainers), I'm okay with that! But I think it may be more convivial to run it ourselves instead of forcing the contributors to do so. 😬

* GitHub allows an PR author to edit files included in the PR from the web.
* Maintainer to fix this **properly** need to add remote, checkout the PR branch and push changes there, which is very annoying and slows down the review and accept process a lot.
* When there are style guides and code formatting rules defined many repositories enforce those standards and it's usual to  expect these policies to be followed  by the contributors when editing the code.

> Could we setup a cronjob to automatically fix prettier on master? Or rely on a GitHub action that will automatically apply prettier on pull requests? 🤔

* Prettier can introduce some necessary or breaking changes (https://github.com/facebook/react-native-website/pull/1900#discussion_r419774676), so IMO the changes should not be applied automatically.

charpeni:
Sorry for the intent here, @simek.

I don't want to say that this is not a good idea, actually, **I think your pull request is the best scenario of all**. Having no Prettier issues is great, and having a CI check to ensure it should be mandatory, but.

I agree that git hooks can be easily disabled. In fact, they're rarely enabled as people who are cloning the repo aren't installing dependencies, they're just editing files and committing them.

There are 3 ways to contribute to `react-native-website`:

- Edit the file directly from GitHub or follow the Edit button from the website.
- Clone the repository locally, edit files, commit them, and wait on the deploy preview to view your changes.
- Clone the repository locally, install dependencies, run Docusaurus locally, edit files, view your changes locally, and commit them.

Usual contributors are probably all using the latest option, but first-time contributors and people submitting PRs on the fly probably don't have their setup done.

Yes, the recommended way of contributing is by forking the repository, but it could be done on the web and directly by following the "Edit" button from the docs.

> Maintainer to fix this properly (before accepting and merge) need to add remote, checkout the PR branch and push changes there, which is very annoying and slows down the review and accept process a lot.

That's why we were exhausted from dealing with that, so we handled it ourselves instead of expecting it from contributors.

Contributors—except top contributors—are not going to look for the style compliance, they want to submit a quick fix for an issue they came through will reading the docs. Most of them will never go back to it to address the reviewer's comments, and collaborators are mostly applying those suggestions to get the pull request merged.

> Prettier can introduce some necessary or breaking changes (#1900 (comment)), so IMO the changes should not be applied automatically.

Sure, but those are rare cases, and the pre-commit hook could apply them automatically anyway.

___

_TL;DR_ What I want to bring here is that enforcing Prettier on pull requests could/will add a lot more stale pull requests that maintainers will fix by hand anyway, and that's annoying.

rachelnabors:
I see both perspectives here, and I have to say that @charpeni's has the Spirit of React Native. 

In an ideal world, I'd love if we could automate Prettier for people from all three use cases and only intervene _if_ there are breaking changes. Otherwise, we risk alienating eager contributors with pedantry ("Hey, you can't contribute here unless you commit to running this thing on your computer first! Take 30 minutes out of your life to fix this typo or go home!") 

Ideas on how that workflow could work?

PR #2026, opened 2020-07-06
replace state usage comparison image with the code blocks

### LINKED ISSUE #2014: Bug in React Native code in example comparing state in React vs React Native
## Description 
In the React Native tutorial's image of example code comparing state in React vs React Native immediately following [this header](https://reactnative.dev/docs/tutorial#there-are-differences-between-react-and-react-native-to-handle-the-state), the React Native code is missing a `.container` class in the creation of the `StyleSheet` object.  

Here's the image copied from https://reactnative.dev/docs/tutorial
image

Here are my edits highlighting the issue. 
image

Apologies if this is not a real bug. I am new to React Native and only noticed the missing `.container` class because the immediately following code example includes it in the `StyleSheet.create()` function along with `.button`.

Simek:
Fixes #2014 

This PR replaces the React and React Native state usage comparison image on the Tutorial page with the two code block arranged in the column (also the problem from the issue above has been fixed). On the small devices blocks are stacked in rows.

Live preview: https://deploy-preview-2026--react-native.netlify.app/docs/next/tutorial#state

### Preview
Annotation 2020-07-06 223146

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -116,6 +116,7 @@ In a React component, the props are the variables that we pass from a parent com
 import React, { useState } from 'react';
 
 
+
 const App = () => {
   const [count, setCount] = useState(0);
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -108,9 +108,71 @@ In a React component, the props are the variables that we pass from a parent com
 
 #### There are differences between React and React Native to handle the state?
 
-![image](https://user-images.githubusercontent.com/20761166/61405629-48270680-a8a8-11e9-906e-aa80d51e51e3.png)
+<div class="two-columns">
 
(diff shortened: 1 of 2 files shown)


gondo:
When is this going to be reflected on the actual website?
Currently https://reactnative.dev/docs/tutorial still shows image with a bug from #2014

Simek:
Hi @gondo, you can find the current docs under https://reactnative.dev/docs/next/tutorial URL. The `next` pages will become default after the next React Native major release (`0.64`), which is not that far away in time. 

It is also possible to port the pages back to the current [versioned docs (`0.63`)](https://github.com/facebook/react-native-website#versioned-docs), but we are currently working on the migration to Docusuaurs V2 and we are trying to reduced the number of updates to make the V2 fork merge conflicts resolution easier (but feel free to prepare the PR if you feel like this should be addressed right away).

PR #2048, opened 2020-07-12
attempt to speed up PRs testing by reducing the number of docs versions

Simek:
In this PR I have added the `versions.json` file overwrite during the CI `test` job (using simple `cat` command). The goal is to speed up the CI testing on commit. Currently the `test` job could take from 35 minutes to the 95 minutes in the worst, observed case which is a very long time for the CI check.

Most of the PR applies only to the `next` and/or the currently active versioned docs so reducing the number of versions looks like a promising optimization and should not affect most of the cases.

I'm creating this PR as draft because it's possible that I have made a mistake or the CI config file will need updates after the PR pipeline run finishes.

Currently the CI version list looks like: `["0.63", "0.62", "0.5"]` so those versions + `next` should be available in the preview.


Simek:
@charpeni Website test/build took only ~2 minutes and whole pipeline not even 5 minutes. I'm happy with the improvement, but I'm not sure about the chosen version range. What do you think? 

Can you also review the changes and confirm that the adjustment will not affect website deployment on merge?

Simek:
It looks like not only the specified version are available in the deploy preview. Maybe it's an effect of the fallback mechanism?

Annotation 2020-07-12 184302

rachelnabors:
Hey hey! My first question is... is there a reason the tests should run _faster_ than they currently do? While few of our PRs are for older versions, if they were breaking PRs, how could we be sure if the tests don't run for those versions? We have in the past made sweeping changes to older PRs, and breaking tests have prevented us from possibly disastrous merges. How would we compensate for that safeguard? And how does that measure against the gain of faster tests?

PR #2052, opened 2020-07-14
Fix missing "in" in intro-react.md

eduardopelitti:
Without the "multiple times" part, the correct phrase would be "You can render this component in multiple places".

This could also be reworked as "You can render this component in multiple places and multiple times without repeating your code...".

### DIFF
--- a/docs/intro-react.md
+++ b/docs/intro-react.md
@@ -229,7 +229,7 @@ export default Cat;
 
 <block class="endBlock devNotes" />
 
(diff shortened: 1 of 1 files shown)


eduardopelitti:
> shera_entrapta
> 
> Nicely caught—thank you :)

Thank you for the great work and the awesome tools you provide and help maintain.

P.S.: And for the Entrapta 😃

PR #2059, opened 2020-07-17
update StatusBar example and docs

Simek:
This PR updates the `StatusBar` component page and introduces following changes:
* rewritten example:
  * simplify code
  * include transition toggler on iOS devices
  * use `SafeAreaView` instead of container styles adjustments
* platform labels
* missing default values added
* missing types added

Also the small CSS adjustments were required because the page contains first occurrence of platform label inside the table and the small fix was required for the labels in the right sidebar.

### Preview
Annotation 2020-07-17 105951


rachelnabors:
Thank you so much for updating this component, @Simek! Merging :)

Issue #1892, opened 2020-04-30
[Suggestion] Proposed content addition to docs/intro-react.md


armandojsgomes:
Hey there! 

In the document [https://github.com/facebook/react-native-website/blob/master/docs/intro-react.md](https://github.com/facebook/react-native-website/blob/master/docs/intro-react.md), I believe it would be interesting to add that functions can only return one element, thus giving a better understanding of the fragment usage (<>). 

I'm starting out with React Native and even following the Fragment documentation link, I had to go to a third party to really understand the purpose of Fragments.

Cheers!

PR #2065, opened 2020-07-17
Added information regarding fragments and components

well-ington:
This PR was created regarding the suggestion #1892 
I've added a note informing the one root element rule for React components and a reference about Fragments.

### DIFF
--- a/docs/intro-react.md
+++ b/docs/intro-react.md
@@ -119,6 +119,8 @@ class Cat extends Component {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Simek:
@well-ington I'm not sure if this additional paragraph will clarify something. There is already similar block with the link to `Fragments` docs at the end of the page.

Annotation 2020-07-19 114438

well-ington:
You're right, it's redundant.
Thanks for your feedback, I'm still learning on how to contribute and address issues 🚀

PR #2075, opened 2020-07-20
Add `expo-secure-store` to a list of libraries to consider for secure storage

sjchmiela:
Noticed this great page and thought `expo-secure-store`, using keychain services on iOS and SharedPreferences and Keystore on Android, is worth mentioning too. 🙂


sjchmiela:
CLA signed! ✅

PR #2079, opened 2020-07-20
Update the support of percentage in documentation of Height And Width file

MianJawadAhmad:
The support of percentage is added two years ago
Update the support of percentage in documentation of Height And Width file

### DIFF
--- a/docs/height-and-width.md
+++ b/docs/height-and-width.md
@@ -7,7 +7,7 @@ A component's height and width determine its size on the screen.
 
 ## Fixed Dimensions
 
-The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are not unitless it support percentage, and represent density-independent pixels.
+The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are not unitless its support percentage and represent density-independent pixels.
 
 ```SnackPlayer name=Height%20and%20Width
 import React from 'react';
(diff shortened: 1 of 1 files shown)


docs/height-and-width.md
```diff
@@ -7,7 +7,7 @@ A component's height and width determine its size on the screen.
 
 ## Fixed Dimensions
 
-The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are unitless, and represent density-independent pixels.
+The general way to set the dimensions of a component is by adding a fixed `width` and `height` to style. All dimensions in React Native are not unitless it supports percentage and represents density-independent pixels.
```
Simek, line 10: From the technical point of view dimensions in React Native **are** unitless - every method related to the dimensions will return an integer (not a string with unit, like for example in CSS) which is a density pixels value. Later in the process many different operations on that value can happen (for example: multiplication by Pixel Ratio, conversion).

For arranging views React Native uses [Yoga](https://github.com/facebook/yoga) layout which besides DP and percentage allow specifying `"auto"` as value (which is also a default). You can find more information about the supported values in the [Yoga documentation](https://yogalayout.com/docs/width-height).

Simek:
@MianJawadAhmad I have created the separate PR to fix the problem in more broad way and because in addition to my comment above I have noticed that percentage values mention should not be placed inside the `Fixed dimensions` section (to avoid potential users confusion).

rachelnabors:
I'm approving @simek's branch and closing this one, but thank you @MianJawadAhmad for bringing this discrepancy to our attention! 

sm_thanks

Issue #2091, opened 2020-07-22
App crashes on the TouchableOpacity Snack Expo example


smakosh:
## Description

![Screen Shot 2020-07-22 at 20 25 02](https://user-images.githubusercontent.com/20082141/88220045-5e556880-cc5a-11ea-95df-ab31e2e82fb3.png)

Simek:
Hi @smakosh, can you provide the link to the page with broken example? I have tested `TouchableOpacity` examples in `0.63` version and in the `next` docs and I have not experience any problems or issues. Also including your OS and browser information could be helpful to reproduce the problem.

smakosh:
Seems to be working now

PR #2092, opened 2020-07-22
docs: add export default App line

smakosh:
To avoid device crashing, `export default App` line had to be added

Broken example link: https://reactnative.dev/docs/0.50/touchableopacity

## Fixes

https://github.com/facebook/react-native-website/issues/2091


smakosh:
Looks like it's an issue from either Snack Expo or when parsing the code, as there is an `export default App` on the example code snippet

PR #2104, opened 2020-07-27
fix Environment Setup page sections, backport all the changes to 0.63

Simek:
Fixes #2019 

This PR fixes the problem described in the issue above and fixes the issue with initial page render (unnecessary sections were visible before the user interacts with tabs).

@rachelnabors Unfortunately I was not able to fix the issue and retain the `devNote` blocks, so the fix requires to removed them. I will work on restoring them in the future. To make the fix available to most of the users I have also backported all the changes from the current page to `0.63` versioned docs.


rachelnabors:
Thanks for nabbing this one! Phew!

PR #2109, opened 2020-07-28
Correct anchor height offset for Black Lives Matter banner

FLGMwt:
|Before|After|
|---|---|
|before|after|

The Black Lives Matter call to action banner added a fixed 100px to the pre-content banner height. Docusaurus seems to have a built-in anchor style to handle this itself, so this just updates that -80 to -180. Mobile was off a bit differently but 160px looked alright there.

EDIT: mobile before and after:

|Before|After|
|---|---|
|before-mobile|after-mobile|

### DIFF
--- a/website/static/css/header.css
+++ b/website/static/css/header.css
@@ -211,6 +211,16 @@ input#search_input_react:focus {
   height: 60px;
 }
 
+.anchor {
+  top: -160px;
+}
+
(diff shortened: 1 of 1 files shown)


FLGMwt:
@Simek 

👍 Done! Put it right after the header container heights since that seemed the most relevant.

PR #2121, opened 2020-08-03
Add background location to dangerous permissions

Sakarisson:
Adds `ACCESS_BACKGROUND_LOCATION` to list of permissions that require prompting the user.


Sakarisson:
CLA signed.

PR #2136, opened 2020-08-11
Adding configureNext.onAnimationDidFail callback on LayoutAnimation

gedeagas:
This PR updates the documentation to reflect the (pending) changes on LayoutAnimation configureNext.onAnimationDidFail callback.

See here for references https://github.com/facebook/react-native/commit/777957c6fba3da642d7f691fd2be44644c66507f


gedeagas:
@rachelnabors @Simek Hi guys, since this callback is fabric only and probably not called from the native side in non-fabric I am curious about the strategy here. how do you guys approach on documenting fabric only API with `noop` for non-fabric?

cpojer (approved):
Thank you!

PR #2137, opened 2020-08-11
adjust BLM banner height and text style

### LINKED ISSUE #2134: [UI/UX] Improve the announcement on smaller screens
## Description

The announcement div, currently used for a Black Lives Matter donation link, is too large, and makes using the docs a huge hassle on smaller screens.

## What is the problem?

On a small laptop screen, the actual docs content ends up being about 1/2 of the entire screen, due to the large announcement, navbar and system elements accumulating. With many pictures in the guides and docs, this makes the user experience on smaller screens poor.

## How can we address it?

We can make two things, either:
1. Add a closing button to the announcement. I guess that wouldn't be too hard, just a small change in CSS visibility or otherwise disabling the div.
2. Make it smaller. Vue's banner on their main page is about 40 pixels high, where it's still quite visible, but doesn't interrupt user's workflow. Even main React page has a smaller banner (60px vs Native's 100px).

## Why is it important?

While the BLM movement is really important, and support should be given for it, it shouldn't interrupt average person's work. I'll still use the docs, but it's really uncomfortable for me to not see the whole picture at the same time.

## Who needs this?

Users with small screens.

Simek:
Fixes #2134

This PR changes the height of BLM banner from `100px` to `60px` and updates the banner text style to ensure that in the most cases text will fit in two lines (previously on some mobile devices text could break to three lines) and additional space can be reduced.

### Preview
**Desktop (Win + 4k)**

**Desktop (MacBook Pro 15)**
Screenshot 2020-08-11 at 21 17 29

**Mobile (iPhone SE)**

### DIFF
--- a/website/static/css/header.css
+++ b/website/static/css/header.css
@@ -204,7 +204,7 @@ input#search_input_react:focus {
 }
 
 .headerWrapper.wrapper {
-  padding-top: 100px;
+  padding-top: 60px;
 }
 
 .navigationSlider .slidingNav {
@@ -225,9 +225,9 @@ input#search_input_react:focus {
   color: #fff;
   font-weight: bold;
   font-size: 24px;
-  padding: 30px;
+  padding: 8px 30px;
   text-align: center;
-  height: 100px;
+  height: 60px;
   position: absolute;
   width: 100%;
 }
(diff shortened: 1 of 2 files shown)


Wuzado:
Thank you very much!

PR #2149, opened 2020-08-15
Adding notes regarding the issue when Android timers are not in sync with the debugger

gedeagas:
This PR adds notes regarding the issue when Android timers are not in sync with the debugger. 

See here for the discussion https://github.com/facebook/react-native/issues/29591.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -57,7 +57,7 @@ To debug the JavaScript code in Chrome, select "Debug JS Remotely" from the Deve
 
 Select `Tools → Developer Tools` from the Chrome Menu to open the [Developer Tools](https://developer.chrome.com/devtools). You may also access the DevTools using keyboard shortcuts (`⌘⌥I` on macOS, `Ctrl` `Shift` `I` on Windows). You may also want to enable [Pause On Caught Exceptions](http://stackoverflow.com/questions/2233339/javascript-is-there-a-way-to-get-chrome-to-break-on-all-errors/17324511#17324511) for a better debugging experience.
 
-> Note: on Android, if the times between the debugger and device have drifted; things such as animation, event behavior, etc., might not work properly or the results may not be accurate. Please correct this by running `` adb shell "date `date +%m%d%H%M%Y.%S`" `` on your debugger machine. Root access is required for the use in real device.
+> Note: on Android, if the times between the debugger and device have drifted; things such as animation, event behavior, etc., might not work properly or the results may not be accurate. Please correct this by running `` adb shell "date `date +%m%d%H%M%Y.%S%3N`" `` on your debugger machine. Root access is required for the use in real device.
 
 > Note: the React Developer Tools Chrome extension does not work with React Native, but you can use its standalone version instead. Read [this section](debugging.md#react-developer-tools) to learn how.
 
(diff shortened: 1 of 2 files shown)


docs/debugging.md
```diff
@@ -57,6 +57,8 @@ To debug the JavaScript code in Chrome, select "Debug JS Remotely" from the Deve
 
 Select `Tools → Developer Tools` from the Chrome Menu to open the [Developer Tools](https://developer.chrome.com/devtools). You may also access the DevTools using keyboard shortcuts (`⌘⌥I` on macOS, `Ctrl` `Shift` `I` on Windows). You may also want to enable [Pause On Caught Exceptions](http://stackoverflow.com/questions/2233339/javascript-is-there-a-way-to-get-chrome-to-break-on-all-errors/17324511#17324511) for a better debugging experience.
 
+> Note: on Android, if the times between the debugger and device have drifted by more than 60s things might not work or results may not be accurate. Please correct this by running `` adb shell "date `date +%m%d%H%M%Y.%S`" `` on your debugger machine.
```
safaiyeh, line 60: We should also mention @AndrewMorsillo case where animations got messed up even with a slight difference.

Something along the lines of, if you are experiencing unexpected animation or event behavior correct the time.

gedeagas: @safaiyeh Working on it. And I might change the wording on `more than 60s` because everything that relies on Timings will be broken if the time is out of sync. what do you think?

safaiyeh: Sounds good!

AndrewMorsillo: Another thing to note, I think `adb shelll date` will only work if you have root access to the device in question. Without mentioning that many people could be confused. Also I’m not sure it’s worth mentioning “60s” since the threshold for problems seems to be closer to 1-2 seconds.

gedeagas: I also think '60s is the worst-case scenario, and `adb shell date` works well on an emulator but you are correct that when using it on a real device you need to have root access. I will add this to notes section

docs/debugging.md
```diff
@@ -57,6 +57,8 @@ To debug the JavaScript code in Chrome, select "Debug JS Remotely" from the Deve
 
 Select `Tools → Developer Tools` from the Chrome Menu to open the [Developer Tools](https://developer.chrome.com/devtools). You may also access the DevTools using keyboard shortcuts (`⌘⌥I` on macOS, `Ctrl` `Shift` `I` on Windows). You may also want to enable [Pause On Caught Exceptions](http://stackoverflow.com/questions/2233339/javascript-is-there-a-way-to-get-chrome-to-break-on-all-errors/17324511#17324511) for a better debugging experience.
 
+> Note: on Android, if the times between the debugger and device have drifted; things such as animation, event behavior, etc., might not work properly or the results may not be accurate. Please correct this by running `` adb shell "date `date +%m%d%H%M%Y.%S`" `` on your debugger machine. Root access is required for the use in real device.
```
gedeagas, line 60: @safaiyeh updated. kindly check

gedeagas: cc @AndrewMorsillo

AndrewMorsillo: Maybe change the command to `` adb shell "date `date +%m%d%H%M%Y.%S%3N`" `` (aadd %3N to the end) to also sync milliseconds. Without that time can still be off by up to 999ms which can cause issues and performance problems.

safaiyeh: Thanks for the change @gedeagas, +1 to @AndrewMorsillo suggestion

gedeagas: @AndrewMorsillo @safaiyeh updated.

gedeagas: I am also updating the adb command on react-native emitTimeDriftWarning 

Related PR. https://github.com/facebook/react-native/pull/29673/

Simek:
Hi @gedeagas & @safaiyeh, thank you so much for working on this update! 

Please let me know when the PR will be ready for the review! 🙂

safaiyeh:
@Simek it should be ready to go!

Simek (approved):
I felt like this note is not that important (like deprecation note, for example) to be almost at the top of the file, so I have moved it at the bottom of "Timers" section. 

Other than that changes LGTM, that was a great collaboration between you to deliver even more quality, thank you both so much again!

PR #2151, opened 2020-08-17
Fixed Linking.getInitialURL and addEventListener

scarlac:
getInitialURL does not need a parameter. It turns the URL. addEventListener needs a 'url' parameter, however it's simply the string 'url' which wasn't clear.

Also, it may return `null` when debugger is enabled.

### DIFF
--- a/docs/linking.md
+++ b/docs/linking.md
@@ -106,13 +106,13 @@ If your app is using [Universal Links](https://developer.apple.com/library/prere
 
 There are two ways to handle URLs that open your app.
 
-#### 1. If the app is already open, the app is foregrounded and a Linking event is fired
+#### 1. If the app is already open, the app is foregrounded and a Linking 'url' event is fired
 
-You can handle these events with `Linking.addEventListener(url, callback)`.
+You can handle these events with `Linking.addEventListener('url', callback)` -- it calls `callback({ url })` with the linked URL
 
 #### 2. If the app is not already open, it is opened and the url is passed in as the initialURL
 
-You can handle these events with `Linking.getInitialURL(url)` -- it returns a Promise that resolves to the url, if there is one.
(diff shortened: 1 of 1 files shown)


docs/linking.md
```diff
@@ -106,13 +106,13 @@ If your app is using [Universal Links](https://developer.apple.com/library/prere
 
 There are two ways to handle URLs that open your app.
 
-#### 1. If the app is already open, the app is foregrounded and a Linking event is fired
+#### 1. If the app is already open, the app is foregrounded and a Linking 'url' event is fired
 
-You can handle these events with `Linking.addEventListener(url, callback)`.
+You can handle these events with `Linking.addEventListener('url', callback)` -- it calls `callback({ url })` with the linked URL
```
Simek, line 111: From the code it looks like the callback returns just the `url` string not wrapped with the object as added comment suggest: 
* https://github.com/facebook/react-native/blob/master/RNTester/js/RNTesterApp.ios.js#L163
* https://github.com/facebook/react-native/blob/master/RNTester/js/utils/URIActionMap.js#L35

scarlac: Are you sure RNTester is being used and tested against?

I have working code that looks like this:
```
Linking.addEventListener('url', ({ url }) => {
  this.props.updateDeepLink(url);
});
```

scarlac: Here's a more terse example from working code, RN 0.61.2:
```js
Linking.addEventListener('url', event => {
  console.log('event', Object.keys(event));
});
// outputs:
// LOG  event ["url"]
```

Simek: Thank you for checking on that! I was able to confirm using latest version of RN that indeed Linking listener returns an event with only one key which is `url`. The RNTester is used and updated, but as this case shows, not every thing is 100% spot on. The work on the RNTesters lately is a bit on hold due to refactor project from MLH team.

PR #2153, opened 2020-08-18
Update getting-started.md

MohamadKh75:
As [node-gyp](https://github.com/nodejs/node-gyp) supports Python 3, we can remove version 2 from docs.

cc @cpojer

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -135,7 +135,7 @@ You will need Node, the React Native command line interface, a JDK, and Android 
 
 <h2>Installing dependencies</h2>
 
-You will need Node, the React Native command line interface, Python2, a JDK, and Android Studio.
+You will need Node, the React Native command line interface, Python, a JDK, and Android Studio.
 
 <block class="native mac windows linux android" />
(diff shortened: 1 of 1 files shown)


Simek:
Refs #2096 

Hi @MohamadKh75, since this guide has been updated lately please checkout the review comments and discussion about Python version in the PR linked above. If you sure that there are no other libraries or core parts that depends on Python 2, please test the clean setup on both macOS and Windows and provide the proof that it worked. We must be sure that this bump will not introduce any regression or problem for the users.

MohamadKh75:
Hi @Simek

I just can confirm about android.

I asked @cpojer about this but it seems in your ref that it will cause problems for Windows users.

Maybe it should be good to separate sections for mobile and desktop :thinking:

sunnylqm:
I think python is now totally unnecessary for windows platform. Just tested on a brand new windows 10 machine with only node installed, initiating is all good.

Simek:
Closing in favor of #2250, thank you for the research. 🙂

PR #2157, opened 2020-08-19
move image assets from docs to statics folder

Simek:
This PR moves the image `assets` directory from `docs` to the `static` folder. This change was recommend to improve the Docusaurus V2 migration process. 

CC @slorber


darshkpatel:
@Simek I believe the blog assets need to be moved too. Some of them weirdly exist in `static/img/blog`  while the others are in `blog/assets`

slorber:
Thanks @Simek 

I agree that blog assets should be moved too, at least if we want such links to work:

```
![Screenshot of LogBox](/blog/assets/0.63-logbox.png)
```

We need them to exist at `./static/blog/assets/0.63-logbox.png`

Or it would require to rewrite blog assets to `Screenshot of LogBox` (colocation does work, this is not a bad idea either as it makes the images work on Github interface too, but it's not supported by v1 afaik so can be done later).

Simek:
@slorber Oh, you mean that we need to move `blog/asset` to `static`, not `static/img/blog` to `static/assets`. That make sense, sorry for my misunderstanding, I will update this PR 🙂

darshkpatel:
I believe this script needs to be updated too 
https://github.com/facebook/react-native-website/blob/3fd1ff6bbd3fee60354988e5b9a13b0c632b801a/website/image-check.js#L51

rachelnabors:
Nice! Thank you for handling this!

PR #2160, opened 2020-08-20
Document FlatList's 'keyboardShouldPersistTaps'

lekterable:
Documentation doesn't mention that `keyboardShouldPersistTaps` property exists not only in the `ScrollView` but also in the `FlatList` which is very confusing as this property can be very useful.

Not sure if there are also other components that have it if so, let me know :)

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -379,6 +379,22 @@ Reverses the direction of scroll. Uses scale transforms of `-1`.
 
 ---
 
+### `keyboardShouldPersistTaps`
+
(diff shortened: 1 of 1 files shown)


Simek:
Hi @lekterable, `FlatList` do not always have this property available. It's the `ScrollList` prop, which is inherited by `FlatList` in the certain situations and there is a message about that fact at the top of `FlatList` reference:
* https://reactnative.dev/docs/flatlist#props
Annotation 2020-08-20 111650

PR #2165, opened 2020-08-20
remove unnecessary argument of render()

yenshan:
Appreciate to the good document.
I think 'props' in render() is unnecessary. This will cause confusion to the biginner like me.


Simek (approved):
Thank you for pointing this out and for the fix! 👍

PR #2173, opened 2020-08-25
ensure React Native's version on Hermes doc

developerantoniosousa:
I have ensured React Native's version on Hermes doc.

### DIFF
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -9,7 +9,7 @@ title: Using Hermes
 
 [Hermes](https://hermesengine.dev) is an open-source JavaScript engine optimized for running React Native apps on Android. For many apps, enabling Hermes will result in improved start-up time, decreased memory usage, and smaller app size. At this time Hermes is an **opt-in** React Native feature, and this guide explains how to enable it.
 
-First, ensure you're using at least version 0.60.4 of React Native.
+First, ensure you're using at least version 0.60.3 of React Native.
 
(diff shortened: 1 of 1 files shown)


developerantoniosousa:
I have singed the Contributor License Agreement after receive the error by react-native-bot

Simek:
Hi @developerantoniosousa, can you provide the context (or the reference commit) for the bump? Are there Hermes issues when working with RN `0.60.3`?

developerantoniosousa:
@Simek 
I haven't seen any issues using Hermes with RN 0.60.3

I have created this PR, because the RN 0.64 wasn't released yet, I think to have it on doc is a problem, isn't it?

Simek:
This sentence states that minimim version of RN that works with Hermes is 0.60.3. Sorry, but I don't understand the reason to change it to the upcomming (not released yet) version.

developerantoniosousa:
OK, It's because this PR would be merged into master, bound with RN 0.60.4. My intention was ensure the version for docs to RN 0.60.3, because there it's 0.60.4. I imagined would be wrong, because the version of doc is RN's 0.60.3. How can I create a PR to the docs 0.60.3 of RN?

Simek:
@developerantoniosousa I'm still not sure what you would like to achieve by this change. The RN changelog shows the reason why `0.60.4` is recommended instead of `0.60.3` as a minimum RN version to work correctly with Hermes: 
* https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0604

> I have created this PR, because the RN 0.64 wasn't released yet, I think to have it on doc is a problem, isn't it?

And I'm still thinking that have you confused minor with major version, `0.60.4` is way older version in comparison to the `0.64.0` which is not released yet.

developerantoniosousa:
I undestood you point now, I have made a comment before where I spoke wrong about version. But my intention was only know if that information (First, ensure you're using at least version 0.60.4 of React Native) could be wrong on docs.

I have thought that would be wrong because the 0.60.4 pointed by docs now (when the stable release is 0.60.3) could be something wrong, because that version isn't released yet.

I have access this part over  [here](https://reactnative.dev/docs/0.60/hermes and saw that version (0.60.4) also appears in older docs' versions too.

PR #1845, opened 2020-04-13
fix(RefreshControl): fix SnackPlayer

AntoineDoubovetzky:
Hi,

I just saw a warning when I opened the snack example, so I fixed it :)

PR #2180, opened 2020-08-28
Edit useCallback hook

mojvan:
Hello..
I think an updated parameter is needed for useCallback hook otherwise we don't need this hook....
Good luck...

### DIFF
--- a/docs/refreshcontrol.md
+++ b/docs/refreshcontrol.md
@@ -17,11 +17,10 @@ const wait = (timeout) => {
 
 const App = () => {
   const [refreshing, setRefreshing] = React.useState(false);
-
   const onRefresh = React.useCallback(() => {
     setRefreshing(true);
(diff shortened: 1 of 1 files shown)


mojvan (commented):
@Simec , I'm waiting for your answer

Simek:
Refs #1845

Hi @mojvan, thank you for proposing those changes, but in this specific case `useMemo` would be enough and generally you need to remember it's just an example. Using `useCallback` here is more coherent with the general usage pattern in "real" use cases (when some async operations will be executed). Also change to the `setRefreshing` value in the changeset cause an incorrect behavior.

mojvan:
@Simek , do you can show me this incorrect behavior?
Thank you

mojvan (commented):
@Simek , can you show me the incorrect behavior?
Thank you

PR #2187, opened 2020-08-31
Edit TextInput example to compare PRs

mojvan:
I added SafeAreaView component because it is necessary in this case and wrote two TextInput components so that the reader would know the difference between value prop and placeholder prop and also understand the use of keyboardType.

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -9,17 +9,25 @@ The most basic use case is to plop down a `TextInput` and subscribe to the `onCh
 
 ```SnackPlayer name=TextInput
 import React, { Component } from 'react';
-import { TextInput } from 'react-native';
+import { TextInput , SafeAreaView} from 'react-native';
 
 const UselessTextInput = () => {
   const [value, onChangeText] = React.useState('Useless Placeholder');
-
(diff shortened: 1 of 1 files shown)


Simek:
Hi @mojvan, thank you for tries to improve the examples, but there were few many minor issues and changes that I would like to see to see in this example and the code has not been run through Prettier, but you have disabled ability to commit to your fork, so I was not able to apply review changes in this PR. 

Please remember next time to set the correct setting of your fork and that you checked the `Allow edits by maintainers` checkbox in the PR, this speeds up the review process, thank you for that in advance! 👍

Simek:
@mojvan I'm sorry if you feel bad about closure, but the settings were correct when you have proposed few earlier PRs, so you have to change the permissions manually. If you would leave the default settings, you would retain the authorship to the changes.

Please take to the consideration that many of maintainers devote their free time to work on the project. In past we have a lot of stale PRs, that required the code change from the author after the review (which also takes time, usually even more than correcting the issues in the code), so editing the PRs became one of the ways to improve the workflow. This is nothing personal, hope you can understand that. 🙂

mojvan:
I'm so sorry

PR #2188, opened 2020-08-31
Modify additional codes

mojvan:


Simek:
Hi @mojvan, thank you for your proposition and time put in, but unfortunately modifying this example in this way do not add any functionality and makes the `title` prop redundant (`index` can be used instead). Also removing the extracted `DATA` and storing `index` as `title` reduce the clarity of the example. 

If you are keen on helping with the website please focus on the reported and confirmed issues first.

PR #2189, opened 2020-08-31
Modernize code examples in the direct manipulation section

steveluscher:
The examples in the Direct Manipulation section look to have fallen victim to a rogue codemod; they are currently a mix of functional and class syntax that doesn't actually work.

In this PR I do too many things (sorry):

* fix indentation (2 spaces everywhere)
* remove styles that don't aid in the understanding of the concept of direct manipulation
* convert everything from class to functional/hooks syntax
* repair the opacity example by dimming the button `onPressIn` instead of `onPress`
* use conventional ref forwarding to achieve the desired effect in the example where we show how to forward the effect of `setNativeProps` from a wrapper component to a child component

T74535407

### DIFF
--- a/docs/direct-manipulation.md
+++ b/docs/direct-manipulation.md
@@ -39,12 +39,8 @@ Let's imagine that `setNativeProps` was not available. One way that we might imp
 const [buttonOpacity, setButtonOpacity] = useState(1);
 return (
   <TouchableOpacity
-    onPressIn={() => {
-      setButtonOpacity(0.5);
-    }}
-    onPressOut={() => {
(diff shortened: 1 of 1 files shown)


Simek (approved):
@steveluscher thank you so much for improving those examples and updating the related sentences, great work! Also the changes description is very informative, thank you for that too! 🙂 

Review comment: I have added missing `useRef` import and pass the examples code through the Prettier (currently JSX in MD files is not correctly formatted when running `yarn prettier` script from this repository).

steveluscher:
Thanks for the catch and the cleanup! 🎉

PR #2191, opened 2020-09-01
update SectionList page, extract ViewToken, add labels

Simek:
Source for the changes:
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/SectionList.js
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/ViewabilityHelper.js
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedSectionList.js

This PR introduces the following changes and it is the initial step on working on the `List` components (some code, styling or formatting might change in the next few PRs related to other lists):
* extracts `ViewToken` object type to the separate page
* updates the `SentionList` props and description
* adds require and platform labels
* updates props order to be alphabetic

@rachelnabors Feel free to change or rephrase the descriptions, they can be better for sure. I had to write them from the scratch because there are no comments in the code currently related to the `ViewToken` structure.

#### SectionList
Annotation 2020-09-01 162419

#### ViewToken
Annotation 2020-09-01 162013

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -447,18 +447,11 @@ If provided, a standard RefreshControl will be added for "Pull to Refresh" funct
 
 ### `onViewableItemsChanged`
 
-```jsx
-(info: {
-    viewableItems: array,
-    changed: array,
-  }) => void
-```
-
 Called when the viewability of rows changes, as defined by the `viewabilityConfig` prop.
 
-| Type     | Required |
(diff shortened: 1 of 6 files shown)


rachelnabors (approved):
Much better than before :)

PR #2194, opened 2020-09-02
Update getting-started.md

prkhandelwal:
Python2 & Python3 are both compatible


rachelnabors (approved):
Thank you for the update!

Issue #1899, opened 2020-05-04
transform matrix documentation.


vaukalak:
## Description

[Documentation](https://reactnative.dev/docs/transforms#transform) doesn't mention the `matrix` transform property, event if it's supported in react-native.

## What is the problem?

I see that `decomposedMatrix` and `transformMatrix` are marked as deprecated in the doc. However following syntax is not mentioned at all:
```
<View style={{ transform: [{ matrix: some4x4Matrix }] }} />
```

So I'm not able to understand is this syntax deprecated / experimental / cross-platform ?

## How can we address it?

If the `matrix` transform is deprecated, it would be great see that mentioned explicitly.
If it's currently supported, and not intended to be deprecated, it would be great to mention that explicitly here. IMHO even if it would be just listed as `{ matrix: CATransform3D }` maybe with link to github file location, it would be a big improbement.

## Why is it important?

Creating UI editor apps without using matrix is very hard. When you build a matrix inside your state, it would be great to just pass it to the UI component.

## Who needs this?

Everyone who works with matrix transformations.

## When should this happen (use version numbers if needed)?

Well, sooner is better :)

Simek:
Indeed using `'Use `transform: [{ matrix: ... }]` instead.` is recommended in the [deprecation warning](https://github.com/facebook/react-native/blob/master/Libraries/DeprecatedPropTypes/DeprecatedTransformPropTypes.js#L25), but there is no `matrix` example on the page. 

I was only able to find few examples of `matrix` usage in the internal [tests](https://github.com/facebook/react-native/blob/master/Libraries/Utilities/__tests__/buildStyleInterpolator-test.js#L268). 

It's quite interesting that there is also a [TODO note](https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheetTypes.js#L512) according to the `matrix` in the types definition file.

ProBeta1:
Shall I work on this @Simek ?

Simek:
Hi @ProBeta1, feel free to work on this issue. Like an author I'm curious too about the `matrix` structure.

vaukalak:
At least my tests shown that when we mount a view with a `matrix` transform, the only way to update that matrix is via `setNativeProps`. Also it would be good to document, that this is a 4x4 matrix (since rn supports rotateX/Y). I'm also unsure about the `MatrixMath` module documentation, it's not exposed as a top level thing.

AbhishekGiri04:
[...]

PR #2197, opened 2020-09-03
docs: add matrix to transform

Naturalclar:
Added matrix to transform props, as matrix is available in the transform property for stylesheet

ref: https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/processTransform.js#L48-L50


Simek (approved):
Thank you for the type update! 👍 Refs #1899

PR #2198, opened 2020-09-03
add guide for FlatList onViewableItemsChanged 

sueLan:
I wrote an article about the `onViewableItemsChanged` in `FlatList`.  [The article is here](https://suelan.github.io/2020/01/21/onViewableItemsChanged/)  @rachelnabors was advising me to add this article to this website. [The conversation is here](https://www.reddit.com/r/reactnative/comments/f15u44/dive_into_onviewableitemschanged_function_in/fz2y1io?utm_source=share&utm_medium=web2x&context=3).  I think it would be better for more people to better understand this API and the implementation behind it. Because the implementation is quite interesting.


nearestnabors (changes requested):
Hi! Thanks so much for contributing! I notice this is a a lot like the article you posted if not a perfect copy. Before it becomes a part of the docs, it will need a little modification:

1. Make sure you're adhering to our [styleguide](https://github.com/facebook/react-native-website/blob/master/STYLEGUIDE.md) as best you can. For instance, you call the reader "we" but the documentation always calls the reader "you" ("we" refers to the people building React Native/the person writing the docs!).

2. It would be great to open with a section on on _why" this is important.

After addressing these, it makes sense to move this out of the blog and under our Guides. We can figure out where it lives then :)

Thanks so much again!

rachelnabors:
It's usually good to open with a small paragraph answering "how, who, what, why, where, and when":

- What are we going to talk about?
- Who would need to do this?
- Why is it important?
- Where would you implement this?
- When would you need to know this?
- How do we do this? (Read below to find out!)

You can see some of that [here](https://reactnative.dev/docs/intro-react-native-components)

website/versionedDocsBlacklist.json
```diff
@@ -44,5 +44,6 @@
   "headless-js-android",
   "signed-apk-android",
   "removing-default-permissions",
-  "optimizing-flatlist-configuration"
+  "optimizing-flatlist-configuration",
+  "onviewableitemschanged-in-flatlist"
```
Simek, line 48: The page id in blacklist is incorrect.

sueLan: Thank you for this info

sueLan: I did some changes. But the article hasn't shown up in the `Basics` section
1. move article into the `docs`, and add `${id}` , `flatlist-on-viewable-item-changed` 
2. add `${id}` in `website/sidebars` 
3. fix the id in `versionedDocsBlacklist.json`. 
save changes, then, I run `yarn start` in the `react-native-website/website` , open `http://localhost:3000/`. But I can't see the article in the `docs-> The basics`

sueLan: @Simek would you shed some light on this?

Simek: @sueLan You have only updated the `next`/`master` docs (https://reactnative.dev/versions), to see the changes in the current docs you need to edit `versioned_sidebar` and add the page to `versioned_docs` for the `0.63` release.

sueLan:
> It's usually good to open with a small paragraph answering "how, who, what, why, where, and when":
> 
> * What are we going to talk about?
> * Who would need to do this?
> * Why is it important?
> * Where would you implement this?
> * When would you need to know this?
> * How do we do this? (Read below to find out!)
> 
> You can see some of that [here](https://reactnative.dev/docs/intro-react-native-components)

Hi, thank you for the details. I just added a small section at the beginning of article to explain why, when, who.  Besides, I also made other changes as you mentioned.  I don't know what specific folder I should put this article.  I put it in the `The basics` folder at first. Maybe, you have some idea?

website/versionedDocsBlacklist.json
Simek: Hmm, maybe when using `versionedDocsBlacklist` only the sidebar change is required. AFAIK blacklist prevents from generating static, versioned pages on new version cut. I'm not sure how it will behave in this situation.

If you will be still struggling with page visibility I will look on this PR locally during the weekend. 🙂

sueLan: ha Thank you ~, I did as you told me, and saw the page.  Besides, I updated a picture.

rachelnabors:
## Editing inside a google doc?

This is really good! @sueLan, normally I work with authors in a Google Doc or some place where we can freely rearrange and expand on the guide's structure. Text editors are... less collaborative! If you'd like to move the content discussion of this PR into a Google Doc, it might make this faster! I made a few edits, but I think it might go faster to make the rest if we use a collaborative space like Google Docs.

## Structural changes other docs need

** Viewablity config has a better explanation under flatlist https://reactnative.dev/docs/flatlist#viewabilityconfig than virtualizedlist https://reactnative.dev/docs/virtualizedlist#viewabilityconfig**

* @Simek Can we extract these to an Object Type or shared props list?
* If not, can we make sure the two are the same?
* These sections should also link to this new doc, and that should be part of this PR.

## More feedbacks

* This doc should come after [Using a ListView](https://reactnative.dev/docs/using-a-listview) in the sidebar.
* Your examples need to go into a Snack player! Check out [this page](https://reactnative.dev/docs/intro-react)'s [source code](https://github.com/facebook/react-native-website/blob/master/docs/intro-react.md) to see how it works.

## Example Structure

I removed the section where you share the types with the audience (copied below) because while that information is important, but it can be found at the link to `ViewabilityConfig`. I think it would work better if you broke down the examples to explain how these values are working in them. Example structure:

* Snack player with example code
* "Let's take a look at what is happening here:"
* <code snippet>
* explanation ("`minimumViewTime` denote the milliseconds an item must be visible before the callback will be fired...")
* more code snippets and their explanations

What do you think?

sueLan:
> Google Doc 
Fine. I need you. Should I create the Google doc or you create it in your google drive? 
> This doc should come after Using a ListView in the sidebar.
ic, I will move it there 
> Snack player
will do this. 
> removed the section `ViewabilityConfig`
nice!  It is better to use link. Looks more concise

Simek:
> @Simek Can we extract these to an Object Type or shared props list?

Sure, I can work on that. This property is used by two components so it is enough to justify the extraction to the Object Type. 🙂

rachelnabors:
@Simek Perfect :)

@sueLan I'll start a collaboration doc. Check your inbox!

rachelnabors:
Also thinking it actually belongs under "Interactions"—it's far too advanced for the Basics section :D

sueLan:
> Also thinking it actually belongs under "Interactions"—it's far too advanced for the Basics section :D
 Agree that it isn't something basic and doesn't belong to Basics Section.  Initially, I think `interactions` is mainly for articles about touch event,  gesture responder. But now, I feel like it makes sense to put this article there.  Because `onViewableItemsChanged` mainly works in scenarios where users scroll the `FlatList`.

PR #2203, opened 2020-09-07
Added workarounds for TextInputs

EyMaddis:
This adds a bit of info for other people struggling with Android TV.

### DIFF
--- a/docs/building-for-tv.md
+++ b/docs/building-for-tv.md
@@ -155,4 +155,7 @@ class Game2048 extends React.Component {
 
 - _Known issues_:
 
-  - `TextInput` components do not work for now (i.e. they cannot receive focus, see [this comment](https://github.com/facebook/react-native/pull/16500#issuecomment-629285638)).
+  - `TextInput` components do not work for now (i.e. they cannot receive focus automatically, see [this comment](https://github.com/facebook/react-native/pull/16500#issuecomment-629285638)). 
+    - It is however possible to use a ref to manually trigger `inputRef.current.focus()`. 
+    - You can wrap your input inside a TouchableWithoutFeedback component and trigger focus in the `onFocus` event of that touchable. This enables opening the keyboard via the arrow keys.
+    - The keyboard might reset its state after each keypress. This might only happen inside the Android TV emulator.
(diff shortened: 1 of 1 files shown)


gedeagas (commented):
Thx for the awesome pr @EyMaddis, can you also add this one https://github.com/facebook/react-native/issues/29774? This issue is about Scrollview still scrollable with dpad when scrollEnabled={false}

EyMaddis:
@gedeagas If the fix is known (and the PR seems to be rather certain...), I do not think that we need to include it.

gedeagas:
It would sometimes until the fix is released on react-native, and given the fix require you to have a custom version of react-native with the workaround applied I think adding this note to the known issue section will help users.

gedeagas:
@Simek what do you think?

Simek:
@EyMaddis Thank you for adding the useful tips for TV builds! 👍 

@gedeagas It looks like the commit you have mentioned in the DPad scroll issue has bee a part of `0.63.2` release, so in theory, the fix is out (https://github.com/facebook/react-native/commit/00c8b3cd09526c4a8afd141344793151ccb9ea0c). If I'm wrong or if I have missed something I will add this known issue in the separate PR, but anyway thank you for the engagement, that's awesome!

gedeagas:
@Simek the commit is not working

Simek:
> @Simek the commit is not working hence there is a new issue report

Can you link the new issue? Are you sure that reporter was using at least `0.63.2`?

gedeagas:
@Simek it's the same issue here https://github.com/facebook/react-native/issues/29774 The issue here is that after testing it this commit https://github.com/facebook/react-native/commit/00c8b3cd09526c4a8afd141344793151ccb9ea0c does not work. PR is already on the way here https://github.com/facebook/react-native/pull/29906.

Simek:
@gedeagas The reporter is using `0.61.5` and is fixing `0.63-stable` not `master`, so honestly IDK what's the future of this fix will be. I'm going to merge this PR, but please free to prepare another one with facebook/react-native#29774 note addition, I'm open to merge that too, this always can be removed when the issue will be fixed. 🙂

Simek (approved):
Thank you for sharing those tips with the community! 👍

PR #2209, opened 2020-09-10
update Text page and add types, add LayoutEvent object page

Simek:
Source:
* https://github.com/facebook/react-native/blob/master/Libraries/Text/Text.js
* https://github.com/facebook/react-native/blob/master/Libraries/Text/TextProps.js
* https://github.com/facebook/react-native/blob/master/Libraries/Types/CoreEventTypes.js

This PR updated the Text page and introduces the following changes:
* add unique for Text component `TextLayout` and `TextLayoutEvent` objects definitions
* update examples to avoid `console.log` usage
* utilize platform labels
* required/optional related changes
* update `PressEvent` related props

Withing this changes also I have extracted the `LayoutEvent` to the separate page (located in Object Types). It is used by multiple components and those components pages also have been updated. 

@rachelnabors I have also decided not split this page into `LayoutEvent` and `Layout` as RN code treat this object just for the sake of UX/readability. Let me know if it is a good decision. Also the type object properties descriptions may need to be updated, because there was no comments in the core code.

### Preview
Screenshot 2020-09-10 183555

Screenshot 2020-09-10 185552

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -341,11 +341,11 @@ Invoked on load error with `{nativeEvent: {error}}`.
 
 ### `onLayout`
 
-Invoked on mount and layout changes with `{nativeEvent: {layout: {x, y, width, height}}}`.
+Invoked on mount and on layout changes.
 
-| Type     | Required |
-| -------- | -------- |
-| function | No       |
+| Type                                 | Required |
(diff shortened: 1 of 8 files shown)


rachelnabors (approved):
sm_thanks
Looks good!

PR #2212, opened 2020-09-15
Update inline requires doc for upcoming API change

motiz88:
Updates the guide on setting up inline requires in accordance with https://github.com/facebook/metro/pull/591, which is a Metro PR that **has not landed yet**. When it does land and gets included in a React Native release, this doc change should accompany it.

### DIFF
--- a/docs/ram-bundles-inline-requires.md
+++ b/docs/ram-bundles-inline-requires.md
@@ -169,7 +169,7 @@ const config = {
       });
       return {
         preloadedModules: moduleMap,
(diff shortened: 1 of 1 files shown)


Simek (approved):
@motiz88 Thank you for the update! 👍 

Since upgrade to [Metro 0.64](https://github.com/facebook/metro/releases/tag/v0.64.0) has landed in the [`0.64-stable` branch](https://github.com/facebook/react-native/commits/0.64-stable) this PR can be merged now.

PR #2215, opened 2020-09-17
Docs: Remove deprecated method `didReceiveLocalNotification` from Push Notification iOS

carloscuesta:
## Description

This is related with the following issue: https://github.com/facebook/react-native/issues/26199

According to [iOS docs](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622930-application?language=objc), `didReceiveLocalNotification` is deprecated and should be replaced by 

> Screenshot 2020-09-17 at 17 09 57

This PR removes the following method from the PushNotificationIOS documentation

### DIFF
--- a/docs/pushnotificationios.md
+++ b/docs/pushnotificationios.md
@@ -69,11 +69,6 @@ And then in your AppDelegate implementation add the following:
  {
   [RCTPushNotificationManager didFailToRegisterForRemoteNotificationsWithError:error];
  }
- // Required for the localNotification event.
- - (void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification
- {
-  [RCTPushNotificationManager didReceiveLocalNotification:notification];
- }
 ```
 
 To show notifications while being in the foreground (available starting from iOS 10) add the following lines:


Simek (commented):
Hi @carloscuesta, unfortunately the method [`didReceiveLocalNotification`](https://github.com/facebook/react-native/blob/master/Libraries/PushNotificationIOS/RCTPushNotificationManager.mm#L215) is still part of the React Native source code, so it should remain documented. 

If it's an important change for you, you can try to remove the `didReceiveLocalNotification` from the core and then repost this PR.

Also `PushNotificationIOS` has been deprecated a while ago and the recommended package should be used instead.

carloscuesta:
No problem @Simek thanks for  taking the time on reviewing the PR! Even though PushNotificationIOS is deprecated I considered that it was a valuable thing to update, that's why I opened it! 😊

Simek:
@carloscuesta Thank you for the effort! This change is still welcome, but you need to update the React Native code first. 🙂 

But this might be a bit problematic, since the whole library is scheduled for the removal from the core.

PR #2217, opened 2020-09-18
Docs: Integration with android fragment

svnm:
@seba1342 and I have added documentation for how to render a React Native component into a Fragment on Android. We are hoping to set up equivalent docs for a similar approach for iOS adding a component to a UIViewController.

We have set up a working example repo for this here: https://github.com/svnm/react-native-android-fragment-example

There was no current documentation on how to achieve this so we thought it would be best to add a new page to the docs. We used the approach set up by hudl https://github.com/hudl/react-native-android-fragment which has been merged in to react native now as [ReactFragment.java](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/ReactFragment.java)

We found some additional discussion and questions on this approach from the following stack overflow posts: https://stackoverflow.com/a/62633927, https://stackoverflow.com/questions/43796433

### DIFF
--- a/docs/integration-with-android-fragment.md
+++ b/docs/integration-with-android-fragment.md
@@ -1,0 +1,68 @@
+---
+id: integration-with-existing-apps
+title: Integration with an Android Fragment
+---
+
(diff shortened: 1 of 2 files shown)


rachelnabors (changes requested):
Wow, what an amazing contribution! Before I can merge, though, I have one favor to ask: please have a look through our [styleguide](https://github.com/facebook/react-native-website/blob/master/STYLEGUIDE.md#you-we-and-they) and adjust your pronouns from "we" to "you". 

Thanks again! Can't wait to accept your PR!

cpojer (approved):
Awesome, thank you for this contribution!

PR #2225, opened 2020-09-27
Adding Android NDK Installation Guide

gedeagas:
This PR aims to add an Android NDK installation guide that will be required by React Native version `0.64`. 

See here for reference. https://github.com/facebook/react-native/commit/18ffe12203d03b4e960d61d7bb50cd02bba94663

@Simek Can you add `Wait on future Release` tag for this?. When it does land and gets included in a React Native release, this doc change should accompany it.

cc @safaiyeh


cpojer (approved):
Thank you!

PR #2242, opened 2020-10-03
Docker for React Native Website

VISWESWARAN1998:
Added Dockerfile for easy deployment.

Thanks and Regards,
Visweswaran N

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,3 @@
-# SWAMI KARUPPASWAMI THUNNAI
-
 FROM node
 
 COPY . /rn_website
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+# SWAMI KARUPPASWAMI THUNNAI
+
+FROM node
+
+COPY . /rn_website
+WORKDIR /rn_website/website
+
+RUN npm install
+
+CMD ["npm", "start"]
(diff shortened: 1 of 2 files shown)


VISWESWARAN1998:
I have considered adding a docker file because anyone with docker can easily deploy and test the website without any tools and installation procedures.

rachelnabors:
That seems really cool, but I'm not sure that's a use case we're ready to support. Could you submit an update to the Repo's README to explain how to use as well?

VISWESWARAN1998:
http://react-native-website.visweswaran.com is deployed using docker for your reference.

Simek (changes requested):
Hey @VISWESWARAN1998, thank you for proposing this change.

Can you make sure that the `Dockerfile` is correct after the website has been migrated to the new framework? And please, can you remove the comment in this file?

Also it is not clear enough for me what are the benefits of offering and using the website with Docker? Can you elaborate on that?

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node
+
+COPY . /rn_website
+WORKDIR /rn_website/website
+
+RUN npm install
```
Simek, line 6: It is possible that `npm` command won't be working correctly after the introduction of Yarn workspaces in the repository. Can you test this out against current `master`?

rachelnabors:
Going to close because Docker support seems like it is out of scope for this project

PR #2244, opened 2020-10-03
remove link in native-modules-ios

luism3861:
this PR remove link in `can be invoked`  in `native-modules.ios.md` I try to found this file in version 
https://github.com/facebook/react-native/blob/0.63-stable/ReactCommon/turbomodule/core/platform/ios/RCTTurboModuleManager.mm#L756  it doesn't exists this specific line.
in master tree with recent version `turbomodule folder ` doesn't exists.


Simek (approved):
Thank you for spotting and removing the invalid link! 👍

PR #2245, opened 2020-10-04
Update network.md to show android non ssl block

s-maurice:
added note for android's android:usesCleartextTraffic flag in manifest file to allow non ssl encrypted networking - to go alongside the existing note for ios.


Simek (approved):
@s-maurice Thank you for adding this information! 👍

PR #2248, opened 2020-10-05
Fix links for image.md

luisaverza:
Fix broken links that redirects to others pages inside react-native-website.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -232,13 +232,13 @@ dependencies {
 | ----- | -------- |
 | style | No       |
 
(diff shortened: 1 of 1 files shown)


luisaverza:
Ops, finally I sign the CLA, is possible to run the check again? 😬

Simek:
Hi @luisaverza, sorry but this is not correct way to fix the problem in V1. 

We are also in process of migrating website to Docusaurus V2, which will also, as a byproduct, fix most of the links issues.

PR #2249, opened 2020-10-05
Update textinput.md

ralph-dev:
This will make it more clear for beginners who are incorrectly trying to use onBlur to do a final action on their textInput's text value.


ralph-dev:
Signed!

Simek (approved):
Thank you for sharing this tip! 👍

PR #2250, opened 2020-10-06
Remove unnecessary python dependency

sunnylqm:
Python is not required anymore on windows platform. The only binary addon for react-native project are some deps of `ws`, [which now have prebuilt binary and are optional](https://github.com/websockets/ws#opt-in-for-performance-and-spec-compliance). 
Tested on a brand new windows 10 image with only node installed, the initiating process goes well.

WX20201006-121604

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -135,7 +135,7 @@ You will need Node, the React Native command line interface, a JDK, and Android 
 
 <h2>Installing dependencies</h2>
 
-You will need Node, the React Native command line interface, latest version of Python, a JDK, and Android Studio.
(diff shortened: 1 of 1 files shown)


Simek (approved):
@sunnylqm Thank you so much for working on the removal verification and posting this PR, great work! 👍 

Sorry that the review process was a bit lengthy than usual but the removal had to be verified independently so we can be sure that it is safe to skip Python installation in the starting guide. 

I can confirm that not only the `init` but also the Android app build was successful on my test VS. I have also ask the community for feedback and it looks like everyone is happy with this change.

PR #2251, opened 2020-10-09
docs: change url for react-native-segmented-control

Naturalclar:
The repository for react-native-segmented-control has recently changed to https://github.com/react-native-segmented-control/segmented-control , so I've fixed the url.

The package name is still `@react-native-community/segmented-control` for now, but this will also change in the future, so I'll make the change once that happens.


Simek (approved):
Thank yo for the update and handling the old versions too! 🙂

PR #2254, opened 2020-10-15
Update List Views -  No need height

j-j-gajjar:


Simek:
hi @j-j-gajjar, thank you for proposing this change. The `height` does not affect the "Web" preview but applies to Android and iOS devices. Thank to this style property on all platforms rows have the same height (otherwise they don't).

PR #2265, opened 2020-10-22
Added onScrollBegin & EndDrag to the documentation

devofthings:
I didn't found anything about these possible actions in this documentation so I added them.
Some other people also had this problem:
https://stackoverflow.com/a/43859386

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -445,6 +445,26 @@ If provided, a standard RefreshControl will be added for "Pull to Refresh" funct
 
 ---
 
+### `onScrollBeginDrag`
+
+Called when the user begins to drag the scroll view.
+
+| Type     | Required |
+| -------- | -------- |
+| function | No       |
+
+---
+
+### `onScrollEndDrag`
+
+Called when the user stops dragging the scroll view and it either stops or begins to glide.
+
+| Type     | Required |
+| -------- | -------- |
+| function | No       |
+
+---
+
 ### `onViewableItemsChanged`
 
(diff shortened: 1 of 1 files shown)


devofthings:
Optional: Would be cool if you can tag this with the hacktoberfest-accepted label. If not its also not a problem I just wanted to improve the documentation again 😸

charpeni:
Deploy preview for *react-native* ready!

Built with commit e68dcfdd46e42c6cec1ef6f5075f82f61d050ecf

https://deploy-preview-2265--react-native.netlify.app

Simek:
Hi @devofthings, there are the [`VirtualizedList` component props](https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js#L1653), then `ScrollList` component inherits them and finally `FlatList` by itself. 

The props you are looking for are documented on the website in here:
* https://reactnative.dev/docs/next/scrollview#onscrollbegindrag

Unfortunately we do not list props inherited by the component, but lately, in `next` branch we have added the inherited props links to the TOCs so it should be easier to find those properties.

devofthings:
@Simek Thats a strange rule but I understand it otherwise you guys need to track everything twice. The links will help! I guess this PR can be closed then?

Simek:
@devofthings Thank you for understanding! 🙂 I can agree that we should think about handling the inherited props in a better way (but not multiply the maintaince work). I will try to sort this out a bit after the V2 migration, maybe we can use MDX fragments for those props.

PR #2267, opened 2020-10-23
Bump Node minimum required version

hramos:
To be merged once https://github.com/facebook/react-native/pull/30237 lands, as it will affect the next OSS release to be cut.
Node 14 becomes the new Active LTS on October 27, 2020, which means we can bump the minimum to Node 12 then.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -152,7 +152,7 @@ brew install node
 brew install watchman
 ```
 
-If you have already installed Node on your system, make sure it is Node 10 or newer.
+If you have already installed Node on your system, make sure it is Node 12 or newer.
 
 [Watchman](https://facebook.github.io/watchman) is a tool by Facebook for watching changes in the filesystem. It is highly recommended you install it for better performance.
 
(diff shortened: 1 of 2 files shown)


charpeni:
Deploy preview for *react-native* ready!

Built with commit 0bb84080cd0b52b56957c4a51832cb94404b26dd

https://deploy-preview-2267--react-native.netlify.app

hramos:
This can be merged now.

Simek:
@hramos Would you like to resolve the conflicts? Or would you like me to resolve them? But generally I think that creating new PR might be easier since in V2 migration the Getting Started page has been split to the fragments. 🙂

hramos:
I'll send a new PR

Simek:
@hramos You can also review my PR which is already pushed #2292.

hramos:
@Simek didn't see this in time! I sent https://github.com/facebook/react-native-website/pull/2301. The Node bump shouldn't apply to older RN versions since the change just landed, so I think the change only needs to be made in the docs/ directory.

PR #2268, opened 2020-10-26
Netlify publish dir should be relative to base dir

slorber:
Fix Netlify config for the archive branch:
- publish URL is relative to base
- don't reduce versions for the production archive.reactnative.dev deployment

cc @Simek


Simek (approved):
LGTM, thank you for working on the archive deployment! 👍

PR #2270, opened 2020-10-28
Update view-style-props.md

kamal-20:


Simek:
Hi @kamal-20, the current styling in example is intended. 

The preview should look like this (with no border radius on the middle element):
Screenshot 2020-10-29 125158

PR #2271, opened 2020-10-29
Remove useless v1 translation file

slorber:
This file is from Docusaurus v1 and is only useful for v1 translated sites.
In Docusaurus v2 it won't be used.

cc @Simek


Simek (approved):
@slorber Thank you for cleaning this up! We are eagerly waiting for the i18n support in V2 🙂

PR #2263, opened 2020-10-20
Migrate React Native website to Docusaurus V2

Simek:
This PR migrates the React Native website to Docusaurus V2. 

This is an effect of collaboration with @slorber and MLH Docusaurus team (@darshkpatel & @teikjun).

It is the initial import, rebased against current `master` in the main repository.

Refs: #2146, #2139 

Fixes #2208, Fixes #2054

### Preview

* https://deploy-preview-2263--react-native.netlify.app

### Noticeable changes for website user
* SPA - reduced page load times, more fluid navigation
* additional dark theme and theme toggle (dark theme Snacks too)
* build in tabs sync
* Snacks appears on mobile without the preview but with link to open example in Expo app
* blog posts are now searchable
* code blocks content can be copied by click - "Copy" button appears on hover
* old versions and `next` page have a notice for users that they are reading old or upcoming version of the page

### Noticeable changes for contributors/maintainers
* announcement bar support out of the box
* MDX fragments - which allows to split content and helps to maintain large pages (example: Environment Setup)
* extended page customizability and theme components swizzling
* improved build times
* still growing set of ready to use plugins - https://v2.docusaurus.io/docs/using-plugins/
* `<!--truncate-->` tag for blog posts to hide the details on the post list, blog posts tags

#### You can read more about V2 features and changes in the official documentation:
* https://v2.docusaurus.io/docs/

PR #2273, opened 2020-10-29
fix Docusaurus v2 CircleCI/GH-pages deployment

slorber:
Follow-up of Docusaurus v2 migration so that site can go live.

https://github.com/facebook/react-native-website/pull/2263

Note: this site is published to GH pages of react-native repo, not react-native-website repo, which is kind of weird

Highly recommend to migrate Netlify for everything (build production, host production, PR deploy previews) instead of splitting each task between CircleCI/GH-pages/Netlify. One tool is simpler to manage than 3, technically but also for permissions.
Using GH pages is risky at this scale because we can only have 1 deployment per repo, which means any error can make the production fail, while with Netlify we can have 2 deployments side by side and just change the DNS to perform a migration. Also, Netlify allows many more options like server redirects, server caching headers (without the need to use CloudFlare on top)


.circleci/config.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       - restore_cache_checkout
       - run_yarn
       - run: yarn ci-check
+      - run: export NODE_OPTIONS=--max_old_space_size=4096
```
slorber, line 63: A site build requires more than default node memory.

Deploy builds the site and use 4go. But tests workflow also build the site, so it should rather use 4go as well.

Simek (approved):
Thank you for working on this fix! Hope this make V2 website live. 🙂

PR #2279, opened 2020-10-29
remove few files, cleanup scripts and meta files after the migration

Simek:
This PR clean up some remnants of V1 after the migration to V2 website:
* `netlify-reduce-versions.js` is no longer required, the similar effect, when more versions appears, can be archived using `onlyIncludeVersions` field from docs plugin: https://v2.docusaurus.io/docs/using-plugins/#docusaurusplugin-content-docs,
* `sync-guides.js` and `versionedDocsBlacklist.json` were required to archive non-versioned pages (but this was not working in V1 for some time too), in V2 we can add another docs plugin with versioning disabled to have static, doc pages not affected by versioning,
* `sync-community-repos.js` was a nice idea (a script which was able to fetch data about the community repositories), but after lastest changes to RNC organization (https://github.com/react-native-community/discussions-and-proposals/blob/master/partners/0002-apply-organization-repository-policy.md) this script makes not much sense anymore (and the values were not used in V1 anywhere).


slorber:
LGTM 👍

PR #2282, opened 2020-10-29
fix homepage animations

slorber:
Make them work on Safari + better docusaurus integration


Simek (approved):
Thank you for fixing the homepage animations, great work! 👍

PR #2284, opened 2020-10-29
wrong prop usage fixed

alierenkara:
Titles prop is given in button component in tutorial, it is fixed as title


Simek (approved):
Thank you for the fix and addressing it in the latest versioned docs too! 👍

PR #2287, opened 2020-10-30
Fix links in components-and-apis.md

tizee:
Links in  https://reactnative.dev/docs/components-and-apis using incorrect relative urls.
Using absolute urls with pattern like `/docs/xxx` instead of `./xxx`.

### DIFF
--- a/docs/components-and-apis.md
+++ b/docs/components-and-apis.md
@@ -5,12 +5,12 @@ title: Core Components and APIs
 
 React Native provides a number of built-in [Core Components](intro-react-native-components) ready for you to use in your app. You can find them all in the left sidebar (or menu above, if you are on a narrow screen). If you're not sure where to get started, take a look at the following categories:
 
-- [Basic Components](components-and-apis#basic-components)
-- [User Interface](components-and-apis#user-interface)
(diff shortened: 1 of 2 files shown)


Simek (changes requested):
Hi @tizee, does the links are not working for you right know? I have no problem with them.

Adding `docs` in URL results in linking always to the latest version of docs, not the version user is currently on.

Also CI in this PR failed due to links issue, can you look at that?

Simek:
Okay, I was able to determine that the problem is with the trailing slash in the link:
* https://reactnative.dev/docs/components-and-apis/

If there is a slash at the end of URL, the links on the initial page are broken, until the next navigation event (which cleans up the slash and fixes the link issues for the rest of visit). Unfortunately currently the slash is added automatically on the first enter to the website (when you are using links directly to the certain pages), even when you are using a link without a trailing slash. 

It looks like the issue is related to this proposal in the Docusuaurs repository:
* https://github.com/facebook/docusaurus/issues/3372. 

We will investigate this problem further.

Simek:
Closing in favor of #2291.

PR #2290, opened 2020-10-30
attempt to remove trailing slash redirects: remove .nojekyll?

slorber:
cc @Simek


Simek:
Refs: #2287

Simek (approved):
Let's try that.

PR #2292, opened 2020-10-30
bump Node required version in environment setup guides

Simek:
This PR introduces the same changes as #2267, but it is based on the code after the migration to the new website.


Simek:
Closing in favor of #2301.

Issue #2291, opened 2020-10-30
Broken links and trailing slash issues on the website


Simek:
## Description

Currently the trailing slash is been added to the initial links leading to the React Native website which results in broken links (leading to 404 page) for some users on the initial page that they are visiting, for example:
* `https://reactnative.dev/docs/components-and-apis` becomes `https://reactnative.dev/docs/components-and-apis/`
* `https://reactnative.dev/docs/testing-overview#static-analysis` becomes `https://reactnative.dev/docs/testing-overview/#static-analysis`
* relative internal link `https://reactnative.dev/docs/environment-setup` becomes `https://reactnative.dev/docs/[NAME_OF_PAGE]/environment-setup`

There are few PRs which were attempting to solve single cases (#2286 and #2287), but I think that we need a global solution for that.

The home page `ActionButton` links are using `baseUrl` so the page on which user lands will also contain trailing slash, so the links will be broken. 

**Updating links to the non-relative version will results in static linking to the latest version (not the version user currently is viewing), which can result in unexpected docs version changes while browsering.**

I'm not sure if there is a way to configure the GitHub Pages or Jekyll redirects in a way that will solve this issue. 

We will be looking further into solving this problem.

## Documentation version

This issue affects the whole website.

PR #2297, opened 2020-10-31
attempt to fix trailing slash issues on GH pages

slorber:
cc @Simek , should fix Github Pages trailing slash issues

/myDoc will not redirect 301 to /myDoc/ anymore

Temporary fixes https://github.com/facebook/react-native-website/issues/2291


Simek (approved):
LGTM, thank you for your work on the fix for those issues! 👍

PR #2299, opened 2020-11-01
Chaged from 'background' to 'backgroundColor'

PhoenixCreation:
Changes:
```diff
- background: "lightGrey"
+ backgroundColor: "lightgrey"
```

In example of LayoutAnimation stylesheet for 'title' contains 'background' property.
It throws error as it should be 'backgroundColor'.
also value of the same is 'lightGrey' which is not supported. It should be 'lightgrey'.

### DIFF
--- a/docs/layoutanimation.md
+++ b/docs/layoutanimation.md
@@ -55,7 +55,7 @@ const App = () => {
 
 const style = StyleSheet.create({
   tile: {
-    background: "lightGrey",
+    backgroundColor: "lightgrey",
     borderWidth: 0.5,
     borderColor: "#d6d7da"
   },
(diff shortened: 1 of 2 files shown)


Simek (changes requested):
Nice catch, thank you! Can you also update the main (`next`) docs at https://github.com/facebook/react-native-website/blob/master/docs/layoutanimation.md?

PhoenixCreation:
@Simek Yeah. Sure. But I will do it tomorrow. Thanks you.

PhoenixCreation:
I have updated at main docs. Let me know if there is any issue with this.

Simek (approved):
LGTM, thank you for fixing this issue! 👍

PR #2301, opened 2020-11-02
Bump minimum Node version to 12

hramos:
Node 14 is the new active LTS as of October 27, 2020, so we have dropped support for Node 10 and declared Node 12 LTS as the minimum supported version going forward. This should only affect releases 0.64 and onward.


Simek (approved):
LGTM, thank you sending this changes again! 👍

PR #2302, opened 2020-11-02
[PlatformColor] updated supported colors section, updated example converted to snack

Simek:
This PR updates the `PlatformColor` API page and introduce the following changes:
* header for supported colors list
* colors list moved above Developer Notes
* mention about supporting both notations (Objective-C and Swift) on iOS
* additional link to supported [UIColor Standard Colors](https://developer.apple.com/documentation/uikit/uicolor/standard_colors) on iOS
* example converted to Snack
* example content updated in a way which shows colors usage from each supported list for both platforms

The changes has been also backported to `0.63` docs. 

Earlier due to lack of support of `PlatformColor` in Expo SDK guide had to use plain code blocks, but in a meanwhile the SDK 39 has been released which includes this API.

### Preview

Screenshot 2020-11-02 215621

### DIFF
--- a/docs/platformcolor.md
+++ b/docs/platformcolor.md
@@ -19,6 +19,17 @@ PlatformColor('bogusName', 'linkColor');
 
 Since native colors can be sensitive to themes and/or high contrast, this platform specific logic also translates inside your components.
 
+### Supported colors
+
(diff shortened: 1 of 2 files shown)


cpojer (approved):
Nice, thank you!

PR #2309, opened 2020-11-03
[image-component] add docu for method parameters and return value

WookieFPV:
The Image component methods "prefetch" and "queryCache" are async functions.
I added this to the docs and added the android only callback of the prefetch method. 
(RN-source: https://github.com/facebook/react-native/blob/master/Libraries/Image/Image.android.js#L197)

The prefetch method returns a Promise<boolean> (that is always true but added it anyway).
Elements of the "queryCache" result can also be "disk/memory", 
(RN-source: https://github.com/facebook/react-native/blob/master/Libraries/Image/Image.android.js#L214)

I could add some example code to the queryCache method.
Not sure if that would be helpful.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -505,16 +505,17 @@ Does not work for static image resources.
 ### `prefetch()`
 
 ```jsx
-Image.prefetch(url);
+await Image.prefetch(url);
 ```
 
-Prefetches a remote image for later use by downloading it to the disk cache
+Prefetches a remote image for later use by downloading it to the disk cache. Returns a promise which resolves to a boolean.
 
 **Parameters:**
 
-| Name | Type   | Required | Description                       |
-| ---- | ------ | -------- | --------------------------------- |
-| url  | string | Yes      | The remote location of the image. |
+| Name       | Type     | Required | Description                                                         |
+| ---------- | -------- | -------- | ------------------------------------------------------------------- |
+| url        | string   | Yes      | The remote location of the image.                                   |
+| callback   | function | No       | [Android only] The function that will be called with the requestId. |
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/image.md
```diff
@@ -505,16 +505,17 @@ Does not work for static image resources.
...
 
 **Parameters:**
 
-| Name | Type   | Required | Description                       |
-| ---- | ------ | -------- | --------------------------------- |
-| url  | string | Yes      | The remote location of the image. |
+| Name       | Type     | Required | Description                                                         |
+| ---------- | -------- | -------- | ------------------------------------------------------------------- |
+| url        | string   | Yes      | The remote location of the image.                                   |
+| callback   | function | No       | [Android only] The function that will be called with the requestId. |
```
Simek, line 518: Can you use the platform label for marking the Android only parameter like on the Modal page? 

https://github.com/facebook/react-native-website/blame/master/docs/alert.md#L211

Simek (changes requested):
Besides the comment above the changes LGTM. Thank you for updating the docs! 👍

docs/image.md
WookieFPV: I used the android label android like on the Alert Page, not sure how this will look.

Simek: Deploy failed because there is a syntax error on the page. It looks like not the whole label HTML has been added.

Simek: > not sure how this will look

You can check how label looks on the Alert page: https://reactnative.dev/docs/next/alert#methods

WookieFPV: thanks for the help, I fixed it

Simek (approved):
Thank you for applying review suggestion! 👍

PR #2315, opened 2020-11-11
Update getting-started.md

Prashoon123:
Please accept my pull request, because earlier the image was not getting shown now I changed the source of the image to that of the website.


Prashoon123:
Please accept my pull request, because earlier the image was not getting shown now I changed the source of the image to that of the website.

PR #2317, opened 2020-11-13
Remove HAXM requirement

louiechristie:
Remove HAXM requirement because hard it's:
1. not needed
2. virtually impossible to install on newer macOS:

## Not needed

 Not used on OS X v10.10 Yosemite and higher:

"For OS X v10.10 Yosemite and higher, the Android Emulator uses the built-in Hypervisor.Framework by default, and falls back to using the Intel Hardware Accelerated Execution Manager (HAXM) if Hypervisor.Framework fails to initialize (such as when running on OS X v10.9 or earlier)."
https://android-developers.googleblog.com/2018/07/android-emulator-amd-processor-hyper-v.html

Not supported on OS X v10.9 or earlier:

"Supported Operating Systems:
...
macOS* 10.12, 10.13, 10.14"
https://github.com/intel/haxm/releases/tag/v7.6.5"
(https://github.com/intel/haxm/releases/download/v7.6.5/haxm-macosx_v7_6_5.zip -> Release Notes.txt)

So by my deduction, it's not needed on any OS X / macOS then?

# Virtually impossible to install on newer macOS

"Since macOS Catalina 10.15 Beta 4, HAXM cannot be upgraded from Android SDK Tools directly because the authorization services upgrade in macOS. A workaround method is to download the release package from the official release page and run 'sudo ./silent_install.sh' manually after extraction."
https://android-developers.googleblog.com/2018/07/android-emulator-amd-processor-hyper-v.html

"Note: For users of macOS 10.13 High Sierra and higher: macOS 10.13 disables installation of kernel extensions by default. Because Intel HAXM is a kernel extension, you might need to manually enable its installation. For more details, see Known Issues."
https://developer.android.com/studio/run/emulator-acceleration#vm-mac

From above you can see because of new Apple security it's virtually impossible to install HAXM on newer macOS's  - particularly for new users who this guide is aimed at. I'm experienced and failed after hours trying to install manually on macOS 10.15, and I know what ```chmod +x``` means (which a lot of new React Native devs probably won't know.)

If we want to encourage more adoption of React Native over, say Flutter, I think getting started needs to be easy for new devs, so I think we should dropping the HAXM requirement in the documentation would be a good step.

### DIFF
--- a/docs/_getting-started-macos-android.md
+++ b/docs/_getting-started-macos-android.md
@@ -38,7 +38,6 @@ Setting up your development environment can be somewhat tedious if you're new to
 - `Android SDK`
 - `Android SDK Platform`
 - `Android Virtual Device`
-- If you are not already using Hyper-V: `Performance (Intel ® HAXM)` ([See here for AMD or Hyper-V](https://android-developers.googleblog.com/2018/07/android-emulator-amd-processor-hyper-v.html))
 
 Then, click "Next" to install all of these components.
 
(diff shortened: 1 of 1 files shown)


cpojer (approved):
👍 

Should we remove this from instructions for older versions too?

PR #2319, opened 2020-11-15
avoid confusing clipboard copy behavior

neodon:
On the React Native documentation website, code snippets have a `Copy` button in the top right corner.  So if the commands for `yarn` and `npm` are in the same code block, the user might click `Copy` and then paste both commands into their shell.


cpojer (approved):
Nice fix!

PR #2143, opened 2020-08-14
Add missing package

ani4aniket:
## Summary
Add the missing `he@1.2.0` package, which was accidentally removed [here](https://github.com/facebook/react-native-website/pull/2129/commits/f38f62f680043b1b65b9043e8eb40085b4a7ef62#diff-07f3a4647460e74f9514d8def137963aL49).

This package was responsible for encoding string which is being used [here](https://github.com/facebook/react-native-website/blob/6e41a04f987b8161baaa578bb3334857e4896348/website/scripts/sync-api-docs/propFormatter.js#L38).

## Changes
Add the missing package `he@1.2.0` as devDependency.

## Changelog
[Internal]

PR #2323, opened 2020-11-16
use Yarn workspaces, convert sync-api-docs and remark Snack plugin

Simek:
Refs #2143

This PR adds Yarn workspaces to the `react-native-website` repository. 

API sync script (`sync-api-docs`) and remark Snack plugin (`plugins/remark-snack-player`) has become the stand alone workspaces, which should make easier to handle the dependencies and help to keep the repositorium tidy.

This migration should also fix the missing package problem mentioned in the referenced PR.

CC @motiz88


cpojer (approved):
Cool!

PR #2340, opened 2020-11-18
Update navigation.md

lguardado:
Use route params rather than a hardcoded name value in the [React Navigation example code](https://reactnative.dev/docs/navigation#usage).

Since the example is using a Params object to pass the profile's Name, it would be nice to actually use that param rather than hardcoding the name.

### DIFF
--- a/docs/navigation.md
+++ b/docs/navigation.md
@@ -115,8 +115,8 @@ const HomeScreen = ({ navigation }) => {
     />
   );
 };
-const ProfileScreen = () => {
-  return <Text>This is Jane's profile</Text>;
+const ProfileScreen = ({ navigation, route }) => {
+  return <Text>This is { route.params.name }'s profile</Text>;
 };
 ```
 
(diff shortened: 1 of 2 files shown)


Simek (changes requested):
Thank you for this code improvement! 👍 

Can you also update the current docs (`Next`) located in the [`docs/`](https://github.com/facebook/react-native-website/tree/master/docs) directory, besides the latest, versioned file?

lguardado:
> Thank you for this code improvement! 👍
> 
> Can you also update the current docs (`Next`) located in the [`docs/`](https://github.com/facebook/react-native-website/tree/master/docs) directory, besides the latest, versioned file?

Thank you for the review @Simek!
Right, great catch!  I just pushed change already but let me know if I'm missing something else.

Cheers!

Simek (approved):
LGTM, thank you for updating the PR! 🙂

PR #2346, opened 2020-11-21
Update text color for dissection images

darshkpatel:
After the v2 migration the website supports dark mode, the dissection image currently has black font over grey background when the website is in dark mode
Screenshot 2020-11-21 at 3 10 23 PM

Since the dissection images were PNG's and not SVG's i have used photoshop to update the text color to `#61DAFB` (Brand Color as configured in the theme) which means there'd be no need to switch the 4 dissection images in the dark mode vs light.

Preview of updated dissection animation: 
Screenshot 2020-11-21 at 3 17 33 PM
Screenshot 2020-11-21 at 3 17 29 PM


Simek (changes requested):
Unfortunately there is visible quality degradation (pixelation) of the recolored text .
Screenshot 2020-11-21 124534
You can try remove part of image and add it back in correct colors instead of recoloring (if the source for those PNGs is not available). This should yield a better effect.

But the best solution (and most time consuming), would be recreating the images in SVG and then recolor text and arrows via CSS.

CC @rachelnabors

darshkpatel:
> You can try remove part of image and add it back in correct colors instead of recoloring (if the source for those PNGs is not available). This should yield a better effect.

Done!

Simek:
> Done!

This looks better, but I have three more, minor requests:
* change the band color to the `--subtle` color (`#7a7a7a`), this shade of gray should be quite good visible in both themes (it is used, for example, in the collapsible sidebar groups headers)
* ensure you are using correct font size (`17px`) with correct text anti-aliasing to match the previous texts
* add the Photoshop source file used for creating new images to the `/design` directory

Simek (approved):
@darshkpatel Thank you so much for all the work you put into this PR! This animation is looking way better now and thanks to source file, the future edits will be even easier! 🙂

PR #2353, opened 2020-11-26
Installation word spelling mistake

hiimharsh:


Simek (changes requested):
Hi @hiimharsh, thank you for the fix! 👍

Can you also update the current docs ("Next") located in the [`docs/`](https://github.com/facebook/react-native-website/tree/master/docs) directory, besides the latest, versioned file? 

The files from `docs/` are used to generate new versioned docs, so without updating them, your change will be only available in `0.63` docs and will not be included further versions.

Simek (approved):

PR #2361, opened 2020-12-01
Update hermes.md to include RN compatibility

Huxpro:
Hermes works more reliably following the versions explicitly documented in its releases page.

This is suggested by the community at https://github.com/facebook/hermes/issues/33#issuecomment-727701625


Simek (changes requested):
Thank you for adding that note! 👍 

Like on the other PR lease updated also the file located in `/docs` to preserve the change for the future docs versions.

PR #2259, opened 2020-10-17
Update tutorial.md

jjmax75:
update text to refer to the current examples ES6 features

PR #2366, opened 2020-12-05
Extends and class are no longer part of the example code 'above'. Add…

magnusbrattlof:
Extends and class are no longer part of the example code 'above' Added the arrow function as an additional example.


Simek (approved):
Hi @magnusbrattlof, thank you for pushing an update for the docs! 👍 

The issue you are referring to has already been fixed in base docs (`/docs/`) in #2259 PR, but versioned docs have not been updated. To avoid unnecessary differences between the docs versions I have copied the fix from base page.

PR #2369, opened 2020-12-05
change Docs sidebar groupings, add small indentation for pages, adjust sidebar design

Simek:
This PR changes the Docs sidebar grouping a bit, utilizing more subgroups (categories):
* Connectivity has been moved into Workflow,
* Interaction and Inclusion groups have been moved into Design,
* Native Modules and Native Components have been merged into Native Code group,
* Guides (Android) and Guides (iOS) have been merged into Guides with categories for each platform.

In addition to the structure changes I have also improved the sidebar design a bit. Not pages includes small indentation so it will be easier at the first glance differentiate group heading from the page. Also the sub groups heading arrows have been shrinked a bit to avoid confusion with main group arrow.

### Preview
Screenshot 2020-12-05 172525
Screenshot 2020-12-05 172733


rachelnabors:
Not a bad idea, @Simek! But I feel like "guides" is offer here—these contain platform-specific information. I wonder if such a heading could be more specific—"Android and iOS guides".

(I was looking through the guides and started to wonder, "Why do we have some of these? Like isn't https://reactnative.dev/docs/running-on-simulator-ios also on https://reactnative.dev/docs/environment-setup ?")

Simek:
Thank you for the feedback @rachelnabors, I have updated the title to the one, you have suggested.

> (I was looking through the guides and started to wonder, "Why do we have some of these? Like isn't https://reactnative.dev/docs/running-on-simulator-ios also on https://reactnative.dev/docs/environment-setup ?")

IDK at this moment of time, I will try to research the history, and consolidate the articles (if possible) in the later date. It would be nice if you can create an issue, so we don't lose this feedback and task alter PR merge 🙂

TheSavior:
I defer to @rachelnabors since this is about docs structure

lunaleaps:
I really like the indentation change! Also +1 to the guides restructure. I think the only thing that makes me pause is the grouping of "native code". Could we leave the Native component and module sections as-is? I think we've taken issue with the term "native" and as our many platform vision gets more realized, we may move to new terminology so I'd want to avoid adding a new term "Native Code".

lunaleaps (changes requested):
Re: comments about "Native Code" grouping

Simek:
I have synced up this branch with `main` so it now contains the styles for the `collapsible: false` entries (which prevents the groups subsection from being intractable. - I think that there is no much gain from allowing collapsing those parts, and this alows to reduce the amount of arrows on the right side of sidebar, which IMO improves the readability a bit). Additionally I have updated the the sidebar grouping as Luna suggested.

One more change that I think of is renaming "Android and iOS guides" to something shorter, like "Other guides", or maybe "Assorted Guides", but I'm not sure if it is a good idea. ATM I don't like the platform name repetition.

### Current Preview

Screenshot 2021-12-24 at 09 44 46

lunaleaps:
LGTM, I think we can circle back on the titling of "Android and iOS guides", or feel free to start another PR for that, let's get these changes in!

lunaleaps (approved):

Simek:
> LGTM, I think we can circle back on the titling of "Android and iOS guides", or feel free to start another PR for that, let's get these changes in!

I was also thinking about this, my ideas were: "Other Guides", "Platform Guides" and "Assorted Guides".

PR #2384, opened 2020-12-15
Add a missing brackets

sunnylqm:


Simek (approved):
Hi @sunnylqm, thank you for pointing this problem out! 

Since the `Button` component do not accept `children` prop there is no need for stand-alone close tag. I found it misleading and not coherent with other examples, so I have change example a bit.

PR #2391, opened 2020-12-17
Fix comma splice

karenying:
by adding a conjunction and making the verb tenses consistent.


Simek (approved):
Thank you for the fix! 👍

PR #2393, opened 2020-12-18
add curly braces in main component example

luism3861:
this PR add curly braces in code example i think it looks bad without curly braces for people try to copy this component!.


Simek (approved):
Nice, thank you for the tweak! 👍

PR #2396, opened 2020-12-20
Update layout-props.md

bronsonavila:
It sounds awkward to say that `flex` in React Native works according to "the Yoga". According to the [Yoga homepage](https://yogalayout.com/), Yoga is "a highly optimized open source layout engine". Thus, it would be clearer to the reader if we change the language to say that `flex` works according to "the Yoga layout engine".

### DIFF
--- a/docs/layout-props.md
+++ b/docs/layout-props.md
@@ -339,7 +339,7 @@ This style takes precedence over the `left` and `right` styles.
 
 ### `flex`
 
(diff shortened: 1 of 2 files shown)


bronsonavila:
I will resubmit after my CLA signature has been processed.

Simek (approved):
Thank you for the update! 👍

PR #2397, opened 2020-12-20
Update platform-specific-code.md

bronsonavila:
- Improve grammar by changing "you can also use it to return platform specific component" to "you can also use it to return _platform-specific components_".


Simek (approved):
Thank you for the fix! 👍

PR #2400, opened 2020-12-22
Remove unnecessary brackets from native-components-ios.md

inokawa:


Simek (approved):
Nice catch, thank you for correcting the example code! 👍

PR #2404, opened 2020-12-23
Update index.js

sunnylqm:
tutorial.md is deprecated


Simek:
Hi @sunnylqm, the Tutorial page is not deprecated, it is just not a part of any sidebar, so the current URL is correct and leads to expected page.

PR #2405, opened 2020-12-23
Link directly to Reactiflux's React Native channels

nickmccurdy:
Reverts #1292 as we now allow direct links to channels again

### DIFF
--- a/website/src/pages/help.md
+++ b/website/src/pages/help.md
@@ -38,7 +38,7 @@ There are a lot of [React Native Meetups](http://www.meetup.com/topics/react-nat
 
 ### Reactiflux Chat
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @nickmccurdy, both links currently works for me. What is the reason form switching from `reactiflux` to the generated id?

nickmccurdy:
The new link goes directly to the #react-native channel, I believe the old one just goes to the #welcome channel if you're new. However that's more of a general discussion channel than a help channel, if you prefer I can change it to #help-react-native, our largest help channel, though it's not under the React Native category.

Simek (approved):
Thank you for explaining the details underneath! Changes LGTM, thank you! 👍

PR #2410, opened 2020-12-25
Improve language used in navigation.md

bronsonavila:
- Change "For iOS with bare React Native project..." to "For iOS in a bare React Native project..."


Simek (approved):
Thank you! 👍

PR #2411, opened 2020-12-25
Update animations.md

bronsonavila:
- Replace unnecessary `React.useEffect` with simply `useEffect` (as the named export `useEffect` has already been imported)


Simek (approved):
Nice catch, thank you! 👍

PR #2422, opened 2020-12-31
Fix typo of uiMode

inokawa:


Simek (approved):
Good catch, thank you! 👍

PR #2424, opened 2021-01-03
docs: add another secure alternative to Async Storage in security page

emeraldsanto:
Hello!

I am the author of [react-native-encrypted-storage](https://github.com/emeraldsanto/react-native-encrypted-storage), secure alternative to Async Storage using Keychain and EncryptedSharedPreferences, and wanted to include it in the security docs, along with the existing options.


Simek (approved):
Thank you for updating the libraries list! 👍

PR #2428, opened 2021-01-05
Update animatedvaluexy.md

samuelClo:
Add "useNativeDriver: false" to remove warning


Simek (commented):
Hi @samuelClo, thank you for sending an update! 

Can you tell me on which platform and exactly when the warning you tried to fix appears?

PR #2430, opened 2021-01-05
Update using-a-scrollview.md

withteja:
default is already the name of the export, I believe giving it variable App won't work.


Simek (approved):
Hi @withteja, thank you for pointing this out. 👍 

The issue have already been fixed in the base docs, so I have just copied our general solution over your code.

PR #2434, opened 2021-01-08
Fix syntax error.

tomfriwel:


Simek (approved):
Good catch! thank you for the fix! 👍

PR #2436, opened 2021-01-10
Fix link to profiling in performance doc

shingt:
Noticed link to profiling is dead on and after 0.62. 
(Before 0.62, `Profiling` section exists on the performance doc just like [this](https://reactnative.dev/docs/0.61/performance#profiling), but looks like it's not there now.)


Simek (approved):
Hello @shingt, indeed, the Profiling part has been extracted to the separate page. Thank you so much for the fix and updating base and most of versioned docs! 👍

PR #2440, opened 2021-01-13
Describe new added feature userInterfaceStyle in ActionSheetIOS

hahtml:
In RN codebase, we have added feature to indicate userInterfaceStyle for native action sheet, and the doc is deprecated.


Simek (approved):
Hi @hahtml, thank you for adding the missing config property! 👍 

Unfortunately, the feature do not work in the latest Expo release, so previews on the website are not affected, but the example works in the bare app.

PR #2444, opened 2021-01-15
Fix spacing for Class Component in ActivityIndicator Docs

vuki656:


Simek (approved):
Nice catch @vuki656, thank you! 👍

PR #2448, opened 2021-01-19
Fix swift bug in native-modules-ios.md

luciancic:
I tried the following code example and got an error. Read [this stackoverflow question](https://stackoverflow.com/questions/39692230/got-is-not-a-recognized-objective-c-method-when-bridging-swift-to-react-native) for the details. Since this affects swift 3+ I propose correcting this code example, and optionally mentioning this quirk in a short sentence.


luciancic:
I signed the CLA. Please re-run the checks

Simek (approved):
Hi @luciancic, thank you so much for adding this fix for Native Modules example! 👍

PR #2450, opened 2021-01-20
Fix the description for the *Component props for vritualized lists

satya164:
The docs for components such as `ListHeaderComponent`, `ListFooterComponent` etc. in the virtualized list components (`FlatList`, `VirtualizedList` etc.) say that it can accept a React Component Class, a render function, or a rendered element. However, they don't really accept a render function.

You can see from the source code that the code never calls the passed prop as a function, only as a component or element: https://github.com/facebook/react-native/blob/6079256fcc222fb2581c804c340905ed9d8ac1ca/Libraries/Lists/VirtualizedList.js#L878. The type annotations for this prop also annotate it as a component or element.

While passing a render function (`ListHeaderComponent={() => <SomeThing />}`) appears to work, it's only because this is also the same signature as a function component. But it can actually cause issues. Since using it this way means every re-render will result in a new component, which means it will get unmounted and remounted causing perf issues as well as losing any local state. I've seen people put text input there which causes issues (https://github.com/facebook/react-native/issues/13365)

So I think it's better to remove the incorrect description from the docs.

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -253,21 +253,21 @@ Rendered in between each item, but not at the top or bottom. By default, `highli
 
 ### `ListEmptyComponent`
 
-Rendered when the list is empty. Can be a React Component Class, a render function, or a rendered element.
+Rendered when the list is empty. Can be a React Component (e.g. `SomeComponent`), or a React element (e.g. `<SomeComponent />`).
 
 | Type                         |
 | ---------------------------- |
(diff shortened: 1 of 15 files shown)


Simek (approved):
Hello @satya164, thank you so much for working on those changes and providing the great description and reasoning! 👍 

Hope this improvement reduce the number of issues or problems the users are experiencing with those properties! 🙂

PR #2451, opened 2021-01-21
Fix FlexShrink default value

Titozzz:
On react-native flexShrink default value is not 1 but 0. 1 is the default value for the web and this can be confusing.

### DIFF
--- a/docs/flexbox.md
+++ b/docs/flexbox.md
@@ -983,7 +983,7 @@ export default FlexWrapLayout;
 
 - [`flexShrink`](layout-props#flexshrink) describes how to shrink children along the main axis in the case in which the total size of the children overflows the size of the container on the main axis. `flexShrink` is very similar to `flexGrow` and can be thought of in the same way if any overflowing size is considered to be negative remaining space. These two properties also work well together by allowing children to grow and shrink as needed.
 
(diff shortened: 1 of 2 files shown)


Simek:
@Titozzz Hmm, according to the Yoga Layout docs, the default value is `1` no matter of platform - https://yogalayout.com/docs/flex/. 

Also I was not able to find the code altering this value in Yoga (https://github.com/facebook/yoga/search?p=1&q=flexshrink),, nor React Native (https://github.com/facebook/react-native/search?p=1&q=flexshrink) source.

Can you provide the link to the source or place where this discrepancy can bee seen?

Titozzz:
https://github.com/facebook/react-native/blob/6e6443afd04a847ef23fb6254a84e48c70b45896/ReactCommon/yoga/yoga/Yoga-internal.h#L143

you can see that even react native web uses this default: https://snack.expo.io/EsBsku4O0

Titozzz:
But I do agree that all the docs say it's 1 which is why it's confusing

Simek (approved):
Thank you so much for investigating this, providing a fix and linking the source code! 👍

jzxchiang1:
This is not complete...

Forgot to update https://reactnative.dev/docs/layout-props#flexshrink

PR #2452, opened 2021-01-21
Update animated.md

nth-chile:
`toValue` in `.timing()` config was missing


Simek:
Hi @nth-chile, thank you for attempting to update the docs! 🙂 

Unfortunately the proposed changes are incorrect, because `toValue` is an Animated API method, not the Animated configuration property.

PR #2457, opened 2021-01-24
Added suggestion for `maintainVisibleContentPosition` support on android

vishalnarkhede:
ScrollView and FlatList supports a prop - `maintainVisibleContentPosition`, which is super useful in implementing bi-directional infinite scroll. Although it is only available for iOS right now. Until react-native team adds support for this prop on android, one could use alternative. - [@stream-io/flat-list-mvcp](https://github.com/GetStream/flat-list-mvcp). 

[#29055](https://github.com/facebook/react-native/issues/29055)

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -340,7 +340,7 @@ Caveat 1: Reordering elements in the scrollview with this enabled will probably 
 
 Caveat 2: This uses `contentOffset` and `frame.origin` in native code to compute visibility. Occlusion, transforms, and other complexity won't be taken into account as to whether content is "visible" or not.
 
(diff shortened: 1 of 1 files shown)


vishalnarkhede:
What do you think about this @Simek ? I am going to maintain this library till we have support for this prop OOTB from RN. Might as well recommend it to other users :)

Simek:
Hi @vishalnarkhede, we usually do not include/recommend the complementary, third-party libraries in the Components/APIs docs, rarely we include them in the guides, like for example `Security`.

Also when the page will be updated to utilize the new platform labels, it should be easier for users to see that the property is only available for the iOS platform.

So at this moment of time, I think that the change is not that necessary.

PR #2458, opened 2021-01-25
Fix ScrollView alwaysBounceVertical doc

MoOx:
It seems to be a mistake from a copy paste on `alwaysBounceHorizontal` above :)


Simek (approved):
Hi @MoOx, thank you so much for spotting and fixing this issue! 👍

PR #2460, opened 2021-01-25
Add Osmi CLI reference as a third-party CLI

DeVoresyah:
Add Osmi CLI reference as a third-party CLI. Osmi CLI is an alternative for Ignite CLI


Simek:
Hi @DeVoresyah,

The Ignite CLI has been featured in the guide because Infinite Red is an official React Native Partner - https://github.com/facebook/react-native/blob/master/ECOSYSTEM.md#partners

It would be hard to list every community CLI and other third-part alternatives, and than maintain the list updated.

I would recommend adding Osmi to the React Native Directory instead (https://reactnative.directory/), where Ignite CLI and other alternatives are already present.

PR #2462, opened 2021-01-27
Update Reactiflux invite link

nickmccurdy:
We have replaced the #react-native channel with #help-react-native, which this now links to directly

Related to #2405


charpeni (approved):
Thanks for sending this!

Issue #2399, opened 2020-12-21
Component props inheritance should be shown on the right side next to props list


SurmaAa:
## Description
I think component props inheritance should be shown on the right side next to props list e.g.

## What is the problem?
The `TouchableHighlight` inherits props from `TouchableWithoutFeedback` but there is no information in the docs right side (https://reactnative.dev/docs/touchablehighlight) that it's inherited, the new learner has to scroll down to get know about this which isn't obvious.

## How can we address it?
By adding a information about inheritance. Maybe like in the picture?
![Imgur](https://imgur.com/90aWlv9.jpg)

## Why is it important?
I think it's important for new learners like me because they could find something faster. I spent ~10 minutes thinking about why example shows `onPress` prop and there is nothing like this in the props list, till I scrolled down and finally there was the information about inheritance

## Who needs this?
I think the new learners mostly 

## When should this happen (use version numbers if needed)?
I used `0.63` docs version

Simek:
Hi @SurmaAa, 

the inherited props are already visible in the `next` version docs. The changes has been implemented in #2168. Hope this improves the readability and fixes the issue you are mentioning.

Note: In the previous version of websites clicking on the link in ToC was redirecting users strait to the parent components page, where in new version Toc links scroll to the section where link to the parent component props is located. This has been reported to the Docusaurus and will be fixed in the future.

Refs: #2190

SurmaAa:
Thanks, cool

Issue #2190, opened 2020-09-01
Docs: Include "Inherits from <Component>" in the props section of the right-hand sidebar


DZuz14:
When visting a link to a components documentation (ie: https://reactnative.dev/docs/flatlist), it is not apparent that a component inherits from another unless you specifically click the "props" link in the right sidebar. Atleast on desktop view, I feel as though a lot of people may not click the "props" link at all, and just instead scan the list of props that are already visible in the menu.

I ran into the problem where I would just look at the props, and think, "oh, well it looks like this component doesn't support that prop". This is especially misleading when you consider ScrollView and Flatlist. For newcomers to React Native, there is a clear distinction that you should use one or the other, and one would assume that these components are completely distinct. However this is not the case, as FlatList inherits props from ScrollView.

 I wasted a lot of time searching on Google to see if `stickyHeaderIndices` prop was deprecated in FlatList, because I didn't see it listed by glancing at the list of props in the right sidebar. It wasn't until I randomly clicked `props` that I found out it inherited props from ScrollView.

It might be nice to add a note in the sidebar that it inherits props, with a link to that component perhaps. Not sure how it should look, but I'd be surprised if I was the only human on Earth that made this same trivial, but easy to make mistake.

P.S. I included a screenshot of where I think it should go. Section is outlined in red in using the glorious Microsoft Paint.

![react-props-list](https://user-images.githubusercontent.com/19492185/91827013-50f1ad80-ec0c-11ea-8c27-9c8d54ab0099.jpg)

Simek:
Refs #2160, #2168

Hi @DZuz14, I think that the issues you are describing has been fixed lately. You can check the `next` version of docs to see the current solution: https://reactnative.dev/docs/next/flatlist

If you have any feedback about the implementation feel free to share it! 🙂 

If the changes in `next` docs are enough to fix the problem, please close this issue.

PR #2463, opened 2021-01-28
Document refreshControl prop

shaebler:
The refreshControl prop on a FlatList is doing its job (same like with ScollScreen), but it is not documented.
I don't know if the refreshControl prop is not officially released yet, or if the doc is missing.

### DIFF
--- a/website/versioned_docs/version-0.63/flatlist.md
+++ b/website/versioned_docs/version-0.63/flatlist.md
@@ -433,7 +433,7 @@ How far from the end (in units of visible length of the list) the bottom edge of
 () => void
 ```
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @shaebler, 

the `refreshControl` is a `ScrollList` property which is inherited by the `FlatList` component. We do not include those props on the list to avoid massive amount of duplications. 

Lately this issue have been brought up a few times (#2190, #2399) and we decided to highlight the parent component props in the sidebar to make it more obvious, that certain component inherits props of other.

You can see the example of this improvement in `next` version of the docs - https://reactnative.dev/docs/next/flatlist#props

PR #2467, opened 2021-02-01
Update flatlist.md and virtualizedlist.md

Ejazshaikh:
FlatList and VirtualizedList's default keyExtractor now checks item.id and item.key as of v60.0

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -389,7 +389,7 @@ Reverses the direction of scroll. Uses scale transforms of `-1`.
 (item: object, index: number) => string;
 ```
 
-Used to extract a unique key for a given item at the specified index. Key is used for caching and as the react key to track item re-ordering. The default extractor checks `item.key`, then falls back to using the index, like React does.
+Used to extract a unique key for a given item at the specified index. Key is used for caching and as the react key to track item re-ordering. The default extractor checks `item.key` and `item.id`, then falls back to using the index, like React does.
 
 | Type     |
 | -------- |
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -389,7 +389,7 @@ Reverses the direction of scroll. Uses scale transforms of `-1`.
 (item: object, index: number) => string;
 ```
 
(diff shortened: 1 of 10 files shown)


Simek:
Hi @Ejazshaikh, can you provide the link to the React Native code, that shows the `keyExtractor` keys order or the change which introduced the additional `id` check?

At this moment, at least the description of `keyExtractor` do not mention the `id` usage:
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/FlatList.js#L121

Ejazshaikh:
Hi @Simek, please refer to this commit in react native which was added in react-native version v0.60.0
https://github.com/facebook/react-native/commit/de0d7cfb79c7f4011d4b6748b1afc656d33fd5ac

Simek:
Thank you for the linking the commit. Indeed the VirtualizedList default `keyExtractor` prop includes the check for `id`:
* https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js#L567

As you noted, this feature have been introduced with 0.60 release, which means that all the versioned docs should also be updated. Would you like to update the PR?

docs/flatlist.md
```diff
@@ -389,7 +389,7 @@ Reverses the direction of scroll. Uses scale transforms of `-1`.
 (item: object, index: number) => string;
 ```
 
-Used to extract a unique key for a given item at the specified index. Key is used for caching and as the react key to track item re-ordering. The default extractor checks `item.key`, then falls back to using the index, like React does.
+Used to extract a unique key for a given item at the specified index. Key is used for caching and as the react key to track item re-ordering. The default extractor checks `item.key` and `item.id`, then falls back to using the index, like React does.
```
Simek, line 392: Also please rephrase the enumeration sentence to:
> [...] checks `item.key`, than `item.id`, and then falls back [...]

Ejazshaikh:
Done

Simek (approved):
Thank you so much for working on updates, good job! 👍

PR #2469, opened 2021-02-02
Update slider.md

Sharcoux:
See [this](https://github.com/react-native-community/discussions-and-proposals/pull/329#issuecomment-771527838) for explanation

### DIFF
--- a/website/versioned_docs/version-0.63/slider.md
+++ b/website/versioned_docs/version-0.63/slider.md
@@ -3,7 +3,7 @@ id: slider
 title: '🚧 Slider'
 ---
 
-> **Deprecated.** Use [@react-native-community/slider](https://github.com/react-native-community/react-native-slider) instead.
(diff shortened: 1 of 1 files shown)


Simek:
Hi @Sharcoux, thank you for sending this PR, but we only link the NPM packages which were extracted directly from React Native and are maintained by the React Native Community. 

Otherwise we will need to link multiple packages for each component to avoid favorizing any of packages.

Sharcoux:
@Simek If you read [this](https://github.com/react-native-community/discussions-and-proposals/blob/master/partners/0001-organization-repository-policy.md#react-native-community-will-transition-towards-hosting-only-the-repositories-that-fit-into-category-3) you will see that the issue doing that is that it prevent better solutions to rise. And especially talking about this specific repository, the lib is not anymore in active development and bugs report and feature request are accumulating and not being treated. Keeping it as an "official" reference prevent the RN ecosystem to equip itself with better tools.

Issue #2454, opened 2021-01-24
PlatformColor dcos example not running on android in example


MananDesai54:
## Description

I tried to run the example on android on the docs itself but it is crashing the expo app, running fine on ios
![Screenshot (563)](https://user-images.githubusercontent.com/52067783/105625391-fffa8400-5e4e-11eb-9de6-a17395774776.png)

## Documentation version
0.63

luism3861:
@MananDesai54  thanks for the issue!.

PR #2472, opened 2021-02-02
fix example in PlatformColor

luism3861:
this PR update android example in `PlatformColor` when you run in android device app crash, i fix that according this closed PR https://github.com/facebook/react-native/pull/27908, reference this issue #2454


Simek (approved):
Hi @luism3861, thank you for working on the fix for that issue! 👍

PR #2474, opened 2021-02-03
Support dependencies on Snack player

IjzerenHein:
# Why

This PR add support for additional Snack dependencies to the remark-snack-player. This allows for fixing missing `expo-constants` warnings and makes the examples future proof, as Snack moves toward a direction where dependencies are fully optional.

image

Secondly, this PR makes it possible to support dependencies that are not loaded by default in Snack, such as `@react-native-community/slider`, which replaces the default `Slider` component. This in turn fixes web preview for all examples that use the `Slider` component, which doesn't work on web. Migrating these examples to `@react-native-community/slider` fixes the web support problem allows for showing the preview in the browser directly.

image

# How

- Add support for `&dependencies=...` query param
- Updated snack player docs
- Updated snack player tests and added dependencies
- Per example updated the "Text style" example, so it also runs on web

# Test plan

- Tests were updated/extended for the remark snack player
- Updated text style example

image

https://snack.expo.io/@hein_expo/textstyleprops


Simek (approved):
Hi @IjzerenHein,

Thank you for adding support for the external dependencies! That is indeed, a nice feature. 👍 

According to your suggestion for Slider, we aim to document as exact as possible state and functionality of build APIs or components in RN, so we probably wouldn't utilize the feature that much.

Also the `expo-constants` are scheduled for removal from the examples, because from the analytics data and feedback we got, it looks like utilizing this package confuses some of users, mainly the newcomers, who just started learning about React Native, or people who have not heard/read about Expo yet.

Hope I was able to explain few things on our side, and just want to emphasize that it's a great contribution, thank you! 🙂

IjzerenHein:
Hi @Simek! I understand what you are saying and agree with keeping examples to a bare minimum. Was just trying to contribute in making sure stuff work on all platforms. And totally understand that web is outside the maintenance area of this repo. Although it is featured on the introduction page, which makes you wonder about its importance and usefulness. Personally I find the ability to target web as well an amazing feature and as it matures this will only grow.  No worries for deleting the other PRs, they were only a couple mins work. Cheers m8 👍

PR #2476, opened 2021-02-03
Support Snack web-preview for Shadow Props example

IjzerenHein:
# This should be merged after #2474

Follow up to #2474

This add support for Snack web preview to the Shadow Props example

image

# How

- Use `@react-native-community/slider` which supports web
- Enable web platform for example


Simek:
Hi @IjzerenHein,

as stated in https://github.com/facebook/react-native-website/pull/2474#pullrequestreview-584289041, the goal of React Native documentation is to describe the build-in APIs and components as close, and as detailed, as possible.

The `web` preview is just a helper tool in the website case, nice thing to have, but if there are any problems or issue with this preview, it should be disabled.

Besides that, `web` is an out-of-tree platform , it's developed independently of React Native. It would be hard to handle all the discrepancies and minor issues in the examples. Users looking for the detailed documentation and examples should visit the platform docs, available at:
* https://necolas.github.io/react-native-web/docs/

It's sad, but I had to close this PR too, but thank you for the work you put in in there.

IjzerenHein:
@Simek Coming back to this issue. As using an out of branch package is a no go, what is the way forward for these examples? The RN `Slider` API is considered deprecated, but it is still used in some examples.

Simek:
@IjzerenHein There are several APIs which are deprecated for the few releases, but cannot be removed from the core right now due FB internal usage of this libraries, and it's nothing more we can do in this matter:
* https://github.com/facebook/react-native/issues/23313

Generally users which have the issues with `Slider` or want an extended functionality should use one of community packages (as the note on the `Slider` page says.

The docs will use `Slider` in examples until it's a part core, and when it will be removed examples should be rewritten (to use other core APIs, the community/3rd part packages should not be a part of the documentation.

IjzerenHein:
@Simek Okay noted, thanks for the reply. As the web-preview is currently not working because of the Slider issue, should we disable the web-platform for these examples?

Simek:
> @Simek Okay noted, thanks for the reply. As the web-preview is currently not working because of the Slider issue, should we disable the web-platform for these examples?

Yup, if there are issues on the web, I think that disabling web examples it's the best what we can do right now, sice web is an out-of-tree platform.

IjzerenHein:
Okay sound good. I'll create a PR for it later this week 👍

IjzerenHein:
See https://github.com/facebook/react-native-website/pull/2656

Issue #2056, opened 2020-07-16
Clarify documentation about when onLayout is called


evelant:
## Description

https://reactnative.dev/docs/view#onlayout

Trying to animate elements in my app to the screen locations of other elements. I get the position of those elements in their onLayout callbacks and cache the results. In some (unknown, undocumented) situations onLayout is not called when an element changes position.

The current documentation states

> Invoked on mount and layout changes with:
>
> {nativeEvent: { layout: {x, y, width, height}}}
> 
> This event is fired immediately once the layout has been calculated, but the new layout may not yet be reflected on the screen at the time the event is received, especially if a layout animation is in progress.

This does not specify that in some situations the layout information will be incorrect or transitional. It also does not specify in which situations `onLayout` will *not* be called. In my case it appears that `onLayout` is only called on mount but not when a parent changing size causes an element to change x/y position but remain the same size.

It would be helpful if the documentation for onLayout were expanded to detail exactly in which situations onLayout will be called and in which situations the position/size information may be inaccurate. What exactly constitutes `layout changes` to fire the event?

edit: here are some related issues detailing the undocumented quirks/bugs with onLayout.
https://github.com/facebook/react-native/issues/28775 (animations of transforms don't trigger onLayout)
https://github.com/facebook/react-native/issues/3282 (measure returns undefined on android unless onLayout or collapsable are specified)
https://github.com/facebook/react-native/issues/24219 (reanimated, standard with expo, changes onLayout behavior)
https://github.com/facebook/react-native/issues/23443 (device rotation doesn't trigger onLayout in some situations)
https://github.com/facebook/react-native/issues/19872 (onLayout removed elevation on Android)

## Documentation version

v0.63

olosegres:
@stale Issue does not fixed

dennisbouwpas:
@Stale Issue does not fixed!

chrisbobbe:
I also think an answer to the following would be great to document: are `x` and `y` relative to the top-left of the whole screen, or to the parent? I [believe](https://github.com/facebook/yoga/issues/787) they're relative to the parent, but I don't want to start a false rumor. 🙂

lon9man:
+1

audrew:
I am trying to find a solution too.  Hopefully, that can help https://reactnative.dev/docs/direct-manipulation

J6ey:
[...]

PR #2478, opened 2021-02-04
KeyboardAvoidingView: Describe `keyboardVerticalOffset` more precisely.

chrisbobbe:
A common use case is to pass the height of a header, if the
`KeyboardAvoidingView` is just below the header without including
it. So, help make it easier to decide whether or not to use this
prop by clarifying what it does. When I read "the react native view"
and "some use cases", I had a vague idea that it probably wouldn't
help in my case, and that it was something I'd reach for if I was
using React Native for parts of my app and not others.

But it turned out to be quite useful for my use case!

### DIFF
--- a/website/versioned_docs/version-0.63/keyboardavoidingview.md
+++ b/website/versioned_docs/version-0.63/keyboardavoidingview.md
@@ -100,7 +100,7 @@ Enabled or disabled KeyboardAvoidingView. The default is `true`.
 
 ### `keyboardVerticalOffset`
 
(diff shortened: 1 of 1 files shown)


chrisbobbe:
My company has already signed the CLA; please see and reply to my comment at https://github.com/facebook/react-native/pull/29926#issuecomment-690794439.

chrisbobbe:
Hmm, something has come up and I'm not sure if I understand how this works, even now; I'm studying the implementation more closely now. 🙂

chrisbobbe:
https://github.com/facebook/react-native-website/issues/2056#issuecomment-773618381

PR #2481, opened 2021-02-06
Inserting omitted code

gondooley:
With Android back button, there was an announcement that the modal window closed, but the code failed to make it invisible.
Inserted line corrects this.


Simek (approved):
Hi @gondooley , thank you pointing this issue out and preparing PR.

As you can see in the base docs, the issue have been already fixed - https://reactnative.dev/docs/next/modal - but have not been ported to the versioned docs.

Ideally, instead of `false` you should use `!modalVisible` and update the Class Component example too. 

But this time I won't request the changes, because I'm feeling like on both, base and versioned docs pages, examples could be refreshed a bit more, so I will work on this shortly.

PR #2486, opened 2021-02-08
Include Android SDK Command-line Tools

apfritts:
While trying to run Android on my mac, `react-native doctor` kept reporting it couldn't find the Android SDK. After a bit of Googling and cursing, I discovered installing the `Android SDK Command-line Tools` resolved this problem. I didn't see any mention in the documentation so figured I'd propose it to be added. Happy to put this in the troubleshooting section if that would be more appropriate.


Simek (approved):
Hi @apfritts, thank you pointing this out for updating the guide! 👍

PR #2494, opened 2021-02-11
Remove unnecessary `

inokawa:


Simek (approved):
Hi @inokawa, thank you for the fix and porting it into versioned docs too! 👍

PR #2496, opened 2021-02-14
Pressable is recommended over TouchableOpacity / TouchableWithoutFeed…

Kotlia:
According to [this](https://reactnative.dev/docs/pressable) document, Pressable is recommended over TouchableOpacity / TouchableWithoutFeedBack.

### DIFF
--- a/docs/button.md
+++ b/docs/button.md
@@ -5,7 +5,7 @@ title: Button
 
 A basic button component that should render nicely on any platform. Supports a minimal level of customization.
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @Kotlia,

Thank you for working o this PR. In general you are right that since `0.63` release `Pressable` is recommended over `Touchables`, however the link following the sentence leads to the `Button` implementation which still uses Touchables:
* https://github.com/facebook/react-native/blob/master/Libraries/Components/Button.js#L289

In this case I think that it's better to leave the pages as is for now to avoid the confusion. The Touchables should be deprecated within few next major releases, so the Button implementation will need to be rewritten. 

The note at the beginning of each Touchable page have been added to prepare developers for the future deprecation, and give them more time to react to future deprecation, since Touchables are the base components, used in almost all React Native apps.

PR #2509, opened 2021-02-17
Update flexbox.md

eze1376:
just to correct its order


Simek (changes requested):
Hi @eze1376, thank you for proposing the change. On the React Native website we are trying to stick to alphabetical order if possible, which is more accessible to the user. 

Can you reorder the sections instead and revert the header to the alphabetical order? It would be nice if you can also update the base docs file, located in `docs` directory (https://github.com/facebook/react-native-website/blob/master/docs/flexbox.md).

Initially most of the descriptions were copied from Yoga docs (https://yogalayout.com/docs/flex/), which explain the unordinary order.

Simek (approved):
Thank you for the review updates! 👍

eze1376:
You're welcome!

PR #2510, opened 2021-02-17
Update flexbox.md

eze1376:
Update this to correct its order

I update this file only to correct its order


Simek (approved):
Thank you for updating the base docs too! 🙂 

Refs #2509

PR #2516, opened 2021-02-24
update link to profiling with systrace

nobi1007:
When clicking on systrace, it should internally route to `.../profiling#profiling-android-ui-performance-with-systrace` and not on `.../performance#profiling-android-ui-performance-with-systrace` which doesn't even exists.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -5,7 +5,7 @@ title: Profiling
 
 Use the built-in profiler to get detailed information about work done in the JavaScript thread and main thread side-by-side. Access it by selecting Perf Monitor from the Debug menu.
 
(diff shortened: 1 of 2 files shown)


Simek (approved):
Hi @nobi1007, thank you for catching this up and pushing the fix! 👍 

FYI: The general performance page have been split into few smaller parts, a while ago, but the issue you have found looks like the remnant of this refactor.

nobi1007:
@Simek is the CI test failing due to those changes, from the logs it seems we might need to free up memory?

Simek:
> @Simek is the CI test failing due to those changes, from the logs it seems we might need to free up memory?

@nobi1007 Nope, PR is fine, sorry, forgot to merge it before the call :) 

The Circle CI for some time have an issue with Teraser plugin used in the Docusuaurs for the production minifiaction, Docusuaurs devs are aware of the problem and there is a chance that currently ongoing migration to Webpack 5 will solve this issue.

PR #2517, opened 2021-02-25
Update Docs for Ignite now using "MobX-State-Tree" for state management

kfiggins:
Just a small fix I noticed while reading through the docs for "More Resources"


Simek (approved):
Hi @kfiggins, thank you for catching this and the update! 👍 

It looks like someone already have improved the Ignite description in the base docs, so I just have moved the updated paragraph, so the content between versions is consistent.

PR #2521, opened 2021-03-01
Update intro-react-native-components.md

leobastin:
While I was learning it looked to me like Native components and Core components are "siblings" from the way sub headings were coming one after the other. The actual idea of "Native Components" being parent of both "Core Components" and "Community Components" was not coming out very clearly (IMHO). I've just tried to re-arrange the existing paragraphs a bit to convey this idea with as minimum changes as possible.


Simek (approved):
Hi @leobastin, thank you for catching this and working on the improvement! 👍 

I think that initial idea behind putting the sentence at the end was to make a smooth transition into next section (`Core Components`), but can I see the problem and agree with your remark.

Welcome to the community! 🙂

PR #2523, opened 2021-03-02
Upgrade to @docusaurus/*@2.0.0-alpha.70

yungsters:
Upgrades to `@docusaurus/*@2.0.0-alpha.70` so that we can also eliminate our dependency on `prismjs@1.22.0`.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -31,8 +31,8 @@
     "lintv": "cd ../ && alex"
   },
   "dependencies": {
-    "@docusaurus/core": "2.0.0-alpha.66",
-    "@docusaurus/preset-classic": "2.0.0-alpha.66",
+    "@docusaurus/core": "2.0.0-alpha.70",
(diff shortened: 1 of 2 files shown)


Simek:
Refs: #2420

@yungsters Unfortunately Docusaurus upgrade isn't that easy, since website is using swizzled components, for example:
* https://github.com/facebook/react-native-website/blob/master/website/src/theme/DocItem/index.js 

Required changes:
* https://github.com/facebook/react-native-website/pull/2420/files#diff-e7a5c0c8a106ee94cb4e389a5e061103a5318667c34ae24fcd652fe873be42b2)

I have put the upgrade portion inside the referenced PR because it was required to enable PWA. I will extract the upgrade portion from the PR mentioned above, and then bump to the newest release.

Simek:
@yungsters Closing in favor of #2525

PR #2532, opened 2021-03-08
Matched the version and link

Teymurian:


Simek:
Closing because of lack of CLA.

Also minimum Node version supported is still 12 and it's signified that you can also have newer version, so probably the sentence should not be changed. Maybe the link should be, but only alternative for now is https://nodejs.org/en/download/releases/ page, which might add too much confusion for the users.

PR #2533, opened 2021-03-08
height and width are not the same on different screens

jimothyhalpert7:
As I learned here:
https://github.com/facebook/react-native/issues/30716

height and width render significantly differently on different screens, and that's to be expected.

### DIFF
--- a/docs/height-and-width.md
+++ b/docs/height-and-width.md
@@ -32,7 +32,7 @@ const FixedDimensionsBasics = () => {
 export default FixedDimensionsBasics;
 ```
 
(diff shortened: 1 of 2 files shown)


brentvatne:
I don't believe that the intention of this sentence is to indicate that the physical size should be the same, but rather that the size is fixed to a number of points and not calculate based on screen size - eg: `flex: 1` at the root of a view tree would fill the entire screen, whereas `width: 150, height: 150` is independent of the device size and will result in 150 points wide and tall. I would leave this sentence in and add additional clarification about there not being any universal mapping from points to physical units of measurement.

Simek (approved):
Hi @jimothyhalpert7, thank you for working on the clarification for the website! 👍

PR #2539, opened 2021-03-12
attempt to fix share metadata

Simek:
It looks like shared blog post have the same metadata set as the home page:

Screenshot 2021-03-12 205951

The `metadatas` field has been moved from V1 and it might be possible that those entries are messing with Docusaurus logic. Official Docusaurus opted-out those tags, and their shared blog post are looking correct, let's try that.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -293,16 +293,6 @@ module.exports = {
     },
     metadatas: [
       {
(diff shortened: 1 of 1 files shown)


Simek:
It looks like this has fixed the issue 🙂 

Screenshot 2021-03-12 211404

PR #2545, opened 2021-03-15
Change styling to note

jimothyhalpert7:
As part of
#2533

The added information felt like it deserves a "note" styling.


Simek (approved):
Hi @jimothyhalpert7, thank you for the follow up PR! 👍

PR #2546, opened 2021-03-16
Update tutorial.md

HughRdgz:
Changed this.props.{NAME} by {this.props.NAME}

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -59,7 +59,7 @@ So this code is defining `HelloWorldApp`, a new `Component`. When you're buildin
 
 Most components can be customized when they are created, with different parameters. These creation parameters are called props.
 
-Your own components can also use `props`. This lets you make a single component that is used in many different places in your app, with slightly different properties in each place. Refer to `props.{NAME}` in your functional components or `this.props.{NAME}` in your class components. Here's an example:
+Your own components can also use `props`. This lets you make a single component that is used in many different places in your app, with slightly different properties in each place. Refer to `props.YOUR_PROP_NAME` in your functional components or `this.props.YOUR_PROP_NAME` in your class components. Here's an example:
 
 ```SnackPlayer name=Hello%20Props
 import React from 'react';
--- a/website/versioned_docs/version-0.64/tutorial.md
+++ b/website/versioned_docs/version-0.64/tutorial.md
@@ -59,7 +59,7 @@ So this code is defining `HelloWorldApp`, a new `Component`. When you're buildin
 
 Most components can be customized when they are created, with different parameters. These creation parameters are called props.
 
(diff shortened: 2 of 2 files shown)


Simek (commented):
Hi @HughRdgz, can you explain why you want to change this note? Currently inline code in there looks fine to me. `{NAME}` is referring to the user defined prop, used here as placeholder, the rest of chain should remain unchanged no matter which name has been chosen.

HughRdgz:
> Hi @HughRdgz, can you explain why you want to change this note? Currently inline code in there looks fine to me. `{NAME}` is referring to the user defined prop, used here as placeholder, the rest of chain should remain unchanged no matter which name has been chosen.

maybe I got confused by the brackets, since these are used in code for interpolation.
When I read the documentation I thought it was referring to how you should write the variable when you want that it be interpolated. But now that you put your point of view, it makes sense and now I think maybe I'm wrong.

Simek:
> maybe I got confused by the brackets, since these are used in code for interpolation.
> When I read the documentation I thought it was referring to how you should write the variable when you want that it be interpolated. But now that you put your point of view, it makes sense and now I think maybe I'm wrong.

Template literals requires a dollar sign before the curly brackets:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

@HughRdgz Maybe we can replace the brackets with other character, but both parentheses `(NAME)` and square braces `[NAME]` probably will cause more confusion. Do you have any other idea?

HughRdgz:
> > maybe I got confused by the brackets, since these are used in code for interpolation.
> > When I read the documentation I thought it was referring to how you should write the variable when you want that it be interpolated. But now that you put your point of view, it makes sense and now I think maybe I'm wrong.
> 
> Template literals requires a dollar sign before the curly brackets:
> 
> * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
> 
> @HughRdgz Maybe we can replace the brackets with other character, but both parentheses `(NAME)` and square braces `[NAME]` probably will cause more confusion. Do you have any other idea?

What do you thing about angle brackets?

Simek:
> What do you thing about angle brackets?
> It's just an idea

@HughRdgz We can try that, can you update the PR and include change also to the base docs at `/docs/`? 

I will look through the code how often we use this format and if any other notes should be changed too, to preserve the consistency across the website.

Simek:
@HughRdgz It looks like this was the only one place on the website where such placeholder format has been used.

I have also received feedback, that using angle brackets will resemble too much of JSX syntax, so at this moment of time I have changed the placeholder name to more descriptive one, which I hope will also make it more obvious, that it's just a pseudo-code example.

PR #2548, opened 2021-03-18
fixed introduction md Tabs functionality

EkkoKo:
In the getting-started page (introduction.md) the Tabs bar is not working properly -
 when I click on web it opens ios, 
when I click on ios  it opens android
and when I click on android it opens web.
The problem seemed to be that the tabs weren't organized in the same order as the getDevNotesTabs array (which was the values for the Tabs element).


Simek (approved):
Hi @EkkoKo, thank you so much for catching this issue! 👍 

It looks like regression after latest Docusuaurs update (#2547), you can see that on deploy previews before the bump, tabs were working correctly:
* https://deploy-preview-2546--react-native.netlify.app/docs/getting-started#developer-notes

Please, if you can, report the issue in the Docusuaurs repository.

The proposed fix looks fine to me and after applying it the tabs are working correctly, thank you for putting time into debugging and solving this! 🙂

PR #2549, opened 2021-03-18
Update TalkBack download info in accessibility.md

kacieb:
Downloading from the Google Play Store is the best way to get access to TalkBack on emulators.

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -279,10 +279,7 @@ if (Platform.OS === 'android') {
 
 To enable TalkBack, go to the Settings app on your Android device or emulator. Tap Accessibility, then TalkBack. Toggle the "Use service" switch to enable or disable it.
 
-P.S. Android emulator doesn’t have TalkBack by default. To install it:
-
-1. Download TalkBack file here: https://google-talkback.en.uptodown.com/android
-2. Drag the downloaded `.apk` file into the emulator
+Android emulators don't have TalkBack installed by default. You can install TalkBack on your emulator via the Google Play Store.
 
 You can use the volume key shortcut to toggle TalkBack. To turn on the volume key shortcut, go to the Settings app, then Accessibility. At the top, turn on Volume key shortcut.
 
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -279,7 +279,7 @@ if (Platform.OS === 'android') {
 
 To enable TalkBack, go to the Settings app on your Android device or emulator. Tap Accessibility, then TalkBack. Toggle the "Use service" switch to enable or disable it.
 
-Android emulators don't have TalkBack installed by default. You can install TalkBack on your emulator via the Google Play Store.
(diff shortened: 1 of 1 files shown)


kacieb:
cc @blavalla - let me know if there's any other info that would make sense to change here!

blavalla (commented):
Do you think we should clarify that you need to use an emulator image with the play store installed?

Google provides these in Android Studio, but it may not be super obvious to someone who is not used to using different emulators.

Here's a screenshot showing what I mean:

image

kacieb:
> Do you think we should clarify that you need to use an emulator image with the play store installed?

That's a great callout, I didn't know that (and wouldn't have known that from the existing wording on the site!).

Maybe something similar to:
"Make sure to choose an emulator with the Google Play store installed. These are available in Android Studio."

Eventually I'd like to overhaul these whole sections to add more details about testing accessibility.

blavalla:
That sounds good to me! Feel free to include my screenshot if you think it's helpful.

docs/accessibility.md
```diff
@@ -279,10 +279,7 @@ if (Platform.OS === 'android') {
 
 To enable TalkBack, go to the Settings app on your Android device or emulator. Tap Accessibility, then TalkBack. Toggle the "Use service" switch to enable or disable it.
 
-P.S. Android emulator doesn’t have TalkBack by default. To install it:
-
-1. Download TalkBack file here: https://google-talkback.en.uptodown.com/android
-2. Drag the downloaded `.apk` file into the emulator
+Android emulators don't have TalkBack installed by default. You can install TalkBack on your emulator via the Google Play Store. Make sure to choose an emulator with the Google Play store installed. These are available in Android Studio.
```
lunaleaps, line 282: I like the idea of using a screenshot as @blavalla mentioned highlighting the column that shows which emulators have Play store installed. Do you think that makes sense?

kacieb: I think it makes sense! I'm not sure how to add an image to the website though (plus make it mobile friendly and whatnot), I'd have to look into how to do that.

lunaleaps: Yea.. good point. I found this usage of an image on "Handling Touches": https://reactnative.dev/docs/handling-touches
with source here: https://raw.githubusercontent.com/facebook/react-native-website/67f71d97500da48de34bced6b35ceddac07738ab/website/versioned_docs/version-0.64/handling-touches.md

Might be able to copy that... not sure if it's been mobile optimized either or different asset sizes etc. Might be worth asking in the OSS chat if anyone has advice. Feel free to drop and add another time if it's too involved, accepting to unblock!

lunaleaps (approved):

docs/accessibility.md
kacieb: I'll file a task to add this later!

Simek:
@kacieb If possible, please use the "squash" merge strategy, to retain the clear Git history and avoid additional commits 🙂 
109860679-bb88b200-7c5e-11eb-80c3-c0f965ed584b

kacieb:
> If possible, please use the "squash" merge strategy, to retain the clear Git history and avoid additional commits

Thanks so much, I didn't realize I could do that on Github! I'll make sure to do that.

PR #2552, opened 2021-03-23
Clarify note about Hermes + RAM Bundles

davidvmckay:
Devs in the community including myself are getting tripped up because it seems `export BUNDLE_COMMAND=ram-bundle` is merely *redundant* with Hermes enabled -- when in fact, it is incompatible.  To successfully build with Hermes, `ram-bundle` must NOT be specified as the bundle command.

e.g.: https://github.com/fastlane/fastlane/issues/17657#issuecomment-732764675

### DIFF
--- a/docs/ram-bundles-inline-requires.md
+++ b/docs/ram-bundles-inline-requires.md
@@ -96,7 +96,7 @@ project.ext.react = [
 ]
 ```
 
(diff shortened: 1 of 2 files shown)


Simek (approved):
Hi @davidvmckay, thank you for the note clarification and putting reference to the issue! 👍

PR #2553, opened 2021-03-24
Updated the path to the Android Sdk.

mikeendale:


Simek (changes requested):
Hi @mikeendale, can you elaborate why you have proposed this change? The default installation path has been changed? If yes, it would be nice to see the reference.

Simek:
It looks like that SDK path line has already been updated within broader refactor in #2991. Closing this as it is done.

PR #2554, opened 2021-03-25
fix return type

Damoness:


Simek (changes requested):
Hi @Damoness, can you elaborate why you have proposed this change?

Damoness:
> Hi @Damoness, can you elaborate why you have proposed this change?
```js
const onSubmit = async () => {
  try {
    const eventId = await CalendarModule.createCalendarEvent(
      'Party',
      'my house'
    );
    console.log(`Created a new event with id ${eventId}`);
  } catch (e) {
    console.error(e);
  }
};
```
The return type of createCalendarEvent  is a number rather than a array

Simek (approved):
@Damoness Thank you for the clarification! 👍

PR #2556, opened 2021-03-25
chore: add nativeID to image.md

nelsonprsousa:
Add nativeID prop to Image component, missing from the docs.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -209,6 +209,12 @@ dependencies {
 
 ## Props
 
+### [View Props](view.md#props)
+
+Inherits [View Props](view#props).
+
+---
+
 ### `accessible`
 
 When true, indicates the image is an accessibility element.
--- a/website/versioned_docs/version-0.64/image.md
+++ b/website/versioned_docs/version-0.64/image.md
@@ -209,6 +209,12 @@ dependencies {
 
 ## Props
 
+### [View Props](view.md#props)
+
+Inherits [View Props](view#props).
+
+---
+
 ### `accessible`
 
 When true, indicates the image is an accessibility element.


Simek (changes requested):
Hi @nelsonprsousa, a you can see in the code `nativeID` is not a `Image` component prop, it's inherited from the `View` component:
* https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageProps.js
* https://github.com/facebook/react-native/blob/master/Libraries/Components/View/ViewPropTypes.js#L457-L464

To avoid the large amount of code duplicates we usually placing the inherited props section with the link to the component at the top of the Props list, as seen in here:
* https://reactnative.dev/docs/scrollview#props

This part is definitely missing for the `Image` component right now. If you can, please update the PR code accordingly to the pattern mentioned above. Additionally if you want to see the changes in for example `0.64` docs, you will need to update also the versioned file under `website/versioned_docs`.

nelsonprsousa:
> Hi @nelsonprsousa, a you can see in the code `nativeID` is not a `Image` component prop, it's inherited from the `View` component:
> 
> * https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageProps.js
> * https://github.com/facebook/react-native/blob/master/Libraries/Components/View/ViewPropTypes.js#L457-L464
> 
> To avoid the large amount of code duplicates we usually placing the inherited props section with the link to the component at the top of the Props list, as seen in here:
> 
> * https://reactnative.dev/docs/scrollview#props
> 
> This part is definitely missing for the `Image` component right now. If you can, please update the PR code accordingly to the pattern mentioned above. Additionally if you want to see the changes in for example `0.64` docs, you will need to update also the versioned file under `website/versioned_docs`.

Hi @Simek ,
Just to be sure, you want me to remove the nativeID prop from Image and add this section to the Image.md just like the scrollview:
image

Right?

Simek:
Yes, that's what I'm asking for! 🙂

nelsonprsousa:
> Yes, that's what I'm asking for! 🙂

Done. Can you please re-check? 🙏 

Also, your opinion on [this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51974) is also appreciated. Do you think I should change something here? It is related to this change.

Simek:
> Also, your opinion on [this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51974) is also appreciated. Do you think I should change something here? It is related to this change.

I think that all inherited props should be defined similar to the RN codebase, so the component props should be extended by parent props, like for example here:
* https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L488

Since the DT file is quite large the links to the specific line don't work that good, there is the code I'm talking about:
```ts
export interface PressableProps extends AccessibilityProps, Omit<ViewProps, 'style' | 'hitSlop'>
```

Like in the example above, `Image` component props should be extended by `View` props omitting the `style`. This props diff also can be seen in the Flow annotation in the RN code:
* https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageProps.js#L59

Simek (approved):
Thank you for the review updates! 👍

PR #2558, opened 2021-03-26
docs: update second Flatlist example.

luism3861:
this PR update second example in Flatlist i think it's better to change the text color if the user clicks and change the backgroundColor!.

Screen Shot 2021-03-25 at 23 19 07


Simek (approved):
Hi @luism3861, that was a good idea, the example looks better now, thank you for the update! 👍

PR #2568, opened 2021-03-31
App variable is not declared

rishikesh2003:
while running this code on expo it says "Can't find variable: App" . So, I declared it using the const keyword and exported it default below the function

### DIFF
--- a/docs/using-a-scrollview.md
+++ b/docs/using-a-scrollview.md
@@ -17,7 +17,7 @@ const logo = {
   height: 64
 };
 
-export default App = () => (
+const App = () => (
   <ScrollView>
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hi @rishikesh2003, thank you for the fix! 👍 

Most of those cases were already fixed, but it looks like we have miss this one in the earlier pass. 🙂

rishikesh2003:
Why the changes still didn't appear in the website. I'm guessing it's the problem in circleci?

Simek:
> Why the changes still didn't appear in the website. I'm guessing it's the problem in circleci?

@rishikesh2003 The changes were applied only to the `next` docs, if you would like to see the changes in older, versioned docs you will need to update more files. You can read more about the versioning system in Readme:

PR #2569, opened 2021-03-31
feat(accessibility): Added topic about accessibility settings

grgr-dkrk:
There is no way to get system settings for accessibility in the [accessibility documentation]https://reactnative.dev/docs/accessibility) (ex: Font Scale, Dark Mode, or others).
I added those supplements.


TheSavior:
cc @amarlette

PR #2570, opened 2021-03-31
improve CI checks, ditch `test` in favor of Netlify, fix Alex warnings

Simek:
This PR aims to improve the general CI experience. For months we have a problem with `test` workflow which is failing randomly on Circle CI due to memory related issues, which were confusing the PR authors. 

After migrating deployment to Netlify and PR preview deploys have added the Circle CI `test` seems redundant, because `test` is executing the same command - `yarn build` - which is used to prepare deployment files on Netlify. The Netlify logs are public so if the build fail the PR author should be able to check the the problem.

Additionality I have added the Prettier check and extended the scope of Alex check to the Circle CI `lint` workflow. This required some minor changes and ignores in the newest blog post (2020+), but after those changes all the checks should pass.


.circleci/config.yml
```diff
@@ -53,27 +53,27 @@ jobs:
...
-          name: Compile website
-          command: yarn test
-          no_output_timeout: 20m
-      - run:
-          name: Check for missing index.html (build errors)
-          command: |
-            if [ ! -f build/index.html ]; then
-              exit 1;
-            fi
+#  test:
```
TheSavior, line 56: Why comment it out instead of deleting it?

website/blog/2020-03-26-version-0.62.md
```diff
@@ -10,6 +10,8 @@ tags: [announcement, release]
 
 Today we’re releasing React Native version 0.62 which includes support for Flipper by default.
 
+<!--alex ignore master-->
```
TheSavior, line 13: Should we just be changing `master` everywhere to something else? Has that conversation happened elsewhere?

.circleci/config.yml
Simek: I have commented out the `test` workflow temporarily for testing, will remove the code before the merge, thank for pointing this out! :)

website/blog/2020-03-26-version-0.62.md
Simek: `master` is there is a reference to the repository branch. Maybe we can change it to `main` or something else, but I don't have sufficient permission for that operation. AFAIK there was no conversation earlier on this topic in website repository.

TheSavior (approved):

PR #2571, opened 2021-04-01
FIX: added StatusBar import

rbareja72:
added import of status bar to to avoid crashing on expo


Simek (approved):
Hi @rbareja72, thank you for catching this up and pushing the fix! 👍

PR #2584, opened 2021-04-07
docs: fix inaccurate example

solidw:
The `WelcomeScreen` returns nothing.
image

Use '(' instead of '{' or add "return (" and ")" seems better.


Simek (approved):
Hi @solidw, thank you for catching this typo and pushing the fix! 👍

PR #2586, opened 2021-04-09
Update the documentation of ActionSheetIOS

evrimfeyyaz:
Mention that the `destructiveButtonIndex` option takes either a number or an array of numbers.


Simek (approved):
Hi @evrimfeyyaz, good catch, thank you for the update! 👍

Issue #2588, opened 2021-04-15
Documentation of `requireNativeComponent` is out-of-date for iOS


TheAlmightyBob:
## Description

https://github.com/facebook/react-native/commit/8dc3ba0444c94d9bbb66295b5af885bff9b9cd34 changed `requireNativeComponent` back around React Native 0.56-0.57 to take 1 argument instead of 3. As part of this, https://github.com/facebook/react-native-website/pull/464 updated the documentation for Android, however the iOS documentation on https://reactnative.dev/docs/native-components-ios#properties was left untouched and still shows `requireNativeComponent` as taking 3 arguments.

## Documentation version

All versions (including `next`) since ~0.56 (though I'm guessing archived docs aren't likely to be updated).

TheAlmightyBob:
The fix was merged and I see it updated on the website... not sure why this didn't auto-close?

PR #2589, opened 2021-04-16
Update references to requireNativeComponent in iOS Native Components doc

TheAlmightyBob:
`requireNativeComponent` has not accepted 3 arguments since React Native 0.56. The Android documentation was updated when that changed (https://github.com/facebook/react-native-website/pull/464), but iOS was missed.

Partially addresses #2588... the versioned docs should also be updated, but since the change would be identical I thought I'd just put up the one file for review first.

The use of comments to describe properties was based on https://reactnative.dev/docs/native-components-android... personally I use Typescript so I'm not sure if there's a standard pattern for that in pure JS.

### DIFF
--- a/docs/native-components-ios.md
+++ b/docs/native-components-ios.md
@@ -315,7 +315,7 @@ class MapView extends React.Component {
 
     // process raw event...
     this.props.onRegionChange(event.nativeEvent);
-  }
+  };
   render() {
     return (
       <RNTMap
@@ -338,7 +338,7 @@ class MyApp extends React.Component {
       latitude: 37.48,
       longitude: -122.16,
       latitudeDelta: 0.1,
-      longitudeDelta: 0.1,
+      longitudeDelta: 0.1
     };
     return (
       <MapView
--- a/docs/native-components-ios.md
+++ b/docs/native-components-ios.md
@@ -95,22 +95,27 @@ Now to actually disable zooming, we set the property in JS:
 <MapView zoomEnabled={false} style={{ flex: 1 }} />
 ```
 
-It is recommended for you to document the component interface in this module (e.g. using Flow, TypeScript, or plain old comments). Here we'll use comments:
+To document the properties (and which values they accept) of our MapView component we'll add a wrapper component and document the interface with React `PropTypes`:
 
 ```jsx
 // MapView.js
+import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


TheAlmightyBob:
Interesting. The things the linting job complained about are not things I touched... I'm guessing the rules changed since this file was last updated.

Simek (changes requested):
Hi @TheAlmightyBob, thank you pointing this problem out and working on the fix. 👍 

However I don't like the comments refactor (but I understand the idea behind). Unfortunately it  currently looks a bit messy and comments are detached from the components and parts their are describing.

Can you convert the code snippets to Flow or eventually TS in this guide, so the example will use some wide-known format?

If you do not want to work on this task, please revert the comment changes and leave only the `requireNativeComponent` fix in this PR.

TheAlmightyBob:
I'm not familiar with Flow, but could convert to TS. Since the Android docs just used comments I had assumed that was preferred and was just trying to be consistent.

I figured removing `PropTypes` was part of the `requireNativeComponent` change since the existing documentation says to use `PropTypes` specifically for passing to `requireNativeComponent` (and since the Android documentation had removed `PropTypes`)... if you think it makes sense to keep `PropTypes` purely as an alternative to comments for visually documenting the interface I'd be happy to keep this change small.

Simek:
@TheAlmightyBob Maybe just leave the `PropTypes` in, for now. 😉 

I will investigate the other options and I will consult with the team on that. Also it looks like Android page is this mentioning `PropTypes` too and it's using them in one of example code:

Screenshot 2021-04-23 111336

docs/native-components-ios.md
```diff
@@ -234,15 +234,7 @@ render() {
 }
 ```
 
-Here you can see that the shape of the region is explicit in the JS documentation - ideally we could codegen some of this stuff, but that's not happening yet.
```
TheAlmightyBob, line 237: The codegen comment I removed just because it seemed irrelevant to the topic, but I admit it's also irrelevant to the `requireNativeComponent` change so could bring it back if you wish

TheAlmightyBob:
Ah, you are right, not _all_ of the PropTypes references were purged from the Android doc...

Simek (approved):
LGTM, thank you for all of the review updates! 👍

PR #2590, opened 2021-04-16
Update ANDROID_HOME to ANDROID_SDK_ROOT

sonicdoe:
[From the Android Studio user guide](https://developer.android.com/studio/command-line/variables):

> `ANDROID_HOME`, which also points to the SDK installation directory, is deprecated.


Simek (changes requested):
Hi @sonicdoe, thank you for reporting this out and improving asset, however it might be a bit early call, since as stated in the linked docs:
> If `ANDROID_HOME` is defined and contains a valid SDK installation, its value is used instead of the value in `ANDROID_SDK_ROOT`.

Which might create other problems or confusion for the users who migrated to the new name. I will request review from others, so we can figure out if it's a good time to update this.

Please also update the Android guide for macOS:
* https://github.com/facebook/react-native-website/blob/master/docs/_getting-started-macos-android.md#3-configure-the-android_home-environment-variable

sonicdoe:
I was surprised about that rule as well. It makes it sound like `ANDROID_SDK_ROOT` is the variable being deprecated.

> Please also update the Android guide for macOS

It’s already updated, [see this diff](https://github.com/facebook/react-native-website/pull/2590/files#diff-f9abd017094048e4a9e4752e397e4c34b330607fc82f56106192d602eef7c371).

Simek:
Hello @sonicdoe, sorry, but I have lost a track of this PR for some reason. 😅 

It look like the path changes have already been applied within #2991, so I think we can close this PR now.

If you seeing anything incorrect or missing in the parts after the refactor please post the new PR with the suggestions!

leotm:
thanks too @sonicdoe btw i'm seeing _Last updated 2022-03-23 UTC_ `ANDROID_SDK_ROOT` is deprecated in favour of `ANDROID_HOME `

> [From the Android Studio user guide](https://developer.android.com/studio/command-line/variables)

Capture
 
Capture

so i think Google have made a u-turn 20 days after
- https://github.com/facebook/react-native-website/pull/2991 (Mar 3) cc @henryhobhouse @cortinico

@Simek i'm happy to create a new issue and PR top revert if i'm seeing straight 😅

(i'm seeing both variables remain on latest GitHub Actions Virtual Environments, so as mentioned both variables will be checked by AGP for consistency anway)

cortinico:
> i'm happy to create a new issue and PR top revert if i'm seeing straight 😅

Yup you're right. We should update the docs to use `ANDROID_HOME` instead of `ANDROID_SDK_ROOT` now to follow the official guidance

PR #2592, opened 2021-04-20
Fix broken mozilla.org link in timers.md

kidroca:
The link would lead to a "Page not found" as the content have been archived
Updated the link to point to the archive

Maybe this should be redacted or point to a more up to date example
As a momentary fix it would at least show the original content behind this link

### DIFF
--- a/docs/timers.md
+++ b/docs/timers.md
@@ -3,7 +3,7 @@ id: timers
 title: Timers
 ---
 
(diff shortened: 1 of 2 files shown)


Simek (approved):
Hi @kidroca, thank you for catching this broken link and pushing an update! 👍 

I have updated the link, so it no longer points to the archived and unmaintained article.

PR #2593, opened 2021-04-21
Update ts-playground url

PiotrNap:
I've updated the playground url, there was an extra closing bracket at the end of the `Hello` component.


Simek (commented):
Hi @PiotrNap, thank you so much for reporting this out! 👍 

Unfortunately there are few more issues with the example itself (even a runtime error) and TS playground code. So instead of merging this PR I will push soon another one fixes for example and updated playground URLs.

Simek:
Closing in favor of #2594.

PR #2597, opened 2021-04-23
Update alert.md

Vishnukr1101:
Button alignment updated

PR #2598, opened 2021-04-23
Update alert.md

Vishnukr1101:
Stylesheet added


Simek (changes requested):
Please add the fix to the main `docs` too, as you have done in the previous PR (#2597).

Simek (approved):
Since you have restored the previous PR, the requested change is no longer valid. 

Thank you for backporting the fix! 👍

PR #2600, opened 2021-04-25
Accessibility improvement

jcabak:
Text is more accessible (correct contrast ratio - [WCAG](https://www.w3.org/TR/WCAG21/#contrast-minimum))


Simek (approved):
Hi @jcabak, thank you for an accessibility contrast updates, LGTM! 👍

PR #2605, opened 2021-05-03
Fix stickySectionHeadersEnabled prop default value

gabrieldonadel:
Fix `stickySectionHeadersEnabled` SectionList prop default value to true


gabrieldonadel:
Hey @Simek could you please review this PR?

Simek (changes requested):
Hi @gabrieldonadel, can you provide link to the code responsible for setting default value for this prop or commit/PR changing the default value?

I would be nice if you can also check from which version of React Native the change has been included. This would help determining how far backported the change need to be.

gabrieldonadel:
@Simek seems that `true` has been the default value on iOS for a long time.
I was able to trace back this 4 years old commit that started it all https://github.com/facebook/react-native/commit/72670bf8d2b1556c2e8e5381477b652a1deaeda2#diff-e01eb4abc822caf23aacfafa7a3fc98bfcd4f61fe34f1def4d8d6d1467b8e969R132

Simek:
@gabrieldonadel Ah, got you. The platform specific default has been defined in the props description, but you are referring to the table missing this information. In cases like this, when value differs between platforms, but prop is used on both the platform labels are used, like for example on the `ActivityIndicator` page:
* https://reactnative.dev/docs/activityindicator#color

It would be nice if you can update the PR accordingly. Also looks like the problem affect all of the versioned docs too, so it would be nice if you can update those files too:

gabrieldonadel:
@Simek I've just fixed all docs including the versioned ones to use the same syntax as the  `ActivityIndicator` page


image

website/versioned_docs/version-0.60/sectionlist.md
```diff
@@ -380,9 +380,9 @@ Rendered at the top and bottom of each section (note this is different from `Ite
 
 Makes section headers stick to the top of the screen until the next one pushes it off. Only enabled by default on iOS because that is the platform standard there.
 
-| Type    | Required |
-| ------- | -------- |
-| boolean | No       |
+| Type    | Required                                                                                         |
```
Simek, line 383: @gabrieldonadel Thank you for the review updates! 

```suggestion
| Type    | Default                                                                                          |
```

Between the versions there were some changes to value tables. Please update the header as suggested above, here and in few other files. Other than that changes LGTM! 👍

gabrieldonadel: @Simek fixed 🙂

Simek (approved):
@gabrieldonadel Thank you for all of the review updates and backporting fix to the older docs versions! LGTM 🚀

PR #2606, opened 2021-05-04
Update references to requireNativeComponent in the versioned iOS Native Components docs

TheAlmightyBob:
This is identical to #2589, but now applying the change to the versioned docs (because the corresponding code change was made prior to RN 0.60)

@Simek (since you reviewed the previous PR)


Simek (approved):
Hi @TheAlmightyBob, LGTM, thank you for updating the older docs too! 👍

PR #2609, opened 2021-05-05
Adds info about one more known issue

olosegres:


TheSavior:
The known issue isn't specifically about multiline text inputs or the enter key. The known issue with TextInput is that when using it as a controlled component (by setting `value`), when someone types that will appear immediately and if you change the value from JS, that will update the native control on a different frame, thus flicker.

PR #2616, opened 2021-05-14
Adding survey link for 1% of visitors

rachelnabors:
This adds a little survey to the site based on MDN's. Only 1% of visitors will see it, and it will help us do better :)


Simek:
Hi @rachelnabors, I was able to fix the build and deploy, but I'm not sure if survey will work correctly. 

`SMCX` is available now on `window`, but maybe an additional call need to be added after my refactor. 

Please test and let me know if survey loads now (if you have an ability to enforce survey display or debug it in any other way).

rachelnabors:
I put the percentage of participants for the form from 1% of traffic to 100% of traffic, and it showed up and linked out correctly. Great job on figuring out how to implement this correctly for Docusaurus!

PR #2621, opened 2021-05-19
Update usewindowdimensions.md

ErickMaeda:
Use object destructuring to avoid call twice `useWindowDimensions()`


Simek (approved):
Hi @ErickMaeda, thank you for this small update for the Diemensions hook! 👍

PR #2624, opened 2021-05-21
Update Direct Manipulation docs to mention TextInput's .clear()

henrymoulton:
I noticed that the docs suggest using `setNativeProps` when it's probably simpler to just call `.clear` on a `TextInput` ref.


henrymoulton:
when I say "same approach", I looked at the source code https://github.com/facebook/react-native/blob/0afd71a18d19edef3d6603cabd2672ffde5ba30e/Libraries/Components/TextInput/TextInput.js#L955

docs/direct-manipulation.md
```diff
@@ -143,6 +143,8 @@ const styles = StyleSheet.create({
 export default App;
 ```
 
+You can use the [`.clear`](https://reactnative.dev/docs/textinput#clear) method to clear the TextInput which clears the TextInput using the same approach
```
Simek, line 146: Please use internal link here and add missing dot at the end of sentence.

Simek (commented):
Hi @henrymoulton, thank you for the update, it seems resonable to change the example and refer to the `clear` method instead! 👍 

Please address the comment under the new sentence and fix the issues, besides that changes LGTM.

docs/direct-manipulation.md
Simek, line 146:
```suggestion
You can use the [`clear`](textinput#clear) method to clear the `TextInput` which clears the current input text using the same approach.
```

henrymoulton:
added feedback @Simek

Simek (approved):

PR #2625, opened 2021-05-23
The intent of StyleSheet.create

puranik3:
It is not clear why StyleSheet.create would be required when a plain object could be used in its place. I have referred this https://stackoverflow.com/questions/38886020/what-is-the-point-of-stylesheet-create to add some details.

### DIFF
--- a/docs/style.md
+++ b/docs/style.md
@@ -40,6 +40,7 @@ const styles = StyleSheet.create({
 
 export default LotsOfStyles;
 ```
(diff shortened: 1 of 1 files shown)


TheSavior:
We don't currently do the optimization where styles become an ID, but this pattern lets us make that optimization in the future.

Also, for things like react-native-web, this pattern is used to create actual style sheets.

Simek:
@puranik3 Can you update the added note according to the Eli's comment? 🙂

PR #2633, opened 2021-05-30
Add a note about the chrome debugger causing useColorScheme to always return "light"

benadamstyles:
I discovered this via [this Stack Overflow answer](https://stackoverflow.com/a/61124384/3098651) which points to [this `react-native` PR](https://github.com/facebook/react-native/commit/f7b90336be25b78935549aa140131d4d6d133f7b).


Simek (approved):
Hi @benadamstyles, thank you for spotting this behaviour out and updating the docs accordingly! 👍

PR #2642, opened 2021-06-08
Fix/Value for flexWrap

timjacksonm:
It works on the sandbox but if you copy the code and run it you get an error.
'no-wrap'. Removed hyphen to be 'nowrap'.


Simek (approved):
Hi @timjacksonm, thank you for catching this issue! 👍

PR #2650, opened 2021-06-21
Remove unused failureCallback in example

Atrox:
This is removing the unused and confusing failure callback in the example code. The text above the example is talking about the node callback convention with a single callback for success and error.

### DIFF
--- a/docs/native-modules-android.md
+++ b/docs/native-modules-android.md
@@ -392,7 +392,7 @@ There are two approaches to error handling with callbacks. The first is to follo
 
 ```java
 @ReactMethod
-public void createCalendarEvent(String name, String location, Callback myFailureCallback, Callback callBack) {
+public void createCalendarEvent(String name, String location, Callback callBack) {
     Integer eventId = ...
     callBack.invoke(null, eventId);
 }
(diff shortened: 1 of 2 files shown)


Simek (commented):
Hi @Atrox, this example is a part of bigger section and shows an alternative approach to handling callbacks, as the paragraph above says:

> There are two approaches to error handling with callbacks. The first is to follow Node’s convention and treat the first argument passed to the callback as an error object.

The example located above this one uses the single argument for callback.

Screenshot 2021-06-23 152708

Atrox:
Are you sure? The way I see it is that the first code block is incorrect. The whole parameter is unused in the code block. The next one seems to be correct because both of the parameters get used.

If not, I'm sorry for misunderstanding that part :) Feel free to close the PR.

image

Simek:
Oh, I see, you have edited the wrong file. In the `next` version (which is shown on my screenshot) examples are correct, but latest versioned docs (for `0.64`) includes the incorrect callback. You need to revert the change in base docs and apply the fix to the correct example in:
* https://github.com/facebook/react-native-website/blob/master/website/versioned_docs/version-0.64/native-modules-android.md

There might be also some changes in corresponding paragraphs which also need to be back-ported. If you don't want to work on this changes please close this PR and I will look into this in soon. 🙂

Atrox:
It's a small fix so I can backport it without a problem if you need me to :). Just to clarify though, isn't that example also wrong in the `next` version? See my screenshot.

image

And if I look at my changed file (which is the `next` file, right?) it looks fine:

image

Simek (approved):
LGTM, thank you for the update! 🚀

PR #2651, opened 2021-06-22
Select a template on initializing project

Tejas-Teju:
Added comment to select a template after running "expo init AwesomeProject" at Line 55.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -52,7 +52,7 @@ npm start # you can also use: expo start
 <TabItem value="yarn">
 
 ```shell
-expo init AwesomeProject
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hi @Tejas-Teju, thank you for adding the note, but I'm not sure if it's not too descriptive. 

The `init` command also can be changed to include template as parameter, which might be a bit cleaner solution (wwith explanation below). 

Also, if it's possible, we preferer to add the longer sentences to the paragraph or quoteblock instead of in-code comments. Please take those tip

Simek (changes requested):
Please address the comment above.

Simek:
Hey @Tejas-Teju, `expo init` is no longer a valid command, and we have refactored the getting started guides since that change has been suggested, so I'm going to close this one.

PR #2655, opened 2021-06-24
update Podfile

follyxing:


Simek:
Hi @follyxing, can you provide the description or the problems you are experiencing or explain your reasoning behind those changes?

follyxing:
Since the **0.64.0** version, some module paths have changed, the Podfile file of this document cannot be used.
fix: [#31440](https://github.com/facebook/react-native/issues/31440)

Simek:
@follyxing Hmm, I'm not sure if it's the correct solution to the Pods problem you are experiencing. 

Can you replace the `podfile` in problematic project with the base file content from RN template (including relative pod imports):
* https://github.com/facebook/react-native/blob/master/template/ios/Podfile

I'm curious if this approach fixes the issue. The docs you are editing were updated a few versions ago, and Pods setup have changed a bit since then.

PR #2684, opened 2021-07-11
Remove react-native-dom from list of out-of-tree platforms

vincentriemer:
I haven't worked/maintained react-native-dom for a year or two now and don't think it makes sense to direct people over to the project anymore. This commit removes it from the list.

### DIFF
--- a/docs/out-of-tree-platforms.md
+++ b/docs/out-of-tree-platforms.md
@@ -7,7 +7,6 @@ React Native is not only for Android and iOS - there are community-supported pro
 
 - [React Native Windows](https://github.com/Microsoft/react-native-windows) - React Native support for targeting Microsoft's Universal Windows Platform (UWP).
 - [React Native macOS](https://github.com/microsoft/react-native-macos) - React Native fork targeting macOS and Cocoa.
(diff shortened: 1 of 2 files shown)


Simek (changes requested):
Hi @vincentriemer, thank you for posting the update! 👍 

Can you also remove the `react-native-dom` entry from, at least, the latest versioned docs too? (https://github.com/facebook/react-native-website/blob/master/website/versioned_docs/version-0.64/out-of-tree-platforms.md) The change to the base file in `docs/` only affect the future version of the docs.

TitusEfferian:
hi, can we put `react-native-web` back to the list? `https://necolas.github.io/react-native-web/`
i think this library still have an active maintainer

things like
```
- [React Native Web](https://necolas.github.io/react-native-web/) - React Native for Web is an accessible implementation of React Native's Components and APIs that is interoperable with React DOM.
```

Simek:
Hi @TitusEfferian, that's a good hint, thank you! 🙂 

`react-native-web` is still actively maintained and used by various companies, like Expo or Twitter. I will take care of the addition.

PR #2685, opened 2021-07-12
Create full_height_android.md

sidferreira:
Adds instructions to have the Android App using full device height, even with notch

### DIFF
--- a/docs/full_height_android.md
+++ b/docs/full_height_android.md
@@ -1,0 +1,121 @@
+---
+id: full-height-android
+title: Full height
+---
+
+By default, React Native Android apps will not render your app below the Status Bar or Navigation bar. You still can use `<StatusBar translucent />` to work around it, but may face some issues when rotating.
+For a better experience, follows:
+
+Add to your main activity:
+```java
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    resetWindowConfiguration();
+    super.onCreate(savedInstanceState);
+  }
+
+  @Override
+  protected void onResume() {
+    resetWindowConfiguration();
+    super.onResume();
+  }
+
+  @Override
+  protected void onRestart() {
+    resetWindowConfiguration();
+    super.onRestart();
+  }
+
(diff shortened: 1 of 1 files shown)


sidferreira:
WiP, suggestions welcome

Simek (commented):
Hi @sidferreira, thank you for pushing a proposition of the new page! 👍 

To make it accessible on the website and in the deploy preview you will need to add entry for the page to the sidebar file (`website/sidebars.json`), as describe in the Docusaurus docs: https://docusaurus.io/docs/sidebar#sidebar-object

docs/full_height_android.md
```diff
@@ -0,0 +1,121 @@
+---
+id: full-height-android
+title: Full height
+---
+
+By default, React Native Android apps will not render your app below the Status Bar or Navigation bar. You still can use `<StatusBar translucent />` to work around it, but may face some issues when rotating.
```
satya164, line 6: Imo we should change the default template to do that. Maybe it made sense when supporting older versions of Android which didn't have translucent status bar, but now it's just inconsistent and differs from how it is on iOS.

sidferreira: @satya164 I got your point, but it's complicated. Android and iOS have different rules and must be respected. If we set Translucent by default, we're forcing a behavior. I prefer translucent tho.

satya164: > Android and iOS have different rules and must be respected

While that's true for a lot of platform-specific UX, there's no such rule that apps shouldn't have a translucent statusbar on Android. A lot of apps I use have a translucent statusbar.  Material Design Guidelines make heavy use of translucent StatusBar etc.

> If we set Translucent by default, we're forcing a behavior

We're forcing a behavior no matter which option we choose. It is still possible to change it either way, so I don't think it makes a lot of difference. In addition, navigation libraries like React Navigation already handle it by default so the majority of the time users wouldn't have to think about it - and they get better styling for StatusBar as it takes the color of the header by default without having to do anything special.

PR #2687, opened 2021-07-13
update word from currently to may

robberfree:
Test on iOS 14.5:
Image component support that four border radius properties now.

### DIFF
--- a/docs/images.md
+++ b/docs/images.md
@@ -218,7 +218,7 @@ Note that you must specify some width and height style attributes.
 
 ## iOS Border Radius Styles
 
-Please note that the following corner specific, border radius style properties are currently ignored by iOS's image component:
+Please note that the following corner specific, border radius style properties are may ignored by iOS's image component:
 
 - `borderTopLeftRadius`
 - `borderTopRightRadius`
--- a/docs/images.md
+++ b/docs/images.md
@@ -218,7 +218,7 @@ Note that you must specify some width and height style attributes.
 
 ## iOS Border Radius Styles
 
-Please note that the following corner specific, border radius style properties are may ignored by iOS's image component:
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hi @robberfree, thank you for the update, however I think this will need a little bit more of research and confirmation.

It would be helpful, if you can provide the PR, commit or changelog entry which indicated that support for those properties has been added to iOS (and if so, what are the exact requirements/circumstances to make the border radius work).

robberfree:
@Simek good suggestions. But if you view the `Image style props` document here: https://reactnative.dev/docs/image-style-props#image-border-radius. You can see that the four `border radius` props are supported on **iOS** platform.

Another thing I check the history of the document: https://github.com/facebook/react-native-website/commits/master/docs/images.md. The sentences about `iOS Border Radius Styles` never have changed since first commit.

thanks.

cc: @hramos

Simek:
@robberfree I have in mind the React Native core repository PRs and commits, or official changelog entries, not the docs.

The code should be the source of truth in this case.

docs/images.md
```diff
@@ -218,7 +218,7 @@ Note that you must specify some width and height style attributes.
 
 ## iOS Border Radius Styles
 
-Please note that the following corner specific, border radius style properties are currently ignored by iOS's image component:
+Please note that the following corner specific, border radius style properties are may ignored by iOS's image component:
```
Simek, line 221:
```suggestion
Please note that the following corner specific, border radius style properties might be ignored by iOS's image component:
```

Simek: I'm going to merge the suggestion and land the PR, so it won'y become stale. 😉 

Also the duplicates start to appear:
* https://github.com/facebook/react-native-website/pull/2908

PR #2688, opened 2021-07-13
Add MDN link for aspectRatio

mattleff:
`aspect-ratio` is now a CSS property with decent [support](https://caniuse.com/mdn-css_properties_aspect-ratio) and an [MDN link](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio). I haven't compared the implementation of `aspectRatio` in RN vs. `aspect-ratio` in the CSS spec in detail, so if they diverge we may want to just remove the comment about `aspect-ratio` being non-standard. Let me know and I can update my PR.

### DIFF
--- a/docs/layout-props.md
+++ b/docs/layout-props.md
@@ -203,7 +203,7 @@ export default App;
 
 ### `aspectRatio`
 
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hi @mattleff, thank you for the update, LGTM! 👍

PR #2690, opened 2021-07-14
Update documentation for `TextInput`'s `autoCapitalize` property to c…

charlesbdudley:
…larify that this property merely suggests the capitalization of characters to the virtual keyboard and doesn't do any filtering/allowing/disallowing.


lunaleaps (approved):
Thanks for investigating this and clearing up!

PR #2698, opened 2021-07-21
website: update Dimensions API, improve required labels.

luism3861:
this PR update  `Dimensions API`, improve required label style.

cc: @Simek


Simek (changes requested):
Hi @luism3861, thank you for the update, great start so far! 👍 

Can you add the [`DimensionsValue`](https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Utilities/Dimensions.js#L21-L25) type, in the similar way to for example Alert `Options` (https://reactnative.dev/docs/alert#options-android)?

After that it would be nice to link it in the `addEventListener` description and other places (if there are any).

luism3861:
> Hi @luism3861, thank you for the update, great start so far! 👍
> 
> Can you add the [`DimensionsValue`](https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Utilities/Dimensions.js#L21-L25) type, in the similar way to for example Alert `Options` (https://reactnative.dev/docs/alert#options-android)?
> 
> After that it would be nice to link it in the `addEventListener` description and other places (if there are any).

hi @Simek  thanks for the review, can you take a look for the new requested changes.! thanks

docs/dimensions.md
```diff
@@ -182,6 +182,21 @@ This should only be called from native code by sending the didUpdateDimensions e
...
+| Type   |
+| ------ |
+| object |
+
+**Properties:**
+
+| Name   | Type                                            | Description                                    |
+| ------ | ----------------------------------------------- | ---------------------------------------------- |
+| window | [addEventListener](dimensions#addeventlistener) | Defines Size of the visible Application window |
+| screen | [addEventListener](dimensions#addeventlistener) | Size of the device's screen                    |
```
Simek, line 202: The correct type of those properties is [`DisplayMetrics`](https://github.com/facebook/react-native/blob/main/Libraries/Utilities/NativeDeviceInfo.js#L22:L27) which also need to be added in the "Type Definitions" section.

By replacing the type in listener I have in mind the portion of description, which now can be replaced with link to the `DimensionsValue` type.

Screenshot 2021-07-22 162916

luism3861: hi @Simek  i took this example table https://reactnative.dev/docs/next/appregistry#registry, for add the values `DimensionsValue`, is that correct ?

Simek: Unfortunately not, but it's not a problem. Let check out this branch and help you a bit. 🙂

luism3861: i tried to put 4 types for every name properties like `window` but i'm no sure looks weird.

Simek: What you have done it's fine, there was a bit confusion on the review request. Also those types are not that straight forward. 

To be honest, probably the small Flow typing refactor in React Native won't hunt. I will try too look at this during weekend. In the other hand people should use `useWindowDimensions` hook instead and this API in a long term is deems to be deprecated, so maybe it's not that important.

Currently we are missing the Android specific optional keys for the dimensions payload, but to document this properly, the small refactor in core will be needed. 

The PR should be now ready to merge, thank you for the updates and work on the docs conversion to the new format! 👍

luism3861: you're right i think most of the people use `useWindowDimensions`. thanks for the review i try to add another required labels for the API docs. 😃

Simek (approved):

PR #2701, opened 2021-07-22
Update Snack to .dev domain

IjzerenHein:
# Why

Updates the Snack player and other Snack URLs to use the new `.dev` domain. *The old `.io` domain still works, but requires an additional redirect on the server.

# How

- Update Snack url in docusaurus.config
- Update Snack urls

# Test plan

- `cd website && yarn start`
- Verified that the Snack player loads as expected


Simek (approved):
Hi @IjzerenHein, thank you for the updates, LGTM! 👍

PR #2702, opened 2021-07-22
Update minimal node requirement to 14

IjzerenHein:
Updates the minimal node requirement in README to node 14.

Node 12 no longer works and throws certain errors:

```
 sync-api-docs % yarn sync
yarn run v1.22.10
$ node sync-api-docs ../../react-native
/Users/hein/repos/Expo/react-native-website/sync-api-docs/generateMarkdown.js:85
  if (method?.params[0]?.type?.raw) {
             ^

SyntaxError: Unexpected token '.'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you are looking for the source code of the [React Native Archive website](htt
 ### Prerequisites
 
 1.  [Git](https://git-scm.com/downloads).
-1.  [Node](https://nodejs.org/en/download/) _(version 12 or greater)_.
+1.  [Node](https://nodejs.org/en/download/) _(version 14 or greater)_.
 1.  [Yarn](https://yarnpkg.com/lang/en/docs/install/) _(version 1.5 or greater)_.
 1.  A fork of the repo _(for any contributions)_.
 1.  A clone of the `react-native-website` repo.


Simek (commented):
Hi @IjzerenHein, currently the Node version is synced with core React Native repository:
* https://github.com/facebook/react-native/blob/main/package.json#L10

This is to avoid different environment requirements for the contributors. Also the Node version is listed in the few places across the docs, and changing it in Readme is not enough.

Additionally the broken part you have mention is a separate project in a website workspace which have not been touched or used in months. Maybe @motiz88 can look at that, but I'm not sure if it's enough to justify the bump.

IjzerenHein:
Hey @Simek. Ok I see. Initially I had a problem running `yarn build` on the website as well. After upgrading to node 14 the problem was gone, but after downgrading to node@12 I couldn't reproduce it, so it might have been something else.
If it's just that script that's hardly used then go ahead and close this PR 👍

Simek:
If I remeber correctly, to run the sync script successfully you had to have React Native checkout in a specific place in relation to the website checkout (probably they need to be in the same parent folder). So maybe this issue is not related to Node after all.

However the initial website build issue might be something worth investigating. Do you have a log or an error message from when this had happend?

IjzerenHein:
> If I remember correctly, to run the sync script successfully you had to have React Native checkout in a specific place in relation to the website checkout (probably they need to be in the same parent folder). So maybe this issue is not related to Node after all.

I do think it is. The script runs fine on node@14, after downgrading to node@12 the error re-emerges

> However the initial website build issue might be something worth investigating. Do you have a log or an error message from when this had happened?

Unfortunately I can't retrieve the log, nor reproduce it :/

Simek:
> I do think it is. The script runs fine on node@14, after downgrading to node@12 the error re-emerges

Thank you for the test and confirmation. I will look at this in a free while.

> Unfortunately I can't retrieve the log, nor reproduce it :/

No problem, most important is that it works correctly now. 🙂

Simek:
I'm going to close this PR for now. Hopefully we will see Node 14 bump in core repository soon. 🙂

PR #2708, opened 2021-07-28
website: add required labels TextInput 

luism3861:
this PR add required labels for specific platform, little changes in style examples, and fix broken link.  additional useEffect  `autoFocus` property.

cc: @Simek

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -109,9 +109,9 @@ Inherits [View Props](view.md#props).
 
 Specifies whether fonts should scale to respect Text Size accessibility settings. The default is `true`.
 
-| Type | Required |
-| ---- | -------- |
-| bool | No       |
+| Type | 
+| ---- | 
+| bool |
 
 ---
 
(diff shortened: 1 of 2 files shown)


Simek:
Hi @luism3861, thank you so much for the update, great work! 

One more thing I want to ask you in this PR is to remove the `Platform` and `Required` columns from table (afaik `TextInput` do not have any required field so removing the column without any other changes could be fine, usually required field needs a special label and move to the top of the props list). When we convert page to the labels we usually also remove those columns:

Untitledx

Additionally, if you want, you can add `Default` column, for the props with default value, as seen for example in here:
* https://reactnative.dev/docs/next/refreshcontrol

But this can also be made in the separate PR, so please choose the resolution as it fit you. 🙂

luism3861:
> Hi @luism3861, thank you so much for the update, great work!
> 
> One more thing I want to ask you in this PR is to remove the `Platform` and `Required` columns from table (afaik `TextInput` do not have any required field so removing the column without any other changes could be fine, usually required field needs a special label and move to the top of the props list). When we convert page to the labels we usually also remove those columns:
> 
> Untitledx
> 
> Additionally, if you want, you can add `Default` column, for the props with default value, as seen for example in here:
> 
> * https://reactnative.dev/docs/next/refreshcontrol
> 
> But this can also be made in the separate PR, so please choose the resolution as it fit you. 🙂

hi @Simek you're right i forgot delete `Platform` and `Required` my last commit has this change, about another Default column i try to send a new PR with this :), thanks!.

Simek (approved):
@luism3861 Thank you for the review updates and working on the labels in the first place! 👍

PR #2711, opened 2021-08-04
Use native-stack for getting started with navigation

satya164:
As of React Navigation 6, we recommend using native-stack by default which will have better performance.


WoLewicki (approved):

Simek (approved):
Hi @satya164, thank you for the update on the navigation guide! LGTM 👍

PR #2713, opened 2021-08-04
chore: fix link in showcase

luism3861:
this PR fix link in showcase for React Native open source apps, something weird happen in dev works but in production still infinite red link home page!.

cc: @Simek

### DIFF
--- a/website/src/pages/showcase.js
+++ b/website/src/pages/showcase.js
@@ -99,7 +99,9 @@ const Showcase = () => {
         </a>
         <p>
           A curated list of{' '}
-          <a href="https://github.com/ReactNativeNews/React-Native-Apps/#readme">
+          <a
+            key="demo-apps"
+            href="https://github.com/ReactNativeNews/React-Native-Apps">
             open source React Native apps
           </a>{' '}
           is also being kept by <a href="https://infinite.red">Infinite Red</a>.
--- a/website/src/pages/showcase.js
+++ b/website/src/pages/showcase.js
@@ -99,11 +99,13 @@ const Showcase = () => {
         </a>
         <p>
           A curated list of{' '}
-          <a
-            key="demo-apps"
-            href="https://github.com/ReactNativeNews/React-Native-Apps">
-            open source React Native apps
-          </a>{' '}
+          <span>
+            <a
+              key="demo-apps"
+              href="https://github.com/ReactNativeNews/React-Native-Apps">
+              open source React Native apps
+            </a>
+          </span>{' '}
           is also being kept by <a href="https://infinite.red">Infinite Red</a>.
         </p>
       </div>
--- a/website/src/pages/showcase.js
+++ b/website/src/pages/showcase.js
@@ -99,11 +99,10 @@ const Showcase = () => {
         </a>
         <p>
           A curated list of{' '}
-          <a href="https://github.com/ReactNativeNews/React-Native-Apps/">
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hi @luism3861, unfortunately looking at the test deploy, the slash at the end of URL was not enough to fix the rehydration issue. 

Can you try to add `key` to the problematic link or even to the both links in this paragraph?

luism3861:
> Hi @luism3861, unfortunately looking at the test deploy, the slash at the end of URL was not enough to fix the rehydration issue.
> 
> Can you try to add `key` to the problematic link or even to the both links in this paragraph?

hi @Simek i added readme key for the github link

website/src/pages/showcase.js
```diff
@@ -99,11 +99,10 @@ const Showcase = () => {
         </a>
         <p>
           A curated list of{' '}
-          <a href="https://github.com/ReactNativeNews/React-Native-Apps">
+          <a href="https://github.com/ReactNativeNews/React-Native-Apps/#readme">
```
slorber, line 102:
```suggestion
          <a key="test" href="https://github.com/ReactNativeNews/React-Native-Apps">
```

Try this instead

Simek: As Sebastien suggested, I have different `key` in mind, sorry for not being precise enough!

```suggestion
          <a key="app-list" href="https://github.com/ReactNativeNews/React-Native-Apps/#readme">
```
But please use something meaningful, so we do not need to push another commit and redeploy it, if the addition of `key` works

luism3861: > Try this instead

@slorber  thank you for the suggestions :),

luism3861: > > Try this instead
> 
> @slorber thank you for the suggestion :),

luism3861: > As Sebastien suggested, I have different `key` in mind, sorry for not being precise enough!
> 
> But please use something meaningful, so we do not need to push another commit and redeploy it, if the addition of `key` fixes the issue.
@Simek  no problem if i use `demo-apps` for the `key`? i pushed my last changes

Simek: `demo-apps` is fine, thank you for the updates! 🙂

luism3861: thanks @Simek  !.

website/src/pages/showcase.js
```diff
@@ -99,11 +99,12 @@ const Showcase = () => {
         </a>
         <p>
           A curated list of{' '}
-          <a href="https://github.com/ReactNativeNews/React-Native-Apps">
+          <a
```
slorber, line 102: try to wrap this link in a <span> 😅

website/src/pages/showcase.js
```diff
@@ -99,11 +99,12 @@ const Showcase = () => {
...
+            key="demo-apps"
```
slorber, line 103: doesn't seem to work

try wrapping this link in a `<span>` instead 🤪 generally changing slightly the markup can solve these weird hydration issues

luism3861: thank you so much  :) !.

Simek (approved):
LGTM! @luism3861 thank you for the all attempts and updates, @slorber thank you for the background on the issue and hints for the fixes! Great teamplay 👍

PR #2714, opened 2021-08-05
docs: update `StatusBarIOS` deprecation message

ecreeth:
`StatusBarIOS` has been removed from core in  facebook/react-native#31466

### DIFF
--- a/docs/statusbarios.md
+++ b/docs/statusbarios.md
@@ -3,6 +3,6 @@ id: statusbarios
 title: '🚧 StatusBarIOS'
 ---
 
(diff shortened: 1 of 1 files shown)


lunaleaps:
@ecreeth Should we just delete the file? Since the docs are versioned does it make sense to keep around? cc @Simek if you have thoughts

lunaleaps (commented):
Should we just delete?

ecreeth:
@lunaleaps see additional info in #1045.

lunaleaps (approved):

PR #2734, opened 2021-08-19
fix warning: functions are not valid as ReactChild

swapanil2021:


Simek (approved):
Hi @swapanil2021, thank you for porting the fix to the latest versioned docs! 👍

PR #2745, opened 2021-08-29
Update debugging.md

Huang-Libo:
Correct url.


Simek (approved):
Nice catch, thank you for the fix! 👍

PR #2747, opened 2021-08-30
Update iOS path to use a different port

karanganesan:
- Changed the path to use latest file and folder structure as per react-native@0.65.1


lucalves (approved):

Simek (approved):
Hi @karanganesan, thank you for spotting this out and preparing the update! 👍 

Not sure how this path survived so long in there, it has been changed with the `0.61` release. 😅

PR #2749, opened 2021-08-31
fix external links to React and React Native repositories

Simek:
This small PR fixes the external links and references using the old main branches name of React and React Native repositories.


yungsters:
For the links, I believe you can replace both `master` and `main` with `HEAD`.

yungsters:
But otherwise, this looks good to me!

yungsters (approved):

Simek:
> For the links, I believe you can replace both `master` and `main` with `HEAD`.

@yungsters Yup, you can, but than GitHub resolves to last detached commit, not the actual main branch:

Screenshot 2021-08-31 213319

PR #2753, opened 2021-09-01
Update flexbox.md

Huang-Libo:
doc: https://reactnative.dev/docs/text-style-props#fontweight


Simek (approved):
Nice catch, thank you! 👍 

The reason why it do not cause the problems immediately is that Expo also allow passing number as arguments (because of the support for the Web). 🙂

PR #2754, opened 2021-09-02
Update flexbox.md

Huang-Libo:
The example imports `SafeAreaView` but does not use it


Huang-Libo:
# Before we use `SafeAreaView`

image

# After using `SafeAreaView`

image

Simek (approved):
Hi @Huang-Libo, thank you for catching this and pushing an update! 👍

PR #2770, opened 2021-09-07
Update fast-refresh.md

KwokKwok:
According to the context, I think it should be "constant" here


Simek (approved):
Hi @KwokKwok, nice catch, thank you! 👍

PR #2773, opened 2021-09-07
Clearify type definition of sendIntent parameter

andreasrueb:
The type definition of sendIntent is quite confusing on first glance. Properties as well as possible types of value are separated by comma. Using this notation should be clearer!

x


Simek (approved):
Hi @andreasrueb, thank you for the update! 👍 

Generally we use the separate rows with dedicated description for the different type options in mixed types but this notation makes sense in here. 🙂

PR #2776, opened 2021-09-08
Add a note about where to place the Gradle vars

raquelmsmith:
This bit of clarifying text would have saved me some time as I tried to figure out where to put the variables so that they wouldn't be checked into my repository.

### DIFF
--- a/docs/signed-apk-android.md
+++ b/docs/signed-apk-android.md
@@ -43,7 +43,7 @@ MYAPP_UPLOAD_KEY_PASSWORD=*****
 
 These are going to be global Gradle variables, which we can later use in our Gradle config to sign our app.
 
-> Note about using git: placing the Gradle variables in `~/.gradle/gradle.properties` instead of `android/gradle.properties` prevents them from being checked in to git. You may have to create the `~/.gradle/gradle.properties` file in your computer user root before you can add the variables.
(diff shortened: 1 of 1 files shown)


docs/signed-apk-android.md
```diff
@@ -43,6 +43,8 @@ MYAPP_UPLOAD_KEY_PASSWORD=*****
 
 These are going to be global Gradle variables, which we can later use in our Gradle config to sign our app.
 
+> Note about using git: placing the Gradle variables in `~/.gradle/gradle.properties` instead of `android/gradle.properties` prevents them from being checked in to git. You may have to create the `~/.gradle/gradle.properties` file in your computer user root before you can add the variables.
```
GijsWeterings, line 46:
```suggestion
> Note about using git: Saving the above Gradle variables in `~/.gradle/gradle.properties` instead of `android/gradle.properties` prevents them from being checked in to git. You may have to create the `~/.gradle/gradle.properties` file in your user's home directory before you can add the variables.
```

GijsWeterings (changes requested):
Hi @raquelmsmith! Thanks for your suggestion! I have a couple minor suggestions on the phrasing, just to avoid people skimming and assuming they need `root` privileges on their system for this. The `~` indicates the user's home directory, so I'd say we just refer to that instead

raquelmsmith:
Good idea, thanks! Suggestion applied.

Simek (approved):
Hi @raquelmsmith, thank you for updating the docs, that might be an useful hint for some of the users! 👍

PR #2786, opened 2021-09-20
rename mistyped sample js file

keremoge:
rename **`CalendarModule.js`** to **`NativeCalendarModule.js`** as it accidentally mistyped


lunaleaps (changes requested):
Thanks for fixing @keremoge, can you also update the versioned docs, you'll find it under website/versioned_docs and fix those files?

keremoge:
> Thanks for fixing @keremoge, can you also update the versioned docs, you'll find it under website/versioned_docs and fix those files?

Done 👍

keremoge (commented):
done

PR #2789, opened 2021-09-23
fix: pagingEnabled for vertical scroll is supported in android

irohitb:


Simek:
Hi @irohitb, thank you for the update! 👍 However it would be nice if you can provide any context for this change.

Link to PR, commit or code in React Native core repository would be great! 🙂

PR #2791, opened 2021-09-24
Update native-modules-android.md

simonbrunou:
Those two methods are missing in the documentation and an exception is sent if they are not in the code.

### DIFF
--- a/docs/native-modules-android.md
+++ b/docs/native-modules-android.md
@@ -516,6 +516,15 @@ private void sendEvent(ReactContext reactContext,
      .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
      .emit(eventName, params);
 }
+@ReactMethod
+public void addListener(String eventName) {
+ // Keep: Required for RN built in Event Emitter Calls.
+}
+
+@ReactMethod
+public void removeListeners(Integer count) {
+ // Keep: Required for RN built in Event Emitter Calls.
+}
 ...
 WritableMap params = Arguments.createMap();
 params.putString("eventProperty", "someValue");
--- a/docs/native-modules-android.md
+++ b/docs/native-modules-android.md
@@ -518,12 +518,12 @@ private void sendEvent(ReactContext reactContext,
 }
 @ReactMethod
 public void addListener(String eventName) {
- // Keep: Required for RN built in Event Emitter Calls.
+  // Set up any upstream listeners or background tasks as necessary
 }
 
 @ReactMethod
 public void removeListeners(Integer count) {
- // Keep: Required for RN built in Event Emitter Calls.
+  // Remove upstream listeners, stop unnecessary background tasks
 }
 ...
 WritableMap params = Arguments.createMap();
--- a/website/versioned_docs/version-0.65/native-modules-android.md
+++ b/website/versioned_docs/version-0.65/native-modules-android.md
@@ -516,6 +516,15 @@ private void sendEvent(ReactContext reactContext,
      .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
      .emit(eventName, params);
 }
+@ReactMethod
+public void addListener(String eventName) {
+ // Keep: Required for RN built in Event Emitter Calls.
+}
+
+@ReactMethod
+public void removeListeners(Integer count) {
+ // Keep: Required for RN built in Event Emitter Calls.
+}
 ...
 WritableMap params = Arguments.createMap();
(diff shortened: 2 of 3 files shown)


simonbrunou:
Signed CLA.

mocktarissa (approved):

satya164 (commented):
Can you post what exception you got and share a link to the native module which implemented this?

simonbrunou:
When omitting `addListener`, this is the exception:
```bash
`new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
```

When omitting `removeListeners`, this is the exception:
```bash
`new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method.
```

I cannot share a link to the native module which implemented this when I was confronted with the problem as it is proprietary source code. However I can put together a file that implements it quickly, if needed.

docs/native-modules-android.md
```diff
@@ -516,6 +516,15 @@ private void sendEvent(ReactContext reactContext,
      .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
      .emit(eventName, params);
 }
+@ReactMethod
+public void addListener(String eventName) {
+ // Keep: Required for RN built in Event Emitter Calls.
```
rubennorte, line 521: I think it's a good idea to add these required methods in the examples, but I'd modify the comment to be more useful for people reading it.

In iOS we have the same (https://reactnative.dev/docs/native-modules-ios#sending-events-to-javascript):
```
@implementation CalendarManager
{
  bool hasListeners;
}

// Will be called when this module's first listener is added.
-(void)startObserving {
    hasListeners = YES;
    // Set up any upstream listeners or background tasks as necessary
}

// Will be called when this module's last listener is removed, or on dealloc.
-(void)stopObserving {
    hasListeners = NO;
    // Remove upstream listeners, stop unnecessary background tasks
}

- (void)calendarEventReminderReceived:(NSNotification *)notification
{
  NSString *eventName = notification.userInfo[@"name"];
  if (hasListeners) { // Only send events if anyone is listening
    [self sendEventWithName:@"EventReminder" body:@{@"name": eventName}];
  }
}
```

rubennorte:
Thanks @simonbrunou for improving our docs! This was definitely incomplete without the examples that you added. Can you please address my comment so we can merge it? Thanks!

simonbrunou:
> Thanks @simonbrunou for improving our docs! This was definitely incomplete without the examples that you added. Can you please address my comment so we can merge it? Thanks!

I think it's done, however I'm not familiar with GitHub's interface so correct me if I did something wrong.

Simek (changes requested):
Hi @simonbrunou, thank you for bringing this topic up! 👍 

Since the change landed in `0.65` release (https://github.com/facebook/react-native/commit/114be1d2170bae2d29da749c07b45acf931e51e2), please update also versioned docs for `0.65` and `0.66` besides the `next` file:

simonbengtsson:
I assume the tag should have been on @simonbrunou @Simek

Simek:
Ugh, yup! Sorry about that! I should not trust the GitHub mention autocomplete. 😅

simonbengtsson:
Haha 😅

Simek:
@simonbengtsson I will mark this conversation as off-topic now, to reduce the noise in the issue. Thank you so much for the comment and your forbearance! 🙂

simonbrunou:
I rebased on main and added the changes to the documentation.

Simek (commented):
It looks like you did not force push after rebase, which leads to malformed PR, please fix that. 😉

simonbrunou:
It should be good now.

Simek (approved):
Thank you for the updates! 👍 

From the technical standpoint it LGTM, but let's wait with the merge for the @rubennorte approval on the wording/description changes. 🙂

rubennorte:
I think we can still improve the comment. I don't think the current one is useful to tell people what this is for, only that's required. We could say something like what we say on iOS.

So instead of:
```
 // Keep: Required for RN built in Event Emitter Calls.
```

We could use something like:
```
@ReactMethod
public void addListener(String eventName) {
  // Set up any upstream listeners or background tasks as necessary
}

@ReactMethod
public void removeListeners(Integer count) {
  // Remove upstream listeners, stop unnecessary background tasks
}
```

@simonbrunou could you please make these changes? Sorry for the back and forth.

rubennorte:
Thanks for improving our website, @simonbrunou!

PR #2804, opened 2021-10-05
[Docs] Add an asterisk note to the hermes bundle is in use

limaAniceto:
Hi there,
Just updating some docs, which I hope you agree with

At Mixcloud we have a bit of a non-standard way of loading the JS bundle 👀 

By reading the docs, you get the idea that as soon as the `HermesInternal` variable is available - that you are using Hermes (and you are).

However, you're still not necessarily making use of the HermesEngine generated bytecode - the `.hbc` file

This is certainly a small change but it makes it clearer and prevents other people from being induced in error like we did :)

Thank you for all your work 👍


docs/hermes.md
```diff
@@ -89,12 +89,14 @@ If you've recently created a new app from scratch, you should see if Hermes is e
...
 ![Where to find JS engine status in AwesomeProject](/docs/assets/HermesApp.jpg)
 
-A `HermesInternal` global variable will be available in JavaScript that can be used to verify that Hermes is in use:
+A `HermesInternal` global variable will be available in JavaScript that can be used to verify that Hermes is in use*:
 
 ```jsx
 const isHermes = () => !!global.HermesInternal;
 ```
 
+*If you are using a non-standard way of loading the JS bundle, it is possible that the `HermesInternal` variable is available but you aren't using the hermes generated bytecode. Confirm that you are using the `.hbc` file and also benchmark the before/after as detailed below.
```
satya164, line 98: I think it's a bit confusing to use **but**, since `HermesInternal` only means you are using Hermes engine, bytecode usage is another thing on top of that and it doesn't contradict whether Hermes is used or not.

imo this not probably makes more sense below the line: `This will compile JavaScript to bytecode during build time which will improve your app's startup speed on device.` as it's related.

docs/hermes.md
```diff
@@ -89,12 +89,14 @@ If you've recently created a new app from scratch, you should see if Hermes is e
 
 ![Where to find JS engine status in AwesomeProject](/docs/assets/HermesApp.jpg)
 
-A `HermesInternal` global variable will be available in JavaScript that can be used to verify that Hermes is in use:
+A `HermesInternal` global variable will be available in JavaScript that can be used to verify that Hermes is in use*:
```
Simek, line 92:
```suggestion
A `HermesInternal` global variable will be available in JavaScript that can be used to verify that Hermes is in use:
```

docs/hermes.md
Simek: Instead using asterisk (is not the common pattern in docs), pleas wrap this information with blockquote (via `>`). 

This will display the message on the website in the form of note. 

With the wording correction, and with the display tweak in, this PR should be ready to merge. 🙂

Simek (changes requested):

PR #2808, opened 2021-10-07
Update Algolia DocSearch appId / apiKey

slorber:
Algolia DocSearch is migrating to a new system. 

All DocSearch sites will need to upgrade in the next months.

See also https://github.com/facebook/docusaurus/pull/5632


slorber:
Replaced by https://github.com/facebook/react-native-website/pull/3022

PR #2819, opened 2021-10-19
Correct documentation for `onSelectionChange`

lyahdav:
https://github.com/facebook/react-native-website/pull/881 added a note to say that `onSelectionChange` only works when `multiline` is enabled but I just tested off RN main branch as of today without `multiline` set and `onSelectionChange` still fires. Maybe this was a bug previously or a change that was made to the API but the docs were never updated.


Simek:
After a quick look in main repo I have found this issue with reference to PR and commit fixing this:
* https://github.com/facebook/react-native/issues/21515

Which seems to fix the `onSelectionChange` on iOS for single line inputs. If this commit is the right one, then it looks like the change has landed in `0.61` release and all docs from this version upwards should be updated.

PR #2832, opened 2021-10-28
Update intro-react.md

mbuildingeveryday:
On this line, "When `this.state.isHungry` is false, the `Button`’s `disabled` prop is set to `false` and its `title` also changes:",
When 'this.state.isHungry' is set to false, disabled prop, which is {!this.state.isHungry}, is set to true by negating of this.state.isHungry.
So, it should be like this. "When `this.state.isHungry` is false, the `Button`’s `disabled` prop is set to `true` and its `title` also changes:"


Simek:
Hi @mbuildingeveryday, thank you for the update, however you need to accept the CLA, otherwise we cannot merge or review your PR.

Simek (approved):
LGTM, thank you catching this! 👍

PR #2833, opened 2021-10-29
docs: fix tag in linking docs

luism3861:
this PR fix close tag in linking docs!.


luism3861:
> Hi @luism3861, can you also update at least the `next` docs? Ideally, it would be nice to update few older versions too, since the change is quite simple to port. 🙂

hi @Simek  thanks for the reply, after search `linking.md ` section  `</queries`> doesn't exist in older versions, next docs it's weird `</queries>`  tag is closed!.

Simek (approved):
Hmm, that weird, but I can confirm that `next` version is already updated, so this change LGTM. 

Thank you for the fix and investigating for the backport request! 👍

PR #2836, opened 2021-10-31

sflahave:
Changed the instruction to hit  `⌘R` to reload to `⌘D` since the former is now the iOS Simulator command to record your screen rather than refresh.


Simek:
Refs:
* https://github.com/facebook/react-native/issues/31467

However I'm not sure if changing docs to mention `⌘D` and select "Reload" option is a valid workaround for the issue in every case, maybe we need to wait for the RN to adjust the key command and then update the docs accordingly.

PR #2850, opened 2021-11-17
Update docs for image.md

erheron:
**Proposed documentation update**

This is missing, but this is an important information. Imagine a view sized 100x150 and an image sized 200x300. The old description could be interpreted in a way, that the image doesn't scale at all!
I agree, once you know it, this might be obvious, but I still feel like it's needed piece of info


lunaleaps (approved):
Looks good to me!

PR #2857, opened 2021-11-21
Remove doc reference to duration in scrollToEnd

le-michael:
Looks like duration prop isn't available anymore for ScrollView.scrollToEnd.

https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Components/ScrollView/ScrollViewCommands.js#L28-L31

According to this comment it was deleted at some point.

https://github.com/facebook/react-native/issues/25589#issuecomment-510488031

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -752,9 +752,9 @@ Scrolls to a given x, y offset, either immediately, with a smooth animation.
 ### `scrollToEnd()`
 
 ```jsx
-scrollToEnd(([options]: { animated: boolean, duration: number }));
+scrollToEnd(([options]: { animated: boolean }));
 ```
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @le-michael, this is a nice catch, thank you for all the references! 👍 

It looks like that the `duration` property was removed since `0.60`, so we need to backport this change to all of the [versioned docs](https://github.com/facebook/react-native-website#versioned-docs). 

Would you also like to take care of those changes?

Fortunately, other components which also uses `scrollToEnd` method like `VirtualizedList` or `FlatList` have the correct definition of the function arguments and params. 😄

le-michael:
Hey @Simek,

I probably won't have time to make those changes! You should probably make a bug for someone else to pick it up 🙂

Simek (approved):
No problem, I will follow up with the PR with remaining updates!

PR #2859, opened 2021-11-23
index should not be used for react list keys

Shubh1902:
Using index as keys can create potential bugs where the item is changed but react would not rerender it since the index is same in that list.

### DIFF
--- a/docs/using-a-listview.md
+++ b/docs/using-a-listview.md
@@ -89,7 +89,7 @@ const SectionListBasics = () => {
           ]}
           renderItem={({item}) => <Text style={styles.item}>{item}</Text>}
           renderSectionHeader={({section}) => <Text style={styles.sectionHeader}>{section.title}</Text>}
-          keyExtractor={(item, index) => index}
+          keyExtractor={(item, index) => `${item}-${index}`}
         />
       </View>
     );


docs/using-a-listview.md
```diff
@@ -89,7 +89,7 @@ const SectionListBasics = () => {
           ]}
           renderItem={({item}) => <Text style={styles.item}>{item}</Text>}
           renderSectionHeader={({section}) => <Text style={styles.sectionHeader}>{section.title}</Text>}
-          keyExtractor={(item, index) => index}
+          keyExtractor={(item, index) => `${item}-${index}`}
```
Simek, line 92: Hi @Shubh1902, thank you for sotting this out! 👍 

Unfortunately this change might also cause an issue or be at least confusing, when the data list includes objects not strings. Storing objects seems to be the most common approach to providing the list data, see:
* https://reactnative.dev/docs/flatlist#keyextractor

What do you think about replacing reference to`${item}` with a hardcoded string for this example?

yungsters:
> Using index as keys can create potential bugs where the item is changed but react would not rerender it since the index is same in that list.

@Shubh1902, this is no different than doing the following, right?

```
function Foo(props) {
  return (
    <View>
      {props.items.map((item, index) => (
        <Text key={index}>{item}</Text>
      )}
    </View>
  );
}

// <Foo items={['a', 'b']} />
// <Foo items={['b', 'a']} />
```

This example will still work correctly, but problems begin to arise when each item is stateful (which isn't the case for `Text` in this example).

So in a sense, the current example is not wrong, but someone who does not understand how `key` works in React may encounter unexpected problems. If we wanted to highlight that, then we may want to change the input data to be include an `id`.

```
function Foo(props) {
  return (
    <View>
      {props.items.map((item, index) => (
        <Text key={item.id}>{item.text}</Text>
      )}
    </View>
  );
}

// <Foo items={[{id: 1, text: 'a'}, {id: 2, text: 'b'}]} />
// <Foo items={[{id: 2, text: 'b'}, {id: 1, text: 'a'}]} />
```

PR #2861, opened 2021-11-25
update and generalize all of the `scrollToEnd` prop descriptions

Simek:
# Why & how

Follow up to #2857 

This PR fixes the invalid `duration` key description for the `ScrollList` and attempts to generalize the description and documentation of all `scrollToEnd` props available in the `*List` components.

In the future, we should also take a deeper look into the the exact exported types per component/list, since usually we do not duplicate the props description if that prop comes from parent/extended component. In case of `scrollToEnd`, it might be only a `VitualizedList` prop, but for now it was more important to remove invalid information to avoid users confusion. 🙂 

I have also removed few in-content ToC's on the deperecated `ListView` pages, since they are the relict of the past, when website do not use the build-in ToC (right sidebar). Somehow this list survived the migration efforts to the Docusuaurs V2. 🤷‍♂️

### DIFF
--- a/docs/virtualizedlist.md
+++ b/docs/virtualizedlist.md
@@ -571,12 +571,20 @@ hasMore () => boolean;
 ### `scrollToEnd()`
 
 ```jsx
-scrollToEnd((params: object));
-
-Valid `params` consist of:
-
-- 'animated' (boolean). Optional default is true.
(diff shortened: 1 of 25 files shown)


yungsters:
Thanks, @Simek!

PR #2862, opened 2021-11-25
docs: update announceForAccessibility with queue

peterc1731:
This PR adds documentation for `announceForAccessibilityWithOptions`, added in https://github.com/facebook/react-native/pull/32637


yungsters:
@Simek — Logistical question… when facebook/react-native#32637 is merged, do I need to wait before merging this?

Simek:
@yungsters Depends on the timeframe, every case need an individual approach. Sometimes you need to wait even for the publish of next release, if upcoming version branch has been already cut, and PR has been merged only on `main`.

This is because `next` doc should always reflect the state of next release (`0.67` ATM), so additional or future changes need to wait until `next` will become the docs for their target release.

yungsters:
So the PR has been merged to main, but 0.67 has already been cut. That means I have to wait until 0.67 is released before I can merge this, right?

cc @lunaleaps, @rachelnabors - This is an interesting point of coordination friction that can also lead to outdated docs, relevant to our discussion about keeping reference docs more up-to-date.

peterc1731:
I've updated this to reflect the changes made when https://github.com/facebook/react-native/pull/32637 was merged

Simek (approved):
I have updated the docs for `0.68`, so this should b ready to ship.

Thank you @peterc1731, for the update in the first place! 👍

PR #2863, opened 2021-11-26
Update props of SectionList to declare inheritance from VirtualizedList explicitly

david50407:
Update props inheritance

Props of SectionList are inherits from VirtualizedSectionList -> VirtualizedList -> ScrollView.

Since VirtualizedSectionList isn't documented on website, IMO, SectionList should declare its props is inherited from VirtualizedList at least.


yungsters:
Makes sense to me. Thanks!

PR #2868, opened 2021-11-29
Update handling-text-input.md to emphasize that setText(t) does not t…

mplawley:
…ake component state for input

Very minor: Since `PizzaTranslator` has `text` as a variable in its state, I propose that the `onChangeText` callback use a different variable name to emphasize that this value is *not* derived from component state but instead is used to set it. I use `t` to avoid verbosity, but `newText` might also be appropriate.

Note: I am a React Native neophyte; if the above goes against stylistic conventions with this framework, then perhaps an explanatory note below this page's code example would suffice.

### DIFF
--- a/docs/handling-text-input.md
+++ b/docs/handling-text-input.md
@@ -18,7 +18,7 @@ const PizzaTranslator = () => {
       <TextInput
         style={{height: 40}}
         placeholder="Type here to translate!"
-        onChangeText={text => setText(text)}
+        onChangeText={newText => setText(newText)}
         defaultValue={text}
       />
       <Text style={{padding: 10, fontSize: 42}}>


yungsters:
Thanks for the contribution! And good catch — shadowing of variables in general can make code harder to read, let alone in example code.

Can you rename `t` to `newText` to make it clearer, though?

In general (across the React projects and internally at Meta), we tend to skew toward more verbose but descriptive variable naming.

mplawley:
> Thanks for the contribution! And good catch — shadowing of variables in general can make code harder to read, let alone in example code.
> 
> Can you rename `t` to `newText` to make it clearer, though?
> 
> In general (across the React projects and internally at Meta), we tend to skew toward more verbose but descriptive variable naming.

Done! Since the variable name change is so small, I took the liberty of squashing the two commits so as no to clutter your repo. Thank you for the reply :raised_hands:

yungsters:
Awesome, thank you again!

PR #2869, opened 2021-11-30
win/android docs: update node install to use nodejs-lts

PaulB-H:
What:
Updated React Native CLI Quickstart/Windows-Android docs to use the LTS version of node during choco install, and recommended use of LTS in general, similarly to how React docs do the same.

Why:
_getting-started-windows-android.md currently lists the choco install command as:
```choco install -y nodejs.install openjdk8```
...which will install node.js version 17 (current)

While you will still be able to init an app with the react-native CLI and node.js 17, the command ```npx react-native run-android``` **_will fail_** due to the same bug affecting [facebook/create-react-app#11562](https://github.com/facebook/create-react-app/issues/11562)


yungsters:
Thanks for the contribution!

PR #2874, opened 2021-12-05
[DOCS] Update flatlist.md

SyntaxError843:
changed the type of `ItemSeparatorComponent` from component to function, since in the given example a function was passed rather than a component, also, in my personal experience when I tried passing a component directly it broke saying it was expecting a string/class/function.


SyntaxError843:
holy shit i change one fucking word and i have to go through so much bullshit to get it noticed wtf is this

PR #2884, opened 2021-12-16
Add "return" details for addEventListener()

gregfenton:
There was no description of the return value from `addEventListener()`.

There currently does not seem to be documentation for [EventSubscription](https://github.com/facebook/react-native/blob/00bb2ba62d86477884484aad0cd72d3627d5f221/Libraries/vendor/emitter/EventSubscription.js#L17), which I would link to from this page if such a doc page existed.

### DIFF
--- a/docs/appstate.md
+++ b/docs/appstate.md
@@ -160,7 +160,7 @@ Received when the user is not actively interacting with the app. Useful in situa
 addEventListener(type, handler);
 ```
 
-Add a handler to AppState changes by listening to the `change` event type and providing the handler
+Add a handler to AppState changes by listening to the `change` event type.  Returns the `EventSubscription`.
 
 ---
 
@@ -170,7 +170,7 @@ Add a handler to AppState changes by listening to the `change` event type and pr
 removeEventListener(type, handler);
 ```
 
-> **Deprecated.** Use the `remove()` method on the event subscription returned by [`addEventListener()`](#addeventlistener).
+> **Deprecated.** Use the `remove()` method on the `EventSubscription` returned by [`addEventListener()`](#addeventlistener).
 
 ## Properties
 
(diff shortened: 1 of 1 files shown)


docs/appstate.md
```diff
@@ -157,20 +157,21 @@ Received when the user is not actively interacting with the app. Useful in situa
 ### `addEventListener()`
 
 ```jsx
-addEventListener(type, handler);
+addEventListener(eventType, listener);
 ```
 
-Add a handler to AppState changes by listening to the `change` event type and providing the handler
+Sets up a function that will be called whenever the specified event type on AppState occurs. Valid values for `eventType` are
+[listed above](https://reactnative.dev/docs/appstate#events).  Returns the `EventSubscription`.
```
Simek, line 164:
```suggestion
[listed above](#events). Returns the `EventSubscription`.
```

This suggestion should fix the Lint error. It also suggest the switch from the absolute link to the relative one.

Simek:
> There currently does not seem to be documentation for [EventSubscription](https://github.com/facebook/react-native/blob/00bb2ba62d86477884484aad0cd72d3627d5f221/Libraries/vendor/emitter/EventSubscription.js#L17), which I would link to from this page if such a doc page existed.

That's a nice catch. Would you like to try adding the type based on the "Object Types" section in "Components" docs and the source file? (https://github.com/facebook/react-native/blob/main/Libraries/vendor/emitter/EventSubscription.js). 

However, as you can see in the file above, there is a note which states that `EventSubscription` is a temporary type, and in the future it will be replaced by `EmitterSubscription` from `EventEmitter`, so I'm no sure if adding the type at this time is a good idea. It might confuse some of the users.

gregfenton:
I guess I'm going to suggest we leave the topic of `EventSubscription` as is.

Simek (approved):
Thank for the correction and review update, LGTM! 👍

PR #2905, opened 2021-12-24
architecture section small tweaks

Simek:
The following PR introduces the following changes to the Architecture:
* rename of Architecture Glossary page to `architecture-glossary` instead of just `glossary` (this matches the convection introduced by `architecutre-overview` and there might be a second "glossary"/"appendix" when the new architecture Playbook will land). Ideally we should start to group the docs files into subfolders (v1 doesn't support this afaik, but it should not be a problem in v2.
* update the Architecture sidebar, so it structure matches the other top-level pages on the website (flipped `collapsible`)
* reorder platforms in one of the sentence to alphabetical order.

### Sidebar preview
Screenshot 2021-12-24 at 12 04 43

### DIFF
--- a/docs/architecture-glossary.md
+++ b/docs/architecture-glossary.md
@@ -1,5 +1,5 @@
 ---
-id: glossary
+id: architecture-glossary
 title: Glossary
 ---
 
@@ -9,7 +9,7 @@ React Native executes the same React framework code as React for the web. Howeve
 
 ## Host platform
 
-The platform embedding React Native (e.g., Android, iOS, Windows, macOS).
+The platform embedding React Native (e.g., Android, iOS, macOS, Windows).
 
 ## Host View Tree (and Host View)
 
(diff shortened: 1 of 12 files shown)


lunaleaps:
Thanks for these changes @Simek! I think we have to update the 0.66 version of the docs too.

I've been doing it manually so far, but do you happen to know if there is a better way?

Simek:
> I've been doing it manually so far, but do you happen to know if there is a better way?

Unfortunately there is no "mechanism" or another way to do this AFAIK. I also backport the changes manually, when it's needed.

Simek:
@lunaleaps The changes has been backported to 0.66 release (including sidebar tweaks).

I would appreciate the review on this, since due to file name change this PR produces a lot of conflicts in time. 😅

lunaleaps (approved):
LGTM!

PR #2906, opened 2021-12-26
Added a known issue

Humbarrt:
This bug is really annoying. It makes masking a textinput virtually impossible. All other established Frameworks dont have this problem.


Simek:
Hello @Humbarrt, it looks like the issue you would like to add is closed and it was not marked with a [Known Issue](https://github.com/facebook/react-native/issues?q=is%3Aopen+is%3Aissue+label%3A%22%F0%9F%93%AEKnown+Issues%22) label, so we cannot accept it here, but anyway thank you for the contribution!

PR #2908, opened 2021-12-28
Updating the word "currently" to "might be"

devamiya:


Simek:
Hi @devamiya, thank you for the update, however it looks like the change is a duplicate of earlier PR mentioned above.

In cases like that, we usually merge the earlier PR if possible. Thanks for understanding!

PR #2912, opened 2021-12-30
docs: unwanted text in code block

lorenzovecchi:
Fixed folder path inside code block


Simek (approved):
Hi @lorenzovecchi, nice catch, thank you for the correction! 👍

PR #2913, opened 2021-12-30
Remove unnecessary keyword

hyunahOh:
useEffect is already imported as itself.


Simek (approved):
Hi @hyunahOh, thank you for spotting this out and pushing the fix! 👍

PR #2915, opened 2022-01-03
feat: add cancelButtonTintColor props in ActionSheetIOS for change cancel tint color

nomi9995:
Reference https://github.com/facebook/react-native/pull/31972
This prop was missing in docs


Simek (approved):
Hello @nomi9995, thank you for the update! 👍 

It looks like the change landed in 0.67 release branch, so there is no need to backport this. Let's ship it before the next docs cut! 🙂

PR #2916, opened 2022-01-03
Button style updated

DevelopedByToby:
Made some style changes to the ActionButton. Please feel free to update or change anything.


Simek:
Hi @ItsTobez, ATM we aim to keep the React and React Native website style quite similar, and button shapes has been based on that principle, as you can see in here:
* https://reactjs.org/

After the new beta site for React will land, we probably will also go through, and revise or update the React Native website design/appearance too, but for now I would like to leave the design as is. Hope you can understand that. 🙂

ItsTobez:
No worries! :)

PR #2921, opened 2022-01-09
Adding a command

DiegoDialga:
While starting the server , I found 'docusaurus' is not recognized as an internal or external command, operable program or batch file .  It took me some time to find the solution and found this command on stackoverflow. If included here, the command might help the user in dealing with such problems and it will save the user's time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ If you are looking for the source code of the [React Native Archive website](htt
 
 1.  `cd website` to go into the website portion of the project.
 1.  `yarn start` to start the development server _(powered by [Docusaurus](https://v2.docusaurus.io))_.
+    **_Note: In case you find any error regarding [Docusaurus], while starting the development server, run the command `yarn upgrade docusaurus --latest`._**
 1.  `open http://localhost:3000/` to open the site in your favorite browser.
 
 ## 📖 Overview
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hello @DiegoDialga,

Can you provide the link to the Stack Overflow post you have mentioned? Also the information what exactly error you were receiving without the upgrade would be helpful.

Generally, running the added command will upgrade the Docusaurus version used on the website to the latest release, sice we are using `^` in a semver, and this might lead to other, unexpected compatibility issues and problems.

Also, did you run the `yarn` command in the `website` directory, as stated in the guide, before running `yarn start` in `website` folder?

DiegoDialga:
Hi @Simek ,
This is the link for the stackoverflow post "https://stackoverflow.com/questions/68632133/how-to-solve-docusaurus-yarn-start-error?noredirect=1&lq=1". I think you are right, I might not have run the yarn command before starting it.
Then I guess adding 'run' before yarn might help otherwise it is not clearly specifying that the user has to run the yarn command before yarn start

Simek:
Thank you for clarifying! 👍 I'm happy that I was able to help!

> Then I guess adding 'run' before yarn might help otherwise it is not clearly specifying that the user has to run the yarn command before yarn start

Running `yarn` step is listed just few lines above, in the Readme file, in the "Installation" section:
* https://github.com/facebook/react-native-website/blob/main/README.md#installation

If you think that the steps can be improved, please make an update to this PR.

Simek (approved):
LGTM, thank you for the updates! 👍

PR #2923, opened 2022-01-10
docs: adds `View props` to props section of Modal

carlos3g:
[here](https://github.com/facebook/react-native/blob/main/Libraries/Modal/Modal.js#L61) we can see that Modal inherits view props, so I'm adding this information to docs


Simek (approved):
Hi @carlos3g, nice catch, thank you for the update! 👍

PR #2924, opened 2022-01-10
Update textinput.md

dragosbocancea:

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -130,11 +130,11 @@ Tells `TextInput` to automatically capitalize certain characters. This property 
 
 ---
 
-### `autoComplete` <div class="label android">Android</div>
+### `autoCompleteType` <div class="label android">Android</div>
 
-Specifies autocomplete hints for the system, so it can provide autofill. On Android, the system will always attempt to offer autofill by using heuristics to identify the type of content. To disable autocomplete, set `autoComplete` to `off`.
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hello @dragosbocancea, thank you for aupdating the docs! 👍 

Can you provide some context for this change?

dragosbocancea:
> Hello @dragosbocancea, thank you for updating the docs! 👍
> 
> Can you provide some context for this change?

While trying to trigger a TextInput field to be autocompleted using a password manager, I have noticed that the former `autoComplete` did not do anything on android, but by replacing it to `autoCompleteType`, as suggested in one stackoverflow 
 post (if I remember correctly), it worked as desired. Therefore I supposed that the correct prop would be `autoCompleteType`, not `autoComplete`.

dragosbocancea:
Up?

Simek:
Unfortunately, I cannot confirm your changes with the React Native code. 

There is no entry name `autoCompleteType` when searching thought the whole repository:
* https://cs.github.com/facebook/react-native?q=autoCompleteType

The prop named `autoComplete` can be found in the `TextInput` JS file as well as in Android native code:
* https://github.com/facebook/react-native/blob/main/Libraries/Components/TextInput/AndroidTextInputNativeComponent.js#L125
* https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/views/textinput/ReactTextInputManager.java#L758

PR #2925, opened 2022-01-11
update name in tutorial.md

EKivutha:
Current name `React-Native-specific` seems like a React Native library as that is the library naming convection, recommended to remove hyphen from the name to React-Native specific


Simek (approved):
Refs:
* #2920

LGTM, thank you for the update! 👍

PR #2928, opened 2022-01-12
switch to esbuild-loader, add Flow highlight support

Simek:
This PR aims to improve the build and deploy speed by switching the Webpack loader to the `esbuild-loader`.

This do not yield same effects as fully switching to full ESBuild pipeline, but should shorten the build times a bit.

I have also added Flow to the additional supported languages by Prism, since the new architecture playbook included few  examples written in Flow and we were using TS rules so far to highlight those code blocks.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -37,6 +37,16 @@ const lastVersion = versions[0];
     locales: ['en'],
   },
   onBrokenLinks: 'throw',
+  webpack: {
+    jsLoader: isServer => ({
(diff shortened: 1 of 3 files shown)


cortinico (approved):

Simek:
FYI. It is know issue that the Docusuaurs do not work with latest version of `esbuild` and `esbuild-loader`. 

This problem probably will be addressed in one of the future release. `jsLoader` configuration is a quite new property and neew few tweak and fixes, however Docusuaurs website already uses it in the production already. 🙂

PR #2935, opened 2022-01-18
Use descriptive component names and fix use of ref

sammy-SC:
Small changes to the playbooks to make JS examples clearer.


cortinico (approved):
LGTM 👍

docs/new-architecture-library-intro.md
```diff
@@ -134,27 +134,29 @@ class MyComponent extends React.Component<Props> {
   }
 }
 
-function MyJSComponent(props) {
-  return (
-    <View>
-      <SubmitButton onSubmit={props.onSubmit} />
-    </View>
-  );
+class ChildComponent extends React.Component<Props> {
```
Simek, line 137: Does the component need to be a class? Generally we prefer functional components in examples if possible.

Ideally it would be nice to provide both examples in tabs like in most of the APIs docs:
* https://reactnative.dev/docs/next/dimensions#example

sammy-SC: Example with function component is below on line 171.
I will think about it and create a separate PR cc @cortinico

PR #2936, opened 2022-01-21
Update ScrollView document to add pointerEvents prop

ryancat:

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -527,6 +527,42 @@ When true, ScrollView allows use of pinch gestures to zoom in and out.
 
 ---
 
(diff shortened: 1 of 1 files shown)


Simek:
Hi @ryancat, thank you for the update suggestion! 👍 Unfortunately `pointerEvents` prop is inherited from the `View` component:
* https://reactnative.dev/docs/view#pointerevents

In cases like this we avoid the prop duplication on the child component, because otherwise list of the props will be quite overwhelming and maintaining the multiple instances of the same prop in various pages would be difficult and it would be easy to miss one. Hope you understand that! 🙂

PR #2945, opened 2022-01-23
Fix flexShrink default value in Layout Props docs

jzxchiang1:
This PR was not complete: https://github.com/facebook/react-native-website/pull/2451/files


yungsters:
Thanks for the fix!

PR #2953, opened 2022-02-03
Update scrollview.md

didactics:

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -429,16 +429,18 @@ Called when the momentum scroll ends (scroll which occurs as the ScrollView glid
 
 ### `onScroll`
 
-Fires at most once per frame during scrolling. The frequency of the events can be controlled using the `scrollEventThrottle` prop. The event has the following shape (all values are numbers):
+Fires at most once per frame during scrolling. The frequency of the events can be controlled using the `scrollEventThrottle` prop. The event has the following shape:
 
 ```js
 {
   nativeEvent: {
+    responderIgnoreScroll,
+    velocity: {x, y},
     contentInset: {bottom, left, right, top},
     contentOffset: {x, y},
     contentSize: {height, width},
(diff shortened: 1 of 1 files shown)


SurajPratap10:
Sign the  Contributor License Agreement first before applying for a Pull Request

Simek (changes requested):
Hello @didactics, thank you for the update! 👍 

Can you align the object shape according to the source code:
* https://github.com/facebook/react-native/blob/main/Libraries/Types/CoreEventTypes.js#L101-L132

But that I mean:
* correctly mark optional fields (`?`)
* revert `zoomScale` add missing type, 
* remove `target` (this is a part of `SyntheticEvent` which `ScrollEvent` implements, we tend to skip inherited fields in doc)
* add `targetContentOffset` and its value.

It would be also nice if you could retain the fields order from the object in the source, but this is not obligatory.

Also if you want to remove the  "all values are numbers" part, I think you need to add those types in object shape. I assume is due to `responderIgnoreScroll` type, but now we miss that information.

docs/scrollview.md
```diff
@@ -429,16 +429,18 @@ Called when the momentum scroll ends (scroll which occurs as the ScrollView glid
 
 ### `onScroll`
 
-Fires at most once per frame during scrolling. The frequency of the events can be controlled using the `scrollEventThrottle` prop. The event has the following shape (all values are numbers):
+Fires at most once per frame during scrolling. The frequency of the events can be controlled using the `scrollEventThrottle` prop. The event has the following shape:
```
Simek, line 432:
```suggestion
Fires at most once per frame during scrolling. The frequency of the events can be controlled using the `scrollEventThrottle` prop. The event has the following shape (all values with unspecified type are numbers):
```
Let's retain this sentence, but alter it a bit.

Simek:
Thanks for pointing this out, closing in favour of mentioned above PR.

PR #2959, opened 2022-02-08
fix: spring damping error in example

swapanil:
In the given example, an error is thrown as spring damping


HyunseungLee-Travis:
#swapanil, seems like a bug in React Native.
The code/example is actually for using `create()`, so there is no need for changing the code to using objects.
I created a [new issue](https://github.com/facebook/react-native/issues/33740) in React Native Repo

Simek (approved):
Let's merge the animation changes for now, since there is a bug related to the `create`.

Thank you for the update @swapanil! 👍

PR #2964, opened 2022-02-11
(Image) Fix documentation for onLoad event.

ErickMaeda:
### Problem
The [onLoad](https://reactnative.dev/docs/next/image#onload) event documentation is not correct, here is the sample:

#### Current with bug
```js
const onLoad = ({width, height, uri}) = {}
```

#### Fixed version
```js
const onLoad = ({nativeEvent: {source: {width, height, uri}}}) = {}
```

It was missing `nativeEvent` and `source` property. If you look at [onLayout](https://reactnative.dev/docs/next/image#onlayout) it has the `nativeEvent` property as well.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -315,9 +315,11 @@ Invoked on mount and on layout changes.
 
 Invoked when load completes successfully.
 
+**Example:** `onLoad={({nativeEvent: {source: {width, height}}}) => setImageRealSize({width, height})}`
+
 | Type                                             |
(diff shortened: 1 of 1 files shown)


lunaleaps (approved):

lunaleaps:
@ErickMaeda Oh one thing I just realized, is that these docs will only appear for `version: next`. If you'd like you can also make a PR for updating existing 0.67 and below docs

PR #2967, opened 2022-02-14
Mark Text selectable prop as android only

randomor:
`selectable` is also android only: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1349b640d4d07f40aa7c1c6931f18e3fbf667f3a/types/react-native/index.d.ts#L830

Also there seems to be a long feature request to support this on iOS? https://react-native.canny.io/feature-requests/p/support-selecting-text-within-text-elements-for-copypaste-etc

### DIFF
--- a/docs/text.md
+++ b/docs/text.md
@@ -571,7 +571,7 @@ When the scroll view is disabled, this defines how far your touch may move off o
 
 ---
 
-### `selectable`
+### `selectable` <div class="label android">Android</div>
 
 Lets the user select text, to use the native copy and paste functionality.
 


Simek:
Hello @randomor, thank you for the update! 👍 

However, in the RN source code, there is no annotation or information that this Android only prop, it's a part of general `TextProps`:
* https://github.com/facebook/react-native/blob/main/Libraries/Text/TextProps.js#L147

So this is something that need a bit more verification, before we can ship this. I will look into this soon.

randomor:
@Simek good catch! I just tested again iOS does offer selectable function, although limited to selecting the whole text, no control for the exact range of text. Please feel free to close this for now.

PR #2971, opened 2022-02-16
Adding Explain Like I'm 5 video to the home page

dmitryvinn:
# Summary

Here at Meta Open Source, we've been creating short-form videos to explain Meta OSS projects in the simplest terms possible. We started embedding these videos into project pages as you can see in [Docusaurus home page](https://docusaurus.io/), [Jest](https://jestjs.io/), [Litho](https://fblitho.com/), [Hydra](https://hydra.cc/), etc..

# Test Plan
Here is a screenshot in how it looks:
image


Simek (commented):
Hello @dmitryvinn, thank you for pushing this update. 👍 

Generally I think it's a good idea to add the 5 minutes video to the homepage. IIRC there were discussions earlier, if we can update or replace Talks video/section (it's a bit outdate), so it might be a good opportunity to do that. 

However, if we decide to leave two sections, I would have a few design related suggestions - it would be nice to stick to alternating background color of sections and separate 5min video with talks this way, also the copy on the newly added part seems a bit to long now, we should aim to fit the description in way, which do not extend the video section container.

Let's ask others about their opinion.

dmitryvinn:
@Simek  absolutely - I will go with whatever approach you and the team decides to take

dmitryvinn:
@Simek I don't see any other comments on this PR - I think replacing the video might be the easiest thing to do, so if you agree, I can update the PR with that design

website/src/pages/index.js
```diff
@@ -79,6 +79,13 @@ Members of the React Native team frequently speak at various conferences.
 <br/><br/>
 You can follow the latest news from the React Native team on Twitter
   `,
+  introVideo: `
+The <a href="https://opensource.facebook.com/">Meta Open Source team</a> has put together a short overview of React Native, 
+where they explained the project in beginner's terms. You can also find 
+other content about Meta Open Source projects on their <a href="https://www.youtube.com/channel/UCCQY962PmHabTjaHv2wJzfQ">YouTube Channel</a>.
```
lunaleaps, line 85: I think this is odd to add -- I don't think coming to the React Native homepage folks are generally looking for other Meta Open Source projects. Can we remove the text "You can also find other content... You can follow the latest Meta OSS news.. " and just keep the twitter follow button?

lunaleaps: As well as re-ordering so the React Native talk video is on top as mentioned earlier?

lunaleaps (changes requested):

dmitryvinn:
What do you think about this version?
image

lunaleaps (approved):
Thanks @dmitryvinn! Looks good to me!

lunaleaps:
cc @Simek if you had more design comments? Maybe those can be addressed in a separate change? As well as updating the 2019 talk with a more current video

Simek (approved):
I would like to experiment with the arrangement a bit in the future, but this can be addressed in separate PR as well as talk video replacement as you suggested. 

After latest updates, the current iteration LGTM, I think that we can ship it. :)

PR #2986, opened 2022-03-02
Adding Social Banner in Support of Ukraine 

dmitryvinn:
## Summary
Our mission at [Meta Open Source](https://opensource.facebook.com/) is to empower communities through open source, and we believe that it means building a welcoming and safe environment for all. As a part of this work, we are adding this banner in support for Ukraine during this crisis.

## Testing
image

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -548,7 +548,7 @@ function SocialBanner() {
     <div className="SocialBanner">
       <div>
         Support Ukraine 🇺🇦{' '}
-        <a href="https://razomforukraine.org/donate">
+        <a href="https://opensource.facebook.com/support-ukraine">
           Help Provide Humanitarian Aid to Ukraine
         </a>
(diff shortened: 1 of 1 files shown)


dmitryvinn-fb:
Note: following this PR, we will be closing https://github.com/facebook/react-native/pull/33192

Simek (commented):
Can we use announcement banner feature for this, instead of implementing custom component?
* https://docusaurus.io/docs/next/api/themes/configuration#announcement-bar

We have used this feature before in few occasions and it should looks something like that:

yungsters:
In the spirit of getting this published as soon as possible, I'm merging this. But let's fix forward if we believe this can be done better.

leggomuhgreggo:
Curious about the decision making process here. This one is -- let's just say [controversial](https://twitter.com/consent_factory/status/1498257865693405185).

PR #2991, opened 2022-03-03
refactor(getting-started): path to sdk binaries

henryhobhouse:
## Description

Changes in two parts. For both Linux and MacOS getting started guides:

* From version 4 of command line tools the path to binaries was updated to `/cmdline-tools/latest/bin` with the previous being deprecated. 
* The name of the path environment variable has been updated to `ANDROID_SDK_ROOT` (https://developer.android.com/studio/command-line/variables#:~:text=ANDROID_HOME%2C%20which%20also%20points%20to%20the%20SDK%20installation%20directory%2C%20is%20deprecated.)

This PR updates the paths for both docs and replaces instances `ANDROID_HOME` with the updated `ANDROID_SDK_ROOT`.

NOTE: there are instances of ANDROID_HOME within the windows getting started guide but as they are backed up with images also utilising this name that I am unable to replace then have opted to leave that guide as it is. 

## Motivation and Context

The reason for the change is that the deprecated version of the SDK manager does not work with the java v11+ so making the getting started guide broken for new users as it recommended to install that version of the JDK. Equally the errors that get spun out are far from clear as to the problem making it hard to realise it requires a simple fix. 

## Types of Changes
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/docs/_getting-started-linux-android.md
+++ b/docs/_getting-started-linux-android.md
@@ -47,23 +47,22 @@ Next, select the "SDK Tools" tab and check the box next to "Show Package Details
 
 Finally, click "Apply" to download and install the Android SDK and related build tools.
 
-<h4>3. Configure the ANDROID_HOME environment variable</h4>
+<h4>3. Configure the ANDROID_SDK_ROOT environment variable</h4>
 
 The React Native tools require some environment variables to be set up in order to build apps with native code.
 
 Add the following lines to your `$HOME/.bash_profile` or `$HOME/.bashrc` (if you are using `zsh` then `~/.zprofile` or `~/.zshrc`) config file:
 
 ```shell
-export ANDROID_HOME=$HOME/Android/Sdk
-export PATH=$PATH:$ANDROID_HOME/emulator
-export PATH=$PATH:$ANDROID_HOME/tools
-export PATH=$PATH:$ANDROID_HOME/tools/bin
-export PATH=$PATH:$ANDROID_HOME/platform-tools
+export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
+export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
+export PATH=$PATH:$ANDROID_SDK_ROOT/cmdline-tools/latest/bin
+export PATH=$PATH:$ANDROID_SDK_ROOT/platform-tools
 ```
 
 > `.bash_profile` is specific to `bash`. If you're using another shell, you will need to edit the appropriate shell-specific config file.
 
-Type `source $HOME/.bash_profile` for `bash` or `source $HOME/.zprofile` to load the config into your current shell. Verify that ANDROID_HOME has been set by running `echo $ANDROID_HOME` and the appropriate directories have been added to your path by running `echo $PATH`.
+Type `source $HOME/.bash_profile` for `bash` or `source $HOME/.zprofile` to load the config into your current shell. Verify that ANDROID_SDK_ROOT has been set by running `echo $ANDROID_SDK_ROOT` and the appropriate directories have been added to your path by running `echo $PATH`.
 
 > Please make sure you use the correct Android SDK path. You can find the actual location of the SDK in the Android Studio "Preferences" dialog, under **Appearance & Behavior** → **System Settings** → **Android SDK**.
 
--- a/docs/_getting-started-linux-android.md
+++ b/docs/_getting-started-linux-android.md
@@ -56,6 +56,7 @@ Add the following lines to your `$HOME/.bash_profile` or `$HOME/.bashrc` (if you
 ```shell
 export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
 export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
+# Commandline Path is determined by version "/cmdline-tools/[VERSION]/bin" latest is the default from Android studio but please check you setup within the SDK Manager
 export PATH=$PATH:$ANDROID_SDK_ROOT/cmdline-tools/latest/bin
 export PATH=$PATH:$ANDROID_SDK_ROOT/platform-tools
 ```
--- a/docs/_getting-started-linux-android.md
+++ b/docs/_getting-started-linux-android.md
@@ -56,8 +56,6 @@ Add the following lines to your `$HOME/.bash_profile` or `$HOME/.bashrc` (if you
 ```shell
 export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
 export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
-# Commandline Path is determined by version "/cmdline-tools/[VERSION]/bin" latest is the default from Android studio but please check you setup within the SDK Manager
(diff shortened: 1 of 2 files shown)


docs/_getting-started-linux-android.md
```diff
@@ -47,23 +47,22 @@ Next, select the "SDK Tools" tab and check the box next to "Show Package Details
...
 
 ```shell
-export ANDROID_HOME=$HOME/Android/Sdk
-export PATH=$PATH:$ANDROID_HOME/emulator
-export PATH=$PATH:$ANDROID_HOME/tools
-export PATH=$PATH:$ANDROID_HOME/tools/bin
-export PATH=$PATH:$ANDROID_HOME/platform-tools
+export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
+export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
+export PATH=$PATH:$ANDROID_SDK_ROOT/cmdline-tools/latest/bin
```
cortinico, line 59: Why this?

cortinico (commented):
Thanks for doing this. I realized that `ANDROID_HOME` has been deprecated in fav. of `ANDROID_SDK_ROOT` and I think it's fair to update our docs.

docs/_getting-started-linux-android.md
henryhobhouse: My understanding is that the `tools/bin` path has been deprecated since v4 of the command line tools. If I try to use the `sdkmanager` binary, from the old path, with JDK v11+ I get the following with literally any argument:

Screenshot 2022-03-03 at 18 09 16

However if I use sdkmanager from `/cmdline-tools/latest/bin`:

Screenshot 2022-03-03 at 18 04 25

In similar fashion if I use binaries from `tools`. They either fail, work unexpectedly, or even state they are deprecated:

Screenshot 2022-03-03 at 18 09 16

cortinico: My question was related to the `cmdline-tools/latest/bin` path which I don't seem to have in my SDK setup

henryhobhouse: This could be bad assumption on my part but assumed that the path for linux would be the same as macos. Assuming you are on a linux machine?

Happy to revert the path change for the linux docs and just leave the env var change if you agree that is the best approach?

henryhobhouse: As a suggestion maybe we should make clear that the path is determined by version. 

i.e. 

`cmdline-tools/[VERSION]/bin`

And then example `latest` as imagine that will be the case for most people who are installing android studio for the first time?

henryhobhouse: I've pushed a change to add a comment explaining the versioning. Let us know what you think and/or if you have better suggestion to present this.

cortinico: I don't have SDK Command-line Tools installed, and I also believe this should not be a hard requirement for the React Native setup. I believe we should be fine by removing it entirely. Which tool are we using from there?

henryhobhouse: I've removed reference to command line path for linux getting started docs.

docs/_getting-started-linux-android.md
```diff
@@ -47,23 +47,21 @@ Next, select the "SDK Tools" tab and check the box next to "Show Package Details
...
 
 ```shell
-export ANDROID_HOME=$HOME/Android/Sdk
-export PATH=$PATH:$ANDROID_HOME/emulator
-export PATH=$PATH:$ANDROID_HOME/tools
-export PATH=$PATH:$ANDROID_HOME/tools/bin
-export PATH=$PATH:$ANDROID_HOME/platform-tools
+export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
+export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
+export PATH=$PATH:$ANDROID_SDK_ROOT/platform-tools
```
cortinico, line 59: Which just those 2 you won't be able to call `sdkmanager` which is inside `tools/bin/sdkmanager`. Could you cross check?

henryhobhouse: `tools/bin/sdkmanager` is deprecated. It is still installed because of reasons explained here: https://issuetracker.google.com/issues/67495440?pli=1#comment43

If you want to have access to the latest SDKManager CLI then you need to install the command line tools. In the comment above you have stated you do not wish to have command line tools as a hard dependency. 

I am not sure how you want to proceed. Revert to using the deprecated binary and caveat you have to use JDK 10 or below or install command line tools?

This maybe easier to jump onto a video call as at present I feel I am running in circles slightly as clearly not explaining the issue well enough.

cortinico: Ahh I see. Thanks for the issue pointer. We're safe with the paths you updated here then. If the user needs to invoke `sdkmanager`, they'll have to manually install the extra package you pointed.

cortinico (approved):

docs/_getting-started-macos-android.md
```diff
@@ -65,30 +65,30 @@ Next, select the "SDK Tools" tab and check the box next to "Show Package Details
...
 ```shell
-export ANDROID_HOME=$HOME/Library/Android/sdk
-export PATH=$PATH:$ANDROID_HOME/emulator
-export PATH=$PATH:$ANDROID_HOME/tools
-export PATH=$PATH:$ANDROID_HOME/tools/bin
-export PATH=$PATH:$ANDROID_HOME/platform-tools
+export ANDROID_SDK_ROOT=$HOME/Library/Android/sdk
+export PATH=$PATH:$ANDROID_SDK_ROOT/emulator
+# Commandline Path is determined by version "/cmdline-tools/[VERSION]/bin" latest is the default from Android studio but please check you setup within the SDK Manager
+export PATH=$PATH:$ANDROID_SDK_ROOT/cmdline-tools/latest/bin
```
cortinico, line 85: Could you revert this as well?

henryhobhouse: I can do however the doc directly references the sdkmanager CLI higher in this file (in lines 71-72). If we remove this we should remove reference to using the SDkManager CLI as well. As above let us know how you want to play it.

cortinico: My bad sorry. I just realized what is the full scope of this change.

I would anyway:
1. Revert this change to align it to the linux one
2. Remove lines 68-73 OR update them to use `$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager` instead of just `sdkmanager`. You would have to add a note **there** that this works only for users that have the `cmdline-tools` package installed.

henryhobhouse: I've removed it. As you say it would be nice to align with the linux docs. 

If anyone wishes to use the CLI they can go to the Android docs which should remain up to date and reduce maintenance on this codebase.

PR #2995, opened 2022-03-07
Clarify intended audience for architecture overview

sammy-SC:
Adding a sentence to clarify that it is not needed to understand architecture overview to use React Native.

Based on feedback: https://github.com/reactwg/react-native-new-architecture/discussions/1#discussioncomment-2265361


Simek (approved):
Looks like nice clarification, thanks! 👍 

Also, I think that we should consider expanding this page a bit in the future. 🙂

PR #2961, opened 2022-02-09
Moving Release Wiki to react-native-website

Megatron4537:
The release wiki can now be found under the community tab

A few things to note:

- The page url changed from  **/help** to **/docs/community**. We can change the suffix but for the page to have a left nav bar it needs the prefix **docs**
- The page release.md becomes redundant we could consider removing it.

[#33014](https://github.com/facebook/react-native/issues/33014)

PR #3000, opened 2022-03-09
Unversion Architecture docs

slorber:
Move out the Architecture tabs docs/sidebar from the versioned docs, as discussed with @cortinico @kelset @Simek 

I created a second Docusaurus docs plugin instance and moved the Architecture docs there + removed the versioned Architecture docs.

As far as I understand you eventually plan to add more unversioned docs, so I named this instance (and folders) `docs_unversioned` so that it is future proof, as it was mentioned some other docs (contributing, tests...) could also be unversioned later.

Also used `routeBasePath: '/'` so that all those docs can have a top-level url: using frontMatter `slug: /architecture` means the doc will be hosted at `/architecture` and not `/docs_unversioned/architecture`. It's up to you to decide if all unversioned docs must be hosted under a shared path segment prefix.

Obviously, the older versioned URLs are slightly different compared to before, so make sure to review this part if you want.
We can set up server-side redirects through Netlify's `_redirect` file if you feel it's necessary

Deploy preview: https://deploy-preview-3000--react-native.netlify.app/architecture/overview


Simek:
@slorber Can't we just move the files to the `website` root, and put them in the architecture directory, just like it has been done in Contributing PR? This will allow use to ditch `docs_unversioned` in path, which I'm not a big fan of.

slorber:
> @slorber Can't we just move the files to the `website` root, and put them in the `architecture` directory, just like it has been done in Contributing PR? This will allow use to ditch `docs_unversioned` in path, which I'm not a big fan of, and simplifies the setup. See:
> 
> * [Moving Release Wiki to react-native-website #2961](https://github.com/facebook/react-native-website/pull/2961)

Didn't know there was already such a PR

We can definitively do that. 

It depends if you want to gather all unversioned docs into the same folder and under the same url prefix (can be `/`) or if you want to use multiple unversioned docs instances (more config/boilerplate, but more flexibility)

Note: Docusaurus currently doesn't support 2 plugin instances using both `routeBasePath: '/'` at the same time.

slorber:
I updated the PR to match the choices already made on the other PR https://github.com/facebook/react-native-website/pull/2961

=> one docs plugin instance per unversioned type of content

lunaleaps (approved):
lgtm! Will let you and @Simek coordinate how to merge

Simek:
I think that the last thing which left there is to define redirect for the moved pages, since the in all communication related to the New Arch before we have used the versioned links. It would be nice to keep them working, without the redirects the links will lead to 404.

website/static/_redirects
```diff
@@ -12,3 +12,17 @@
...
+/docs/architecture-overview             /architecture/overview
+/docs/fabric-renderer                   /architecture/fabric-renderer
+/docs/render-pipeline                   /architecture/render-pipeline
+/docs/xplat-implementation              /architecture/xplat-implementation
+/docs/view-flattening                   /architecture/view-flattening
+/docs/threading-model                   /architecture/threading-model
+/docs/architecture-glossary             /architecture/glossary
+
+
+
```
Simek, line 28:
```suggestion
# Architecture pages move to unversioned docs
/docs/architecture-overview             /architecture/overview
/docs/fabric-renderer                   /architecture/fabric-renderer
/docs/render-pipeline                   /architecture/render-pipeline
/docs/xplat-implementation              /architecture/xplat-implementation
/docs/view-flattening                   /architecture/view-flattening
/docs/threading-model                   /architecture/threading-model
/docs/architecture-glossary             /architecture/glossary

/docs/next/architecture-overview             /architecture/overview
/docs/next/fabric-renderer                   /architecture/fabric-renderer
/docs/next/render-pipeline                   /architecture/render-pipeline
/docs/next/xplat-implementation              /architecture/xplat-implementation
/docs/next/view-flattening                   /architecture/view-flattening
/docs/next/threading-model                   /architecture/threading-model
/docs/next/architecture-glossary             /architecture/glossary

/docs/0.66/architecture-overview             /architecture/overview
/docs/0.66/fabric-renderer                   /architecture/fabric-renderer
/docs/0.66/render-pipeline                   /architecture/render-pipeline
/docs/0.66/xplat-implementation              /architecture/xplat-implementation
/docs/0.66/view-flattening                   /architecture/view-flattening
/docs/0.66/threading-model                   /architecture/threading-model
/docs/0.66/architecture-glossary             /architecture/glossary
```

A chunk of new redirects there, but they should cover all of the cases IIRC.

Also what's the difference between those redirect and https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects ?

kelset (commented):
thanks for all your work 👏👏👏

One thing I was wondering - could we maybe add to the readme.md (or the contributing.md) for this repo the instructions for how to add an unversioned section (/move something to unversioned)? I mean, I'm not expecting many more aside from this and the contributing one to be unversioned, but would be good to have some reference to look at.

By taking a quick look at this PR, it seems that there are a few things do to:
* the .md files need to be in their own folder at root level of the `website` folder
* `docusaurus.config.js` need a direct reference
* the category needs to not be in the `sidebars.json` but in their own sidebars<nameOfCategory>.json
* might have to add some references in `static/_redirects` to ensure that old urls get pointed to the new places

anything else?

cortinico (approved):
That's amazing, thanks for tackling this @slorber 

I've quickly checked the setup and it looks good to me. I'd defer the full review to @Simek and @kelset which already pointed out the most important changes.

website/static/_redirects
slorber: image

Those are server-side Netlify redirects (not related to Docusaurus at all, it is CDN config): https://docs.netlify.com/routing/redirects/

slorber:
👍 I'll let @Simek follow-up 

Docusaurus has a dedicated doc page for docs plugin multi-instance, which is exactly what we are using here: https://docusaurus.io/docs/docs-multi-instance

But it's possible to be more specific to the RN website use case, or eventually improve our official doc if you feel something should be clarified

Simek:
I have backported the changes to the New Arch pages which were causing the conflicts, I have also extracted the common docs options to the separate object which is then spread into all of the docs configs. After those changes and PR seems to be ready to merge.

I have also tested the redirects on the deploy preview and everything seems to be alright.

Big thanks for the @slorber for all the work he has done here, for the tips and guidance, also thanks to the others who posted the reviews and provided great feedback! ❤️ 

Let's ship it! 🚀

kelset:
@Simek just to be clear:
* contributing section is still versioned
* readme still doesn't have info about unversioning stuff
right? Did I miss something?

Not a big deal, we can handle with a follow up PR, just want to make sure we're not losing pieces down the road

Simek:
> * contributing section is still versioned

It should be unversioned, as it was when I was testing the PR last time.

> * readme still doesn't have info about unversioning stuff
> Not a big deal, we can handle with a follow up PR, just want to make sure we're not losing pieces down the road

Nope, but we can expand few sentences here and there, as well as update the directories structure schema in Readme. But generally I think that linking to the Docususaurus docs for the non-custom stuff is a good approach. Feel free to propose PR here and to the Docusuaurs website (if needed). If you not interested in this, I will add updating the instructions to my TODOs and look at this soon.

kelset:
sounds good, thanks for the clarifications Simek!

PR #3002, opened 2022-03-09
Hyphenation normal enum

zegermouw:
Added the normal enum to all docs where android_hyphenationFrequency was used. 
As asked in [2982](../2982)


Simek (approved):
Thanks @zegermouw for picking up this task, great work! 👍

PR #3006, opened 2022-03-11
Corrected outdated command

AbdoAlghanim:
I tried to use the "expo init" command after installng expo globally and I was simply greeted with an error that stated : "bash: expo: command not found".
After reading more about the problem on the internet, I concluded that the correct command is "npx expo".
You may want to search more about the problem but the issue is old and is more common on MacOS.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -42,7 +42,7 @@ Then run the following commands to create a new React Native project called "Awe
 <TabItem value="npm">
 
 ```shell
-expo init AwesomeProject
+npx expo init AwesomeProject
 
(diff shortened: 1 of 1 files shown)


pushpender-singh-ap:
When we use expo-cli no need `npx expo init projectname`
It's working `expo init projectname` without `npx`

Simek:
I can confirm what the other said in responses, the `expo` command should not be used with `npx`, you need to globally install `expo` package first. 

I'm going to close this one, but if you are still experiencing the problems feel free to report issue in the Expo repo:
* https://github.com/expo/expo/issues

PR #3017, opened 2022-03-16
📚Improve documentation about JS Specs

cipolleschi:
This PR improve the documentation about creating specs in JS for the new architecture.

* It specifies that the Turbomodules spec file must be named `Native<MODULE NAME>.js`
* It adds a paragraph on how to write a Spec for a fabric component
* It specifies that a Fabric Component spec file must be named `<MODULE NAME>NativeComponent.js`
* It adds an example for a Fabric Component spec.


dmitryrykun (approved):

cortinico (approved):
Awesome stuff!

PR #3022, opened 2022-03-18
WIP: upgrade Docusaurus

slorber:
Upgrade of Docusaurus:
- upgrade Algolia config (cd https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration)
- upgrade Google Analytics location
- migrate some CSS selector breaking changes (cf https://github.com/facebook/docusaurus/pull/6925)

---

WIP as there are some remaining CSS issues to fix

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -42,9 +42,9 @@
     ]
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-beta.14",
-    "@docusaurus/plugin-pwa": "^2.0.0-beta.14",
-    "@docusaurus/preset-classic": "^2.0.0-beta.14",
+    "@docusaurus/core": "0.0.0-4738",
+    "@docusaurus/plugin-pwa": "0.0.0-4738",
+    "@docusaurus/preset-classic": "0.0.0-4738",
     "docusaurus-plugin-sass": "^0.2.1",
     "esbuild-loader": "2.13.1",
     "react": "^17.0.2",
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -59,6 +59,12 @@ module.exports = {
       '@docusaurus/preset-classic',
       /** @type {import('@docusaurus/preset-classic').Options} */
       ({
+        googleAnalytics: {
+          trackingID: 'UA-41298772-2',
+        },
+        gtag: {
+          trackingID: 'UA-41298772-2',
+        },
         docs: {
           path: '../docs',
           sidebarPath: require.resolve('./sidebars.json'),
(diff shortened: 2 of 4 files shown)


website/package.json
```diff
@@ -42,9 +42,9 @@
     ]
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-beta.14",
-    "@docusaurus/plugin-pwa": "^2.0.0-beta.14",
-    "@docusaurus/preset-classic": "^2.0.0-beta.14",
+    "@docusaurus/core": "0.0.0-4738",
+    "@docusaurus/plugin-pwa": "0.0.0-4738",
+    "@docusaurus/preset-classic": "0.0.0-4738",
```
Simek, line 47: Can we wait until next `beta` is released? The class name issue doesn't seems that crucial to stick to nightlines. 😅 

Saying that, I was planning to upgrade to latest beta this weekend, so if changing that to `beta.17` also helps that would be fine.

Simek:
Looking at the deploy, there are few thing which doesn't seems correct right now:
* hover effect for the new theme toggler
* new top page navigation readability could be improved, especially for active element in light mode (I also don't like the hover effect there, for non interactable parts, like top level sidebar groups, which can be misleading)
* it looks like there is regression which sidebar headers styling in the dark mode (dark font is used)

Also, personally I think that including the page name in this navigation is redundant and looks a bit weird, when title is listed twice, and both those text are non-interactable. Besides the "home" link it doesn't have any other function, besides displaying navigation structure:

Screenshot 2022-03-18 194226

Is it possible to turned this feature off? Not sure if we want that, but it would be nice to know.

slorber:
Thanks for the review

This is a WIP PR to help you with the breaking changes and so that I know what else breaks on upgrade.

I'll be on holiday and will do a release before so if you want to give it a try until then let me know, otherwise we'll wait 2 weeks 😄

Josh-Cena:
> Besides the "home" link (which do the same as logo) the new component doesn't add any other function, besides displaying navigation structure:

That's because the RN website currently doesn't take advantage of the new category index pages. If a category has a link associated, the respective breadcrumb item will be clickable. But given the flat nature of the website's navigation system, maybe the breadcrumbs aren't useful at all?

Josh-Cena:
The color mode toggle hover style is a bug with Docusaurus. It only takes the default navbar style into account

Simek:
Thanks for the explanation @Josh-Cena! 

Maybe we leverage this feature in the future, while reworking Architecture and Contributing parts, but for now I have disabled them in my PR, via `breadcrumbs: false` option added to all instances of the `docs` plugin.

slorber:
Replaced by https://github.com/facebook/react-native-website/pull/3059

PR #3025, opened 2022-03-23
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/sync-api-docs/generateMarkdown.js
+++ b/sync-api-docs/generateMarkdown.js
@@ -281,29 +281,29 @@ function preprocessDescription(desc) {
   });
 
   if (tabs === 2) {
-    const firstExample = desc.substr(desc.search('```SnackPlayer') + 1);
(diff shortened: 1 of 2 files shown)


Simek (approved):
Hello @CommanderRoot, thank you for the update. 👍

Just FYI, the code you have been editing is not used on the website and it was a test project during one of MLH batches (to extract documentation from the RN flow files), which seems to be discontinued for now. We left the code in, just in case it will come back on agenda in the future.

PR #3031, opened 2022-03-25
update project structure in Readme file

Simek:
# Why

The update of the project structure in the Readme file has been request by @kelset here:
* https://github.com/facebook/react-native-website/pull/3000#issuecomment-1064201720

# How

I have updated the project structure to match latest changes and I have switched a bit the way in witch the structure is displayed, removing the example files and replacing them with more generic descriptions which I hope clarify the placement of files a bit more.

I have also updated the paragraphs around the graph so they refer to the updated structure, however I feel like those parts can be phrased a bit more clearly, but I have tried my best as a non-native speaker. 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,11 @@ If you are looking for the source code of the [React Native Archive website](htt
 
 ## 📖 Overview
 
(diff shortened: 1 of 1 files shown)


cortinico (approved):
LGTM

PR #3032, opened 2022-03-25
hide VersionDropdown on non-versioned pages, reduce the visible versions list

Simek:
# Why & How

Since latest changes to the docs structure and introducing multiple docs instances part of the website is versioned, and part of the website is static.

Due to those changes I feel like the `VersionDropdown` adds a bit of unnecessary confusion, so I have decided to swizzle it and propose a change that hides the version selector from all of the non-versioned pages. The selector is only visible when viewing the pages from `Guides`, `APIs` and `Components`. This might not be in line with other Docususaurus sites, but I feel like it's a better UX in our case.

I have decided to swizzle because AFAIK Docusuaurs do not offer a configs for the changes I would like to achieve. 

The code I have added has been annotated with comment starting with `(CUSTOM)`, so we can identify the part we edited during Docusuaurs updates in the future.

In addition to that, I have reduced the number of visible versions in the dropdown to 5 latest releases + `next`.

## Preview
Screenshot 2022-03-25 at 14 12 06
Screenshot 2022-03-25 at 14 11 41

### DIFF
--- a/website/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
+++ b/website/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
@@ -15,8 +15,8 @@ import {
 import {useDocsPreferredVersion} from '@docusaurus/theme-common';
 import {translate} from '@docusaurus/Translate';
 
-const getVersionMainDoc = (version) =>
-  version.docs.find((doc) => doc.id === version.mainDocId);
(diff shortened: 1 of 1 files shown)


cortinico (approved):
Change looks good to me. I'm not a web expert but we shoudl be good merging this 👍

Simek:
Let's ship this then, we can always revert this commit, if the change breaks anything. 😉

PR #3039, opened 2022-03-30
[Guide - The New Architecture] TurboModules as Native Modules

cipolleschi:
This PR adds the guide to create a backward compatible TurboModule. 

Draft while we wait for Android.

### DIFF
--- a/docs/the-new-architecture/backward-compatibility-turbomodules.md
+++ b/docs/the-new-architecture/backward-compatibility-turbomodules.md
@@ -3,10 +3,365 @@ id: backward-compatibility-turbomodules
 title: TurboModules as Native Modules
 ---
 
-This section describes the required steps to ensure that a TurboModule can be used as a Native Module.
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import constants from '@site/core/TabsConstants';
+import BetaTS from './\_markdown_beta_ts_support.mdx';
 
-The section explains:
+:::info
+The creation of a backward compatible TurboModule requires the knowledge of how to create a TurboModule. To recall these concepts, have a look at this [guide](pillars-turbomodules).
 
-- How to avoid installing dependencies when they are not needed
-- The usage of compilation pragmas to avoid compiling code that requires types from the codegen
-- API uniformity in JS, so that they don’t have to import different files
+TurboModules only works when the New Architecture is properly setup. If you already have a library that you want to migrate to the New Architecture, have a look at the [migration guide](../new-architecture-intro) as well.
+:::
+
+Creating a backward compatible TurboModule lets your users continue leverage your library, independently from the architecture they use. The creation of such a module requires a few steps:
+
+1. Configure the library so that dependencies are prepared set up properly for both the Old and the New Architecture.
+1. Update the codebase so that the New Architecture types are not compiled when not available.
+1. Uniform the JavaScript API so that your user code won't need changes.
+
+<BetaTS />
+
+While the last step is the same for all the platforms, the first two steps are different for iOS and Android.
+
+## Configure the TurboModule Dependencies
+
+### <a name="dependencies-ios" />iOS
+
+The Apple platform installs TurboModules using [Cocoapods](https://cocoapods.org) as dependency manager.
+
+Every TurboModule defines a `podspec` that looks like this:
+
+```ruby
+require "json"
+
+package = JSON.parse(File.read(File.join(__dir__, "package.json")))
+
+folly_version = '2021.06.28.00-v2'
+folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
+
+Pod::Spec.new do |s|
+  # Default fields for a valid podspec
+  s.name            = "<TM Name>"
+  s.version         = package["version"]
+  s.summary         = package["description"]
+  s.description     = package["description"]
+  s.homepage        = package["homepage"]
+  s.license         = package["license"]
+  s.platforms       = { :ios => "11.0" }
+  s.author          = package["author"]
+  s.source          = { :git => package["repository"], :tag => "#{s.version}" }
+
+  s.source_files    = "ios/**/*.{h,m,mm,swift}"
+  # React Native Core dependency
+  s.dependency "React-Core"
+
+  # The following lines are required by the New Architecture.
+  s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
+  s.pod_target_xcconfig    = {
+      "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
+      "CLANG_CXX_LANGUAGE_STANDARD" => "c++17"
+  }
+
+  s.dependency "React-Codegen"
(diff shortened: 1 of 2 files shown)


cortinico:
> Draft while we wait for Android.

Are you fine with me pushing on your branch?

cipolleschi:
@cortinico Yes, push freely on this branch!

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
+Whether you are using Flow or TypeScript for your specs, we understand which architecture is running by checking whether the `global.__turboModuleProxy` object has been set or not.
+
+:::caution
+The `global.__turboModuleProxy` API may change in the future for a function that encapsulate this check.
+:::
+
+- If that object is `null`, the app has not enabled the TurboModule feature. It's running on the Old Architecture, and the fallback is to use the default [`NativeModule` implementation](../native-modules-intro).
+- If that object is set, the app is running with the TurboModules enabled and it should use the `Native<MyModule>` spec to access the TurboModule.
+
+<BetaTS />
```
cortinico, line 365: This is quite late in the page and can be lost. Should probably be moved ahead?

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
 
-- How to avoid installing dependencies when they are not needed
-- The usage of compilation pragmas to avoid compiling code that requires types from the codegen
-- API uniformity in JS, so that they don’t have to import different files
+TurboModules only works when the New Architecture is properly setup. If you already have a library that you want to migrate to the New Architecture, have a look at the [migration guide](../new-architecture-intro) as well.
+:::
+
+Creating a backward compatible TurboModule lets your users continue leverage your library, independently from the architecture they use. The creation of such a module requires a few steps:
+
+1. Configure the library so that dependencies are not installed in the old architecture.
```
cortinico, line 19: Q: This is a bit unclear to me. It seems to be mostly iOS focused (there is no reference to 'dependencies' for Android). Perhaps a more generic wording like:

> 1. Configure the library so that it can expose the correct APIs for both old and new architectures?

Or "so that it configures correctly for both old and new..."

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
+    s.dependency "ReactCommon/turbomodule/core"
++ end
+end
+```
+
+This `if` guard prevents the dependencies from being installed when the environment variable is not set.
+
+### Android
+
+To create a module that can work with both architecture, you need to configure gradle to choose which files need to be compiled depending on the chosen architecture. This can be achieved by using different source sets in the gradle configuration.
```
cortinico, line 101:
```suggestion
To create a module that can work with both architectures, you need to configure Gradle to choose which files need to be compiled depending on the chosen architecture. This can be achieved by using **different source sets** in the Gradle configuration.
```

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
+
+:::note
+This approach can lead to duplicated code, but it is the safest and cleanest way to achieve the desired result. You will see how to reduce the duplication in the next section.
```
cortinico, line 104: I would reword to:

> Please note that this is currently the **suggested** approach. While it might lead to some code duplication, it will ensure the maximum compatibility with both architectures. You will see how to reduce the duplication in the next section.

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
++            }
++        }
+    }
+}
+```
+
+This changes do three main things:
+
+1. The first lines define a function that returns whether the New Architecture is enabled or not.
+2. The `buildConfigField` line defines a build configuration boolean field called `IS_NEW_ARCHITECTURE_ENABLED`, and initialize it using the function declared in the first step.
```
cortinico, line 139:
```suggestion
2. The `buildConfigField` line defines a build configuration boolean field called `IS_NEW_ARCHITECTURE_ENABLED`, and initialize it using the function declared in the first step. This will allow you to check at runtime if a user has specified the `newArchEnabled` property or not.
```

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
++ #endif
+
+@end
+```
+
+This snippet uses the same `RCT_NEW_ARCH_ENABLED` flag used in the previous [section](#dependencies-ios). When this flag is not set, Xcode skips the lines within the `#ifdef` during compilation and it does not include them into the compiled binary.
+
+### Android
+
+To create a backward compatible TurboModule you have two different source sets that are loaded and compiled depending on the proper architecture.
```
cortinico, line 179:
```suggestion
As we can't use conditional compilation blocks on Android, we will define two different source sets. This will allow to create a backward compatible TurboModule with the proper source that is loaded and compiled depending on the used architecture.
```

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
+
+Therefore, you have to:
+
+1. Create a Native Module in the `src/oldarch` path. See [this guide](../native-modules-intro) to learn how to create a Native Module.
+2. Create a TurboModule in the `src/newarch` path. See [this guide](./pillars-turbomodules) to learn how to create a TurboModule
+
+and then gradle could decide which implementation to pick.
```
cortinico, line 186:
```suggestion
and then instruct Gradle to decide which implementation to pick.
```

cortinico (approved):
Great stuff. Mostly nits 👍

docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -3,10 +3,363 @@ id: backward-compatibility-turbomodules
...
+
+1. The first lines define a function that returns whether the New Architecture is enabled or not.
+2. The `buildConfigField` line defines a build configuration boolean field called `IS_NEW_ARCHITECTURE_ENABLED`, and initialize it using the function declared in the first step.
+3. The last lines leverage the function declared in step one to decide which source sets we need to build, depending on the choosen architecture.
+
+## Update the codebase
+
+### iOS
+
+The second step is to instruct Xcode to avoid compiling all the lines using the New Architecture types and files when we are building an app with the Old Architecture.
```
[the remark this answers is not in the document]
cipolleschi: If I search for `new ` in VS code, it only finds an instance of it inside the `podfile`, which is correct to be in lowercase. Could you point out where you found the `new` instances?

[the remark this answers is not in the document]
cortinico: 1. _markdown_beta_ts_support.mdx has one lowercase
2. Line 19 has a mixture: `1. Configure the library so that dependencies are prepared set up properly for both the old and the New Architecture.`

PR #3045, opened 2022-03-30
docs: update running ios simulators 0.68 version

luism3861:
@Simek  i forgot update specially 0.68 version.


Simek (approved):
Thank you for the update and great attention to details as always @luism3861! 👍

PR #3057, opened 2022-04-06
fix website GA config

Simek:
Refs:
* https://github.com/facebook/docusaurus/pull/5832

This PR moves the GA config to the preset to fix the degraded data issue.

CC @slorber

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -91,6 +91,12 @@ module.exports = {
             require.resolve('./src/css/versions.scss'),
           ],
         },
+        googleAnalytics: {
+          trackingID: 'UA-41298772-2',
+        },
+        gtag: {
+          trackingID: 'UA-41298772-2',
+        },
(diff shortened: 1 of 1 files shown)


lunaleaps:
Whoa thank you @kelset and @Simek for tracking this down and fixing! Will monitor if this fixes, thank you guys again!

slorber:
Note: we'll soon have to migrate all sites to Google Analytics 4, as former analytics are being shut down by Google in 2023

https://github.com/facebook/docusaurus/issues/7221

As the data can't be migrated from old to new analytics, it may be better to run old/new analytics together for this year, otherwise the transition will more brutal if you rely on analytics


I have no idea who manages the analytics accounts at Meta and how it's used 🤪 just wanted you to be aware @Simek @lunaleaps

lunaleaps:
Thanks for the heads up! I'll make an issue for this

PR #3058, opened 2022-04-06
Fix GA

kelset:
This fixes the problem we are seeing with GA, per https://github.com/facebook/docusaurus/pull/5832

Initially I also tried to bump docusaurus to the latest beta too but some unrelated things were breaking, so I decided to just rollback and do simply the config move.


kelset:
nvm - closing this off, @Simek beat me by a few mins 🤣 I knew I shouldn't have tried bumping docusaurus 🤣 https://github.com/facebook/react-native-website/pull/3057

PR #3065, opened 2022-04-07
Fix iOS and Android labels under accessibilityHint

anastasialanz:
These labels were starting the sentence and made it look off.

Before: 
https://reactnative.dev/docs/accessibility#accessibilityhint
Screen Shot 2022-04-07 at 5 07 24 PM

After: 
Screen Shot 2022-04-08 at 8 50 28 AM

## Validation
- Pull down this branch
- Run `yarn start`
- Go to [http://localhost:3000/docs/next/accessibility](http://localhost:3000/docs/next/accessibility)
- Verify that the iOS and Android labels look correct

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -63,9 +63,11 @@ To use, set the `accessibilityHint` property to a custom string on your View, Te
 ```
 
 <div class="label ios basic">iOS</div>
+
 In the above example, VoiceOver will read the hint after the label, if the user has hints enabled in the device's VoiceOver settings. Read more about guidelines for `accessibilityHint` in the [iOS Developer Docs](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint)
 
 <div class="label android basic">Android</div>
+
 In the above example, TalkBack will read the hint after the label. At this time, hints cannot be turned off on Android.
 
 ### `accessibilityLanguage` <div class="label ios">iOS</div>
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -63,7 +63,7 @@ To use, set the `accessibilityHint` property to a custom string on your View, Te
 ```
 
 <div class="label ios basic">iOS</div>
-In the above example, VoiceOver will read the hint after the label, if the user has hints enabled in the device's VoiceOver settings. Read more about guidelines for accessibilityHint in the [iOS Developer Docs](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint)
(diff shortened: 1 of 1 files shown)


docs/accessibility.md
```diff
@@ -62,9 +62,11 @@ To use, set the `accessibilityHint` property to a custom string on your View, Te
 </TouchableOpacity>
 ```
 
-iOS In the above example, VoiceOver will read the hint after the label, if the user has hints enabled in the device's VoiceOver settings. Read more about guidelines for accessibilityHint in the [iOS Developer Docs](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint)
+#### <div class="label ios">iOS</div>
+In the above example, VoiceOver will read the hint after the label, if the user has hints enabled in the device's VoiceOver settings. Read more about guidelines for accessibilityHint in the [iOS Developer Docs](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint)
```
Simek, line 66:
```suggestion
<div class="label ios basic">iOS</div>
In the above example, VoiceOver will read the hint after the label, if the user has hints enabled in the device's VoiceOver settings. Read more about guidelines for `accessibilityHint` in the [iOS Developer Docs](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint)
```

We have an additional class `basic` which removes the skeuomorphic start of label, I think it will work better here. Also wrapping with H4 do no add anything, only produces the weird hash links, let's just skip those tags.

docs/accessibility.md
```diff
@@ -62,9 +62,11 @@ To use, set the `accessibilityHint` property to a custom string on your View, Te
...
 
-Android In the above example, TalkBack will read the hint after the label. At this time, hints cannot be turned off on Android.
+#### <div class="label android">Android</div>
```
Simek, line 68:
```suggestion
<div class="label android basic">Android</div>
```

Same as above.

docs/accessibility.md
anastasialanz: I didn't know that existed! Thank you!

anastasialanz:
@Simek Thanks for the feedback! I updated the PR with the correct labels.

Simek (changes requested):
Thank you for the updates! 🙏 

However, it looks like the headers were ensuring the block display of labels, so we need to wrap them with `<p>` or other block tag, to fix the display:
Screenshot 2022-04-08 165718

Maybe also an empty line between `<div>`s and text will help parser to distinguish HTML from MDX parts.

anastasialanz:
> Thank you for the updates! 🙏
> 
> However, it looks like the headers were ensuring the block display of labels, so we need to wrap them with `<p>` or other block tag, to fix the display: Screenshot 2022-04-08 165718
> 
> Maybe also an empty line between `<div>`s will help parser to distinguish HTML from MDX parts.

I added a space between the labels and text and it looks like that fixed the spacing issue.

Simek (approved):
LGTM, let's ship this! 🚀

PR #3068, opened 2022-04-10
Add Kotlin examples in the "Android Native UI Components" documentation

RickardZrinski:
Updates the [Android Native UI Components](https://reactnative.dev/docs/native-components-android) page with Kotlin snippets.


docs/native-components-android.md
```diff
@@ -44,31 +65,71 @@ public class ReactImageManager extends SimpleViewManager<ReactImageView> {
...
+Properties that are to be reflected in JavaScript needs to be exposed as setter method annotated with `@ReactProp` (or `@ReactPropGroup`). Setter method should take view to be updated (of the current view type) as a first argument and property value as a second argument. Setter should be declared as a `void` method in Java or with return type `Unit` (or omitted) in Kotlin and should be `public`. Property type sent to JS is determined automatically based on the type of value argument of the setter. The following type of values are currently supported (in Java): `boolean`, `int`, `float`, `double`, `String`, `Boolean`, `Integer`, `ReadableArray`, `ReadableMap`. The corresponding types in Kotlin are `Boolean`, `Int`, `Float`, `Double`, `String`, `ReadableArray`, `ReadableMap`.
```
RickardZrinski, line 98: I wasn't quite sure how to formulate this section. Especially the `void` / `Unit` part. Suggestions are appreciated!

It might be overly explicit to list all the types as well?

RickardZrinski:
@cortinico, @ShikaSD, @mdvacca

docs/native-components-android.md
```diff
@@ -44,31 +65,71 @@ public class ReactImageManager extends SimpleViewManager<ReactImageView> {
...
 
 <!-- alex enable primitive -->
 
+<Tabs groupId="android-language" defaultValue={constants.defaultAndroidLanguage} values={constants.androidLanguages}>
+<TabItem value="kotlin">
+
+```kotlin
+  @ReactProp(name = "src")
+  fun setSrc(view: ReactImageView, sources: ReadableArray?) {
+    view.setSource(sources)
```
cortinico, line 116:
```suggestion
view.sources = sources
```

I believe the IDE would suggest you to use property accessors. Could you cross check?

docs/native-components-android.md
```diff
@@ -44,31 +65,71 @@ public class ReactImageManager extends SimpleViewManager<ReactImageView> {
...
+  }
+
+  @ReactProp(name = "borderRadius", defaultFloat = 0f)
+  override fun setBorderRadius(view: ReactImageView, borderRadius: Float) {
+    view.setBorderRadius(borderRadius)
```
cortinico, line 121:
```suggestion
    view.borderRadius = borderRadius
```

I believe the IDE would suggest you to use property accessors. Could you cross check?

docs/native-components-android.md
```diff
@@ -44,31 +65,71 @@ public class ReactImageManager extends SimpleViewManager<ReactImageView> {
...
+  }
+
+  @ReactProp(name = ViewProps.RESIZE_MODE)
+  fun setResizeMode(view: ReactImageView, resizeMode: String?) {
+    view.setScaleType(ImageResizeMode.toScaleType(resizeMode))
```
cortinico, line 126:
```suggestion
    view.scaleType = ImageResizeMode.toScaleType(resizeMode)
```

Same as above: I believe the IDE would suggest you to use property accessors. Could you cross check?

docs/native-components-android.md
```diff
@@ -86,9 +147,26 @@ Setter declaration requirements for methods annotated with `@ReactPropGroup` are
   }
 ```
 
+</TabItem>
+</Tabs>
+
 ### 4. Register the `ViewManager`
 
-The final Java step is to register the ViewManager to the application, this happens in a similar way to [Native Modules](native-modules-android.md), via the applications package member function `createViewManagers.`
+The final Java/Kotlin step is to register the ViewManager to the application, this happens in a similar way to [Native Modules](native-modules-android.md), via the applications package member function `createViewManagers.`
```
cortinico, line 155:
```suggestion
The final step is to register the ViewManager to the application, this happens in a similar way to [Native Modules](native-modules-android.md), via the applications package member function `createViewManagers`
```

docs/native-components-android.md
```diff
@@ -86,9 +147,26 @@ Setter declaration requirements for methods annotated with `@ReactPropGroup` are
...
+
+<Tabs groupId="android-language" defaultValue={constants.defaultAndroidLanguage} values={constants.androidLanguages}>
+<TabItem value="kotlin">
+
+```kotlin
+  override fun createViewManagers(
+      reactContext: ReactApplicationContext
+  ): List<ViewManager<*, *>> {
+    return listOf(ReactImageManager(reactContext))
+  }
```
cortinico, line 165:
```suggestion
  override fun createViewManagers(
      reactContext: ReactApplicationContext
  ) = listOf(ReactImageManager(reactContext))
```

docs/native-components-android.md
```diff
@@ -273,8 +488,121 @@ public class MyFragment extends Fragment {
...
+import com.facebook.react.uimanager.ThemedReactContext
+import com.facebook.react.uimanager.ViewGroupManager
+import com.facebook.react.uimanager.annotations.ReactPropGroup
+import kotlin.properties.Delegates
+
+class MyViewManager(
+    private val reactContext: ReactApplicationContext
+) : ViewGroupManager<FrameLayout>() {
+  private var propWidth by Delegates.notNull<Int>()
+  private var propHeight by Delegates.notNull<Int>()
```
cortinico, line 519: Not sure about this. Should they just be `var Int?`?

docs/native-components-android.md
```diff
@@ -273,8 +488,121 @@ public class MyFragment extends Fragment {
...
+    val myFragment = MyFragment()
+    val activity = reactContext.currentActivity as FragmentActivity
+    activity.supportFragmentManager
+        .beginTransaction()
+        .replace(reactNativeViewId, myFragment, reactNativeViewId.toString())
+        .commit()
+  }
+
+  fun setupLayout(view: View) {
+    Choreographer.getInstance().postFrameCallback(object: Choreographer.FrameCallback {
```
cortinico, line 572: I believe here we could use SAM conversion (i.e. you don't need `object` and the `override` method declaration.

docs/native-components-android.md
```diff
@@ -407,8 +735,35 @@ public class MyViewManager extends ViewGroupManager<FrameLayout> {
...
+
+import com.facebook.react.ReactPackage
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.uimanager.ViewManager
+
+class MyPackage : ReactPackage {
+  ...
+  override fun createViewManagers(
+      reactContext: ReactApplicationContext
+  ): List<ViewManager<*, *>> {
```
cortinico, line 758: Similarly this could be simplified

cortinico (commented):
Great work 👍 Really like how the components code looks like with Kotlin. Left some suggestions and with them we should be good to go

docs/native-components-android.md
cortinico: Added a suggestion on how to improve it

RickardZrinski:
Cheers for the comments! I'll read through and fix them during the weekend.

docs/native-components-android.md
RickardZrinski: It doesn't, no. `setSource` doesn't have a corresponding `getSource` so there's no property accessor. Both must exist for a property accessor to be available. Makes sense but TIL (new to Kotlin)!

docs/native-components-android.md
RickardZrinski: Can do. Delegates are quite nice because we delegate the null-handling (accessing them before they're initialized will throw an exception with a nice message) and we don't have to spam `requireNotNull` each and every time we access the variables. In this case we just access them in a single function so it doesn't offer a lot of benefit though.

And for the purposes of this tutorial it might be too confusing / advanced so yes, let's make them nullable and use `requireNotNull`.

docs/native-components-android.md
RickardZrinski: That was my thought as well but it causes issues with scoping of `this`. We call `Choreographer.getInstance().postFrameCallback(this)` inside the lambda. `this` will reference our `MyViewManager` instead of the `Choreographer.FrameCallback`.

So unfortunately we need to use the `object` syntax in this case so that `this` is scoped correctly.

image

RickardZrinski: Might be possible through some other syntax though so be happy to change it into a lambda if you can find one. I couldn't though.

docs/native-components-android.md
RickardZrinski: I formulated it slightly different from your suggestion. Let me know if you prefer yours 👍

docs/native-components-android.md
cortinico: Makes sense 👍

docs/native-components-android.md
cortinico: You're right. As this is a java interface, we'll have to use this Compiler Plugin here:
https://kotlinlang.org/docs/sam-with-receiver-plugin.html
(which we can't assume our user are enabling)

cortinico:
Thanks for the follow up. There are still a couple of open thread, but we're heading in the right direction 👍

docs/native-components-android.md
RickardZrinski: Same as `setSource`. No property accessor due to no getter.

docs/native-components-android.md
RickardZrinski: Same as `setSource`. No property accessor due to no getter. (Confusingly there is a `scaleType` property accessor on the parent class `ImageView` but we want to call the `setScaleType` method on `ReactImageView`).

RickardZrinski:
> Thanks for the follow up. There are still a couple of open thread, but we're heading in the right direction 👍

I've closed two more threads (the property accessor ones). Sorry, they were hidden/collapsed in the GitHub UI so I didn't see them. Any other comments I've missed? There's one still open (as far as I can tell) but I didn't want to close it as I didn't formulate it exactly as you suggested so just wanted some feedback on it.

PR #3078, opened 2022-04-15
fix: mistake is function call.

DibyajyotiMishra:
there was a mistake which was accidentally introduced while calling the function. The function invocation missed the `this` keyword.

### DIFF
--- a/docs/optimizing-flatlist-configuration.md
+++ b/docs/optimizing-flatlist-configuration.md
@@ -136,7 +136,7 @@ render(){
 
   <FlatList
     data={items}
-    renderItem={renderItem}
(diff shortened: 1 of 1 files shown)


Simek (changes requested):
Hello @DibyajyotiMishra, nice catch, thank you for the update! 👍 

However, it looks to me, that you have edited the wrong example (functional one instead of class). Let's fix this, and we can ship the fix. 🙂

DibyajyotiMishra:
Oh yes! I accidentally updated the wrong one. Sorry!

Simek (approved):
Thank you for the correction! 🚀

Issue #3096, opened 2022-04-30
The banner "Support Ukraine 🇺🇦 Help Provide Humanitarian Aid to Ukraine." is too large


HyunseungLee-Travis:
## Description

The computer sized banner for React-Native website is good and it is very helpful to know.

However, in mobiles (starting from about width 460 px), the banner is too large.

Would be very helpful if the banner's text font size is decreased when the screen size is small.

## about Width 460
![width-460-px](https://user-images.githubusercontent.com/78584173/166107146-7b9ae71f-6999-4c52-9f6f-a78b1895d50d.png)

## about Width 311
![width-311-px](https://user-images.githubusercontent.com/78584173/166107214-38dfca87-4a71-446e-84e8-104f1688cb4d.png)

## How can we address it?

Please fix the banner's font size to make it smaller in Mobile/IPad/Watch Screens.

Or move them somewhere else so it does not take up the whole screen.

## Why is it important?

Very hard to scroll through contents. 

Especially since this banner appears everywhere and cannot be closed.

## Who needs this?

Everybody. From people learning, referencing the API docs, contributors and all the people who visits this website or uses React Native.

## When should this happen (use version numbers if needed)?

This should happen immediately.

## Code
```css
@media only screen and (max-width: 768px) {
  .announcement {
    font-size: 18px;
  }
}

@media only screen and (max-width: 500px) {
  .announcement {
    font-size: 15px;
    line-height: 22px;
    padding: 6px 30px;
  }
}
```

seems to be the code to make the fonts smaller, but I believe it needs to be `div[class^="announcementBarContent"]` instead of `.announcement`.

PR #3098, opened 2022-04-30
customTheme.css (Make Font size smaller in banner)

HyunseungLee-Travis:
# Issue

The class for the font size in media seems to be incorrect. It does not make the font size of the banner smaller.

# Before
before

# After
after


HyunseungLee-Travis:
Issue #3096

Simek (approved):
Hello @HyunseungLee-Travis, nice catch, thank you for the fix! 👍 

We probably should also consider removing period from the end of the announce, since it leads to the additional, third line in some width ranges.

PR #3102, opened 2022-05-01
Make the 'new' badge float right in mobile

HyunseungLee-Travis:
# Issue
before - mobile - ipad

As seen on the picture above, the 'new' badge is right next to 'pressable' text.

In current React Native website, the 'new' badge (on computers) floats right.

However, switching the screen to mobile/ipad, the badge appears next to the text instead of floating right.

# What changed
Using `margin-left: auto;` will make the button flat right inside the div in mobile screens.

# After
after - mobile - ipad

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1480,6 +1480,7 @@ td .label {
   position: relative;
   top: 1px;
   margin-right: -6px;
+  margin-left: auto;
   padding: 3px 6px;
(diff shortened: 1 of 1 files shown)


Simek (approved):
Hello @HyunseungLee-Travis, nice catch, this was the intended design from the start. 👍 

Unfortunately Docusaurus styles and selectors may change from release to release, which may lead to small regressions like that on update.

PR #3119, opened 2022-05-18
Suggestion for improvement on Mac M1 chipset

ScreamZ:
Currently, the documentation recommends installing cocoapods from gem.

Some users are asking for M1 support, especially on expo (eas-cli). Which makes it hard to play with rosetta CLI (`arch -x86_64 pod install`) bypass. (https://expo.canny.io/feature-requests/p/eas-build-support-for-mac-m1)

As seen here https://github.com/CocoaPods/CocoaPods/issues/9907 cocoapods has been updated to support arm64 m1 architecture, but you need to install it from brew.

### DIFF
--- a/docs/_getting-started-macos-ios.md
+++ b/docs/_getting-started-macos-ios.md
@@ -43,13 +43,15 @@ Using the default Ruby available on macOS will require you to use `sudo` when in
 
 ```shell
 sudo gem install cocoapods
+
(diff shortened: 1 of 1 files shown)


Simek:
Hello @ScreamZ, according to #3384 it looks like the M1 workaround is no longer needed, and we plan to remove the M1 notes all along, so I'm going to close this PR.

Feel free to jump into the conversation in the mentioned PR, if you think that we are missing something.

PR #3122, opened 2022-05-18
Remove "Add Babel Plugins" Section

dmitryrykun:
This PR removes The New Architecture/Enabling Fabric on iOS/[Add Babel Plugins](https://reactnative.dev/docs/next/new-architecture-app-renderer-ios#3-add-babel-plugins) section.
`babel-plugin-codegen` is not stabilized yet, and is not beneficial for Fabric as of now.
The plugin will become a part of the Metro preset in the future, so this section is effectively redundant.


cipolleschi (approved):
Wonderful, thanks for this PR!

PR #3129, opened 2022-05-21
Update PWA Manifest and Maskable images

HyunseungLee-Travis:
# Problem
Maskable images should have enough padding for the logos to appear when in minimum safe area masks.
manifest.json should include description, orientation and id.

# Why it is a problem
1. Best to show them in both Mobile and Desktop cases (description)
2. Automatically generated in Chrome (>98) but people using <98 might have a problem (id)
3. Don't have enough padding to show the React Native logo in minimum safe areas (maskable images)
4. Define orientation (orientation)

# Before Fix
before_description
before_id
before_maskable
before_orientation

# Fixed
after_description
after_id
after_maskable
after_orientation


Simek (changes requested):
Hi @HyunseungLee-Travis, thank you spotting the problem and for the updates for the PWA assets. 

However, those images are autogenerated by `pwa-asset-generator` package:
* https://github.com/facebook/react-native-website/blob/main/website/package.json#L30

Can you check if it possible to modify the command or source image in the way which produces similar result to your manual edit? Padding looks like the parameter we could add:
* https://github.com/onderceylan/pwa-asset-generator#readme

HyunseungLee-Travis:
Hmmm... Not sure but changing the pwa:generate command does not change the images.
Are these images generated during deployment or by the developers.
Also, I am not sure what I am doing wrong but the newly generated images have white background.
I only changed the `--padding` so I am not sure what I did wrong.

Simek:
Thank you for the update! 👍 

This is fine, just run `yarn pwa:generate` locally to regenerate the files, commit them and this PR should be ready to ship! 🙂

HyunseungLee-Travis:
Not really sure what I am doing here but the background color is white and splash images are generated instead of the manifest icons. Not sure if `pwa:generate` generates manifest icons.

website/package.json
```diff
@@ -27,7 +27,7 @@
     "language:lint": "cd ../ && alex && case-police 'docs/*.md' -p brands,general,products,softwares",
     "language:lint:versioned": "cd ../ && alex . && case-police '**/*.md' -p brands,general,products,softwares",
     "ci:lint": "yarn lint && yarn lint:versioned && yarn language:lint:versioned",
-    "pwa:generate": "npx pwa-asset-generator ./static/img/header_logo.svg ./static/img/pwa --background #20232a --icon-only -opaque true --maskable true --type png"
+    "pwa:generate": "npx pwa-asset-generator ./static/img/header_logo.svg ./static/img/pwa --padding 12.5% --background #20232a --icon-only -opaque true --maskable true --type png"
```
Simek, line 30:
```suggestion
    "pwa:generate": "npx pwa-asset-generator ./static/img/header_logo.svg ./static/img/pwa --padding '12.5%' --background #20232a --icon-only -opaque true --maskable true --type png"
```

From the command help, it looks like the padding parameter value need to wrapped with quotes. 

This might be the issue which cause malformed command output.

HyunseungLee-Travis:
I think this looks good.

maskable images

Simek (approved):
Thank you for all the work, LGTM! 👌

PR #3143, opened 2022-06-01
fix Prism CodeBlock appearance after the update

Simek:
# Why

After the latest Docusuaurs update it look like the theme variable overwrites the background color set in the custom Prism theme:
* https://github.com/facebook/react-native-website/blob/main/website/core/PrismTheme.js#L4

This results in display issues, especially in light mode:
Screenshot 2022-06-01 203429

# How

Overwrite the CodeBlock container background color, loose the class name selectors, remove the not working dark mode style.

I have re-enabled the `--deepdark` tweak, but it leads to the situation where Header and CodeBlocks have the same background color, so let's avoid that.

# Preview

Screenshot 2022-06-01 205848

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -686,7 +686,7 @@ So we need to "revert" some CSS vars to not enforce dark mode
 
     .menu__list-item a {
       color: var(--ifm-color-emphasis-700);
-      padding: 5px 12px;
+      padding: 4px 12px;
     }
 
     .menu__list-item-collapsible a[class*="menuLinkText"] {
@@ -847,7 +847,7 @@ aside[class^="theme-doc-sidebar-container"] {
   .menu__link--sublist {
     font-size: 15px;
     font-weight: 600;
-    padding: 5px 12px !important;
+    padding: 4px 12px !important;
     color: var(--light);
 
     &:after {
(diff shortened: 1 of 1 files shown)


Simek:
We were using this custom theme for a few years, and every Docusaurus release until latest update was working fine with the `background` set in the Prism theme. However I will investigate this a bit deeper later today.

Refs: https://github.com/facebook/docusaurus/pull/7172

Simek:
* https://github.com/FormidableLabs/prism-react-renderer/blob/5a4a1a0b7b68fa7c029bbe5e94527c53b2c6daac/tools/themeFromVsCode/src/transformSettings.js

slorber:
@Simek this file seems related to VS Code tooling 

Also this theme has a `background`, but not sure it's useful in practice and there's also `backgroundColor`: https://github.com/FormidableLabs/prism-react-renderer/blob/5a4a1a0b7b68fa7c029bbe5e94527c53b2c6daac/tools/themeFromVsCode/src/transformSettings.js

---

I'd rather ask the maintainer: @jpdriver should we support in Docusaurus a fallback so that a Prism theme with `plain.background` maps to `plain.backgroundColor`?

jpdriver:
@slorber unfortunately i'd say yes -- go with implementing a fallback for now.

we're hoping to overhaul styling for `prism-react-renderer` in a more significant way (adopting CSS variables) soon -- so hopefully this won't always be needed! 🙇🏻‍♂️

PR #3157, opened 2022-06-12
Remove backtick from readme

HyunseungLee-Travis:
'open http://localhost:3000' was surrounded with backtick which presented itself as a code and not a link.

# Before
`open http://localhost:3000`

# After
open http://localhost:3000

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If you are looking for the source code of the [React Native Archive website](htt
 
 1.  `cd website` to go into the website portion of the project.
 1.  `yarn start` to start the development server _(powered by [Docusaurus](https://v2.docusaurus.io))_.
-1.  Open http://localhost:3000/ to open the site in your favorite browser.
+1.  Open http://localhost:3000/ site in your favorite browser.
 
 ## 📖 Overview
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ If you are looking for the source code of the [React Native Archive website](htt
 
 1.  `cd website` to go into the website portion of the project.
 1.  `yarn start` to start the development server _(powered by [Docusaurus](https://v2.docusaurus.io))_.
-1.  `open http://localhost:3000/` to open the site in your favorite browser.
+1.  Open http://localhost:3000/ to open the site in your favorite browser.
```
Simek, line 34:
```suggestion
1.  Open http://localhost:3000/ site in your favorite browser.
```

Small simplification.

Simek (changes requested):
Hello @HyunseungLee-Travis, nice catch!

This line is referring to the `open` command on macOS:
* https://scriptingosx.com/2017/02/the-macos-open-command/

However, since the website can be run on different platforms now, it make sense to alter this fragment.

We can also recommend there the [NPM open CLI package](https://github.com/sindresorhus/open-cli) instead, but I think that just letting users know that they need to open the mentioned address in their favorite browser is fine.

I have just left the small suggestion, but besides that it LGTM! 👍

PR #3170, opened 2022-06-21
replace old blockquotes with admonitions, add codeblocks titles (part 1)

Simek:
# Why

This PR is the first in series, in which I going to replace legacy styling and workarounds with new Docusuaurs features (I have decided to split the changes into few PR so the amount of changes won't be overwhelming for the reviewers).

# How

The first change is to replace the the old `blockquote` blocks which were used for emphasising the content before the [Admonitions](https://docusaurus.io/docs/markdown-features/admonitions) was a thing. I have chosen the styling and added a custom tile (if needed) based on the current "notes" content.

Ideally, we want to get rid of all `blockquote` like that, to be able to use the component for the regular quotes, but this might take a longer while, so ATM I have put my focus on the unversioned docs.

The second change is that CodeBlock now supports `title` parameter, and in this PR I have replaced names in comments by this parameter for the affected CodeBlocks.

# Test plan

The changes have been tested by running the docs website locally.

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -6,7 +6,9 @@ description: Create mobile apps accessible to assistive technology with React Na
 
 Both Android and iOS provide APIs for integrating apps with assistive technologies like the bundled screen readers VoiceOver (iOS) and TalkBack (Android). React Native has complementary APIs that let your app accommodate all users.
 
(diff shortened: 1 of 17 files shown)


cortinico (approved):
LGTM 👍

PR #3175, opened 2022-06-23
added a note in running your app on real device section

wael-fadlallah:
While I was trying to run an app on my tablet I faced an issue that the device was listed as unauthorized, and while it was the only device listed on the list when I run `npm run android` the emulator was opened
I did a google search and end up with this fix

### DIFF
--- a/docs/running-on-device.md
+++ b/docs/running-on-device.md
@@ -43,7 +43,9 @@ emulator-5554 offline   # Google emulator
 
 Seeing `device` in the right column means the device is connected. You must have **only one device connected** at a time.
 
-> If you see `unauthorized` in the list you will need to run `adb reverse tcp:8081 tcp:8081` and press allow USB debugging on the device.
+:::note
+If you see `unauthorized` in the list you will need to run `adb reverse tcp:8081 tcp:8081` and press allow USB debugging on the device.
+:::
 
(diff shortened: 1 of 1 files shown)


docs/running-on-device.md
```diff
@@ -43,6 +43,8 @@ emulator-5554 offline   # Google emulator
 
 Seeing `device` in the right column means the device is connected. You must have **only one device connected** at a time.
 
+> If you see `unauthorized` in the list you will need to run `adb reverse tcp:8081 tcp:8081` and press allow USB debugging on the device.
```
Simek, line 46:
```suggestion
:::note
If you see `unauthorized` in the list you will need to run `adb reverse tcp:8081 tcp:8081` and press allow USB debugging on the device.
:::
```

Let's use the [Admonition](https://docusaurus.io/docs/markdown-features/admonitions) instead of the `blockquote` - we are in process of replacing note like those (they are legacy).

Simek (changes requested):
Hello @wael-fadlallah, thank you for adding this hint! 👍 

I have left one suggestion regarding note styling, besides that it LGTM.

Simek (approved):
LGTM, thank you for the docs update! 👍

PR #3177, opened 2022-06-24
Modify the top navbar to be less full

kelset:
This is a follow-up to a conversation that happened in the RN Discord; I saw the [Flutter website](https://flutter.dev/) having a clean and organised navbar and I wondered if that could have been replicated for the RN website.

Luckily, as pointed out by @slorber, Docusaurus supports this dropdown behaviour directly: https://docusaurus.io/docs/next/api/themes/configuration#navbar-dropdown

so the changes are fairly minimal, from a code perspective.

What I ended up doing for this first iteration is to group under "development" the 4 elements (in alphabetical order) "Architecture", "API", "Components", "Guides" and then re-add to the navbar the "Showcase" element.

Here are a few screenshots to show the result:
Screenshot 2022-07-08 at 14 53 14

Screenshot 2022-07-08 at 14 57 01

(but also ofc verify via the deployed preview)

---

A potential "second iteration" of this idea might make a second "dropdown" that could be something like "Community" -> "Get Help", "Events", "Contributing" or something along those lines... but first the get-help page needs a big update (it really needs one) and the confs stuff would have to be split out from that. Again, I'm just thinking out loud atm.

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -886,8 +886,7 @@ aside[class^="theme-doc-sidebar-container"] {
   }
 
   .menu__link--sublist {
-    font-size: 15px;
-    font-weight: 600;
+    font-size: 17px;
     padding: 4px 12px !important;
     color: var(--light);
 
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -596,10 +596,10 @@ a[class*="tagRegular"] {
       user-select: none;
     }
 
-    &.dropdown {
+    &.dropdown:not(.dropdown--right) {
       a {
         font-weight: 400;
-        font-size: 17px;
+        font-size: 14px;
       }
       ul {
         /* Control navbar dropdown alignment */
@@ -607,6 +607,18 @@ a[class*="tagRegular"] {
         left: 0;
       }
     }
+
+    &.dropdown--right {
+      .dropdown__menu {
+        min-width: 168px;
+      }
+    }
+  }
+
+  .dropdown > .navbar__link:after {
+    margin-left: 8px;
+    top: 1px;
+    opacity: 0.75;
   }
 
   .navbar__logo {
@@ -886,7 +898,7 @@ aside[class^="theme-doc-sidebar-container"] {
   }
 
   .menu__link--sublist {
-    font-size: 17px;
+    font-size: 15px;
     padding: 4px 12px !important;
     color: var(--light);
 
@@ -894,6 +906,10 @@ aside[class^="theme-doc-sidebar-container"] {
       background-size: 1.66rem 1.66rem;
       margin-right: -6px;
     }
+  }
+
+  .menu__link--sublist-caret {
+    font-size: 17px;
   }
 
   .menu__list {
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -599,7 +599,7 @@ a[class*="tagRegular"] {
     &.dropdown {
       a {
         font-weight: 400;
-        font-size: 14px;
+        font-size: 17px;
       }
       ul {
         /* Control navbar dropdown alignment */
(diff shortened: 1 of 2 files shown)


kelset:
or did you mean the bottom blue border for when you are in any of the "developer" areas? I just realized that border seems to be there only when you are on that area/page

kelset:
update: I've updated the PR to re-order the elements in the Development section has @Simek suggested, and updated a few CSS lines so that the menu in both small and big navbar look consistent.

From my perspective I think this can go in as is - and I'd like to then follow up on the idea that was tossed around a bit about doing a second one for Community section based on a few conversations we've been having.
But that would be a follow up PR so that it'd benefit from the CSS changes.

website/src/css/customTheme.scss
```diff
@@ -591,7 +591,7 @@ a[class*="tagRegular"] {
     &.dropdown {
       a {
         font-weight: 400;
-        font-size: 14px;
+        font-size: 17px;
```
Simek, line 594: This changes the version dropdown `font-size` too, it's better to adjust the rule to match only version dropdown, or add an overwrite for the menu dropdowns only.

Also 17px seems a little too big for the dropdown entries, I suggest to go with 1rem, since the default size (.875rem) seems definitely bit small.

website/src/css/customTheme.scss
```diff
@@ -878,8 +878,7 @@ aside[class^="theme-doc-sidebar-container"] {
   }
 
   .menu__link--sublist {
-    font-size: 15px;
-    font-weight: 600;
+    font-size: 17px;
```
Simek, line 881: Same story here, this alters the `font-size` of certain sidebar elements, not only for mobile unfortunately.

To ensure that you only target the elements that you want, you probably need to put this change inside correct `@media` query.

Simek (commented):
Thank you for the updates Kelset! 🙏 

I have left few suggestion regarding CSS changes, minor things, but unfortunately we need to be careful about custom styles with Docusaurus. 😅 

Whole idea of condensing menu and dropdown order seems fine for me, however I'm not a fan of exposing Showcase in the sidebar (yet, until we give this page a little bit more care and attention), but if others think otherwise we can ship the changes with this entry. 🙂

lunaleaps:
Couple of feedback items
1.  Right now there is a mismatch of the title under "Development" and the title on the left-hand bar. [Mismatch of "Guides" and "The Basics"](https://user-images.githubusercontent.com/1309636/187265963-476f3469-757d-4db6-a3de-0357cca49afc.png), [APIs vs API](https://user-images.githubusercontent.com/1309636/187266131-6fae7baa-72cb-4a4f-be87-1b9cfc40d89a.png)
2. In hamburger menu, when you're on one of the "Development" sub-pages, we don't auto un-collapse "Development" to show that you're under there. See in video how I had to click on "Development" title to see that I was under "Basics"?

https://user-images.githubusercontent.com/1309636/187266357-d941cca3-91dc-4bf6-8a38-66a82002feea.mp4

lunaleaps (changes requested):
See feedback

kelset:
about #2, I tried looking into the [docusaurus config](https://docusaurus.io/docs/api/themes/configuration/#navbar-dropdown) but couldn't find an option to manage the uncollapse. @slorber, @Simek do you have any ideas of how we could handle this? 🤔 I fear that it might be a limitation of Docusaurus in handling 2 level deeps navigation (development->guides->the basics)

That said, since it seems to me that it'd be a minor optimization for small screens I wouldn't consider it a blocker to get this PR merged.

kelset:
(btw, I'd like to get this https://github.com/facebook/react-native-website/pull/3328 merged first)

kelset:
now that #3328 is merged I think this one is ready to do - @lunaleaps @Simek what do you think? (see my comments above about the requested changes)

lunaleaps:
Looks good, I'll merge

PR #3178, opened 2022-06-24
docs(jdk): recommend JDK 11. JDK18 will not work. JDK17 may not work < 0.68

mikehardy:
I get reports of build errors from users that have chosen JDK18, that JDK does not work yet and should not be included either explicitly or even as part of an acceptable range ("...or newer")

See https://github.com/react-native-device-info/react-native-device-info/issues/1431#issuecomment-1165443830

I will add a second commit here in a moment for linux setup

### DIFF
--- a/docs/_getting-started-linux-android.md
+++ b/docs/_getting-started-linux-android.md
@@ -10,7 +10,7 @@ Follow the [installation instructions for your Linux distribution](https://nodej
 
 <h3>Java Development Kit</h3>
 
-React Native requires at least the version 8 of the Java SE Development Kit (JDK). You may download and install [OpenJDK](http://openjdk.java.net) from [AdoptOpenJDK](https://adoptopenjdk.net/) or your system packager. You may also [Download and install Oracle JDK 14](https://www.oracle.com/java/technologies/javase-jdk14-downloads.html) if desired.
+React Native currently recommends version 11 of the Java SE Development Kit (JDK). You may encounter problems using higher JDK versions. You may download and install [OpenJDK](http://openjdk.java.net) from [AdoptOpenJDK](https://adoptopenjdk.net/) or your system packager.
 
 <h3>Android development environment</h3>
 
--- a/docs/_getting-started-macos-android.md
+++ b/docs/_getting-started-macos-android.md
@@ -28,7 +28,7 @@ brew install --cask zulu11
 
 The Zulu OpenJDK distribution offers JDKs for **both Intel and M1 Macs**. This will make sure your builds are faster on M1 Macs compared to using an Intel-based JDK.
 
-If you have already installed JDK on your system, make sure it is JDK 11 or newer.
+If you have already installed JDK on your system, we recommend JDK 11. You may encounter problems using higher JDK versions.
 
 <h3>Android development environment</h3>
 
--- a/docs/_getting-started-windows-android.md
+++ b/docs/_getting-started-windows-android.md
@@ -18,7 +18,7 @@ Open an Administrator Command Prompt (right click Command Prompt and select "Run
 choco install -y nodejs-lts openjdk11
 ```
 
-If you have already installed Node on your system, make sure it is Node 14 or newer. If you already have a JDK on your system, make sure it is version 11 or newer.
+If you have already installed Node on your system, make sure it is Node 14 or newer. If you already have a JDK on your system, we recommend JDK11. You may encounter problems using higher JDK versions.
 
 > You can find additional installation options on [Node's Downloads page](https://nodejs.org/en/download/).
 


mikehardy:
I chose 14 as the max because I thought it was a well-maintained release but as I consult the version table it appears that 17 is an LTS, and currently 17 is supported. I'll modify it to recommend 11 or 17 as those are LTS versions so should receive updates through system package managers and thus be reasonable choices

https://en.wikipedia.org/wiki/Java_version_history

leotm (commented):
thx for picking this up, going w my spidey tingles in future to more proactively suggest/pr updates

cortinico:
I think we should stringly enforce Java 11 for now. I haven't got the time to look into this specifically, but I believe JDK 17 won't work at all (for the same reason JDK 18 won't work)

mikehardy:
I have specific user feedback that JDK17 is working...at least mostly? - 

https://github.com/react-native-device-info/react-native-device-info/issues/1431#issuecomment-1165708413

> very good and best !- yes all but some projets were working in JDK17

Let me check

mikehardy:
My react-native-firebase / react-native build qualifier has no trouble with JDK17 :shrug: - https://github.com/mikehardy/rnfbdemo/blob/main/make-demo.sh - used `brew install --cask temurin-17` and verified PATH etc. Built it debug and release mode, hermes enabled etc.

It is always possible I made a mistake but it appears to work fine, it is an LTS, and another user confirmed it worked as well.

So TL;DR: @cortinico I think JDK17 is good actually, an unexpected result on my part to be honest but happy about it since it's now the most current LTS.

cortinico:
> Built it debug and release mode, hermes enabled etc.

Have you tried to run it also?

mikehardy:
@cortinico 

> Have you tried to run it also?

Of course! :-) I also just ran my main work project (which has a large pile of dependencies, not all current) and it seemed fine. JDK17 seems fine, with success reports from template and in work projects and no failure reports. Absence of proof (of failure) is not proof of absence (of failure) but with zero reports of failure and success in all the things I normally test I'm not sure how else to qualify it, it seems good to me

leotm:
my results btw from earlier

Capture

cortinico:
> my results btw from [earlier](https://github.com/leotm/react-native-template-new-architecture/issues/709)

Thanks @leotm This clarifies a lot. I'm wondering if before we update this doc, I'd like to investigate what is the underlying issue with JDK 18 so we can just recommend to use any JDK >= 11

mikehardy:
Having an open upper end is always dangerous IMHO
Here's the underlying JDK18 issue (it was JDK16 and I guess 17 too, but now it's JDK18?) https://issuetracker.google.com/issues/186806275

mikehardy:
Feedback from previous user that prompted this PR for me is that JDK17 will work with RN68ish but rn67 is problematic for JDK17. 

*if* (big if? is it?) the docs are for new installs, then that tells me current state of PR (which recommends 11-17) is okay. If we want the docs to also be backwards compatible then it should be "use 11"

I continue to assert that open-ended upper range is dangerous as current android studio tooling is not compatible with JDK18

https://github.com/react-native-device-info/react-native-device-info/issues/1431#issuecomment-1166412075

That sounds all concrete and like an unmove-able stance but really I'm just trying to work with the data I've got and be helpful. Hopefully that comes through. Cheers

cortinico:
> Having an open upper end is always dangerous IMHO

Agree. Yet having to test the whole range is also a pain.
Also JDK 18 is soon to be superseeded by JDK 19 (18 was a short term support version).
I understand that users wants to update every tool they install, but we should probably pay more attention to the `doctor` command and make sure the required tools are aligned.

> Feedback from previous user that prompted this PR for me is that JDK17 will work with RN68ish but rn67 is problematic for JDK17.

Why was it problematic for RN67?

mikehardy:
> Why was it problematic for RN67?

Unsure exactly, it was logged in the google issue tracker which smells (warning: assumption ahead) like a gradle version compatibility thing between gradle / android gradle plugin. Perhaps it couldn't handle the new JDK17 ASTs or something I dunno

> I understand that users wants to update every tool they install, but we should probably pay more attention to the doctor command and make sure the required tools are aligned.

+1 on that! You know I've got a stack of doctor enhancements I'd like to work through but then firebase-ios-sdk threw the `use_frameworks!` bomb at my available time :sweat_smile: . I did think the docs update would help as something quick though.

So Perhaps just say "JDK11" and stick with it, sometimes being opinionated and simple is better than complicated. I don't feel strongly but I like fewer issues in my repo. If saying "JDK11" and done works here, that works for me

mikehardy:
Wow, apparently using JDK18 can result in ABI breaks in distributed apps on Android 8/8.1 - https://github.com/invertase/react-native-firebase/issues/6338 - so, not sure what to do about that but I *definitely* would not include JDK18 in any range we recommend at the moment

mikehardy:
That caused me to come look at this again - @cortinico I see you :heart:ed my comment about just saying "JDK11" so I altered all suggestions to just say 11 and warn about incompatibilities in higher versions with more specifics in the commit message

I squashed all the commits to a single one and rebased against current main here

should be good?

cortinico (approved):

cortinico:
> should be good?

Yes that's good. We can merge this.
FYI @mikehardy @leotm, I've opened this for the sake of visiblity:
- https://github.com/facebook/react-native/issues/34103

PR #3181, opened 2022-06-27
Update release steps

fortmarek:
When doing the release of `0.69.0`, we have noticed the guide for releasing a stable minor had a few inconsistencies.

### DIFF
--- a/website/contributing/release-stable-minor.md
+++ b/website/contributing/release-stable-minor.md
@@ -13,9 +13,7 @@ title: Release Stable Minor
 
 ```bash
 # In your react-native checkout, on the release branch of the version
-./scripts/bump-oss-version.js
-> What version are you releasing?
-# Your version
+./scripts/bump-oss-version.js -v x.y.z -t <YOUR_CIRCLE_CI_TOKEN>
 > Do you want this to be latest?
 # Generally yes. This updates npm registry to point to this version as "latest"
 ```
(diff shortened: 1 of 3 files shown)


website/contributing/release-stable-minor.md
```diff
@@ -29,14 +27,26 @@ title: Release Stable Minor
 
 0.66 stable is out!
```
Simek, line 28: Probably we should change the version in the whole article (currently we have the mix content mentioning 0.66 and 0.69) or just write the sentences in version-agnostic way (X.Y.Z).

fortmarek: I think it might be better to use a specific version, let me change it

kelset: maybe just use a `0.M.P` type of thing?

kelset: tbh it's fine either way 👍

kelset (commented):
LGTM 👍

website/contributing/release-stable-minor.md
fortmarek: If we want to use `0.M.P`, the better thing to do here would be to change it everywhere in the release steps. I don't mind doing it but maybe should be a different PR?

Simek (approved):
Thank you for the updates, LGTM! 👍

PR #3210, opened 2022-07-14
Correct minor mistake in native-components-android.md

KartonM:
I've fixed a small error in the `native-components-android.md` page

> ... "default" ~~can~~ value can be specified ...


Simek (approved):
Hello @KartonM, nice catch, thank you for the correction! 👍

PR #3218, opened 2022-07-20
update Docusaurus and other dependencies

Simek:
# Why

Docusaurus RC.1 has been released a while ago, let's update the website to use it.

# How

This PR updates the Docusaurus packages and few other dependencies used for website or in the workspace.

I did not spotted any regressions in styles/UI, I have also checked the changelog, and it seems like none of the few breaking changes do not affect the website.

ESLint checks also passed after the updates without any additional changes.


website/blog/2021-08-26-many-platform-vision.md
```diff
@@ -47,7 +47,7 @@ As another example, our research into user perception of speed on the web led to
 
 In addition to domain-specific engineers and meetups and conferences, each platform also brings other unique players solving similar problems. On the web, React (which directly powers React Native) frequently draws inspiration from other open source web frameworks like [Vue](https://vuejs.org/), [Preact](https://preactjs.com/), and [Svelte](https://svelte.dev/). On mobile, React Native has been inspired by other open source mobile frameworks, and we have been learning from other mobile frameworks built inside Facebook.
 
-<!-- alex ignore challenged -->
```
lunaleaps, line 50: Where does alex define these mappings i.e (special -> "challenged")?

lunaleaps (approved):
This looks good, we might need to fix the lock file conflicts.

website/blog/2021-08-26-many-platform-vision.md
Simek: Not sure TBH 😅 I have used the error messages to correct that. 

They include the correct keyword at the end of the error messages.

Simek:
Thank you for the review Luna! 🙏 Will resolve the lock conflict tomorrow. 

And since the stable Docusaurus version has already been released, I will follow this PR with another Docusaurus bump. Or maybe I should do the bump in this PR, wdyt?

lunaleaps:
Hmm.. maybe let's do it in a separate PR since we have all these lint clean-ups here too

PR #3234, opened 2022-07-29
[Documentation] add running your app on a daily basis section

charlotteisambert:
Add a section in the getting started documentation on how to launch an app without going through `react-native run-ios` every time.

I learned after one year of react-native development that I didn't actually need this command to open my app and start Metro. I hope this section can save time to other react-native beginners.

Capture d’écran 2022-07-29 à 18 11 20


3imed-jaberi (commented):
sgtm +1

PR #3239, opened 2022-08-03
Remove release supporter roles

lunaleaps:
Remove release supporter roles as discussed in release retrospective for 0.69


lunaleaps:
Addressed in https://github.com/facebook/react-native-website/pull/3240

PR #3242, opened 2022-08-04
Add retrospective as a copilot responsibility

fortmarek:
Copilot should run a retrospective after a new minor is released (at least in cases when there's enough to talk about) cc @kelset


kelset:
sounds good 👍

cortinico (approved):

PR #3243, opened 2022-08-04
Fixed pod install instructions

scarlac:
Yarn must be run prior to pod install, otherwise pod install fails with "Error: Cannot find module 'mkdirp'"

### DIFF
--- a/website/contributing/how-to-run-and-write-tests.md
+++ b/website/contributing/how-to-run-and-write-tests.md
@@ -25,11 +25,16 @@ yarn run lint
 
 ### iOS Tests
 
-Start off by running `pod install` inside the `RNTester` directory. This will set up your native dependencies and create a `RNTesterPods` Xcode workspace.
+Start off by going to the `packages/rn-tester` directory, then:
+1. Run `yarn`
+2. Run `pod install` (You may have to run `brew install cmake` prior)
+
+This will set up your native dependencies and create a `RNTesterPods` Xcode workspace. 
+**Note:** It may take a long time to complete "Installing hermes-engine (1000.0.0)" as it must build Hermes from source.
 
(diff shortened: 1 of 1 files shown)


website/contributing/how-to-run-and-write-tests.md
```diff
@@ -25,9 +25,14 @@ yarn run lint
...
-Start off by running `pod install` inside the `RNTester` directory. This will set up your native dependencies and create a `RNTesterPods` Xcode workspace.
+Start off by going to the `packages/rn-tester` directory, then:
+1. Run `yarn`
+2. Run `pod install` (You may have to run `brew install cmake` prior)
 
-Then, go back to the root of your React Native checkout and run `yarn` followed by `yarn start`. This will set up your JavaScript dependencies.
+This will set up your native dependencies and create a `RNTesterPods` Xcode workspace. 
+**Note:** It may take a long time to complete "Installing hermes-engine (1000.0.0)" as it must build Hermes from source.
+
+Then, go back to the root of your React Native checkout and run `yarn`. This will set up your JavaScript dependencies.
```
cipolleschi, line 35: RN-Tester has its own [README](https://github.com/facebook/react-native/tree/main/packages/rn-tester) about how to run it: it works and it is more precise.
What about linking that README here, instead of writing a doc that we have to keep in sync, and improve that README instead?

cipolleschi (changes requested):
Thanks for doing this. I left a suggestion to lighten the maintenance burden. Let me know what do you think.

website/contributing/how-to-run-and-write-tests.md
scarlac: Great idea! I didn't know. I'll update it.

3imed-jaberi (commented):
+1

cipolleschi (approved):

PR #3248, opened 2022-08-08
Update to Docusaurus stable release, fix alert admonition styling

Simek:
# Why & How

This PR updates the Docusuaurs to the latest and stable release! 🎉

There is no changelog published yet from latest RC, but there were almost no changes between those versions.

Testing the bump locally I was not bale to spot any visual regressions.

I have also utilize this bump to address the warning admonitions style, which since bump to RC started to match again against one of our legacy CSS selectors, which lead to unexpected effect (text centered, small font), those issue have been fixes as well as double underline for links in admonitions.

# Preview

Screenshot 2022-08-08 140012

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -506,7 +506,7 @@ html[data-theme="dark"].community-page {
   }
 
   .margin-top--md {
-    margin-top: 8px !important;
+    margin-top: 0.33rem !important;
   }
 }
 
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -22,7 +22,7 @@
   --ifm-code-font-size: 90%;
   --ifm-code-border-radius: 3px;
   --ifm-blockquote-color: var(--ifm-font-color-base);
-  --ifm-blockquote-font-size: 15px;
+  --ifm-blockquote-font-size: 16px;
   --ifm-table-head-color: var(--subtle);
   --ifm-link-hover-decoration: none;
   --ifm-navbar-background-color: var(--deepdark);
@@ -405,13 +405,13 @@ html[data-theme="dark"] article .badge {
   margin-bottom: 1.5rem !important;
 
   div[class^="admonitionContent"] {
-    h5 {
-      font-size: 14px;
-    }
-
     a {
       border-bottom: 0;
     }
+  }
+
+  div[class^="admonitionHeading"] {
+    font-size: 14px;
   }
 }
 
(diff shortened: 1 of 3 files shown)


cortinico (dismissed):

cortinico:
I've accidentally approved this one without checking the live preview. I'm handing over the @huntie for a full-review

website/src/css/customTheme.scss
```diff
@@ -399,24 +399,24 @@ html[data-theme="dark"] article .badge {
...
-  background-color: var(--rn-note-background);
-  font-size: 14px;
-  text-align: center;
+.admonition,
+.theme-admonition {
+  font-size: var(--ifm-blockquote-font-size);
+  margin-bottom: 1.5rem !important;
 
-  .margin-top--md {
-    margin-top: 0.33rem !important;
```
huntie, line 408: These styles have been correctly reset for the general case (bug introduced in https://github.com/facebook/react-native-website/pull/2263/commits/7a22db1cb93d66f7033469b3f19e60fd3e581d91), but we still want to preserve this for `.theme-doc-version-banner` — please can we move these rules into the below selector (line 493)? :)

image

website/src/css/customTheme.scss
```diff
@@ -399,24 +399,24 @@ html[data-theme="dark"] article .badge {
...
+  div[class^="admonitionContent"] {
+    h5 {
+      font-size: 14px;
+    }
```
huntie, line 410: As far as I can tell, this `h5` rule is unused. Let's remove it :)

website/src/css/customTheme.scss
```diff
@@ -22,7 +22,7 @@
   --ifm-code-font-size: 90%;
   --ifm-code-border-radius: 3px;
   --ifm-blockquote-color: var(--ifm-font-color-base);
-  --ifm-blockquote-font-size: 16px;
+  --ifm-blockquote-font-size: 15px;
```
huntie, line 25: This variable is potentially misnamed, as it is only used by `.admonition`. Since the default font size for page content is `17px`, I'm not sure there's much justification to reduce this value.

**Optional**: I would be in favour of removing this line _and_ removing our `font-size` and `margin-bottom` overrides to the `.admonition` selector — falling back to the Docusaurus theme defaults (which are only 1px different).

Simek:
Thank you for the review and catching those issues! 🙏 Will address therm soon.

Generally our styles are messy because they came with us a long way, and they dates back to v1. Ideally we should rewrite them all to be honest at same point. 🙂 

Those top banners are something legacy, that we additionally customized before the Swizzling was an option. We probably should just replace them with admonitions or custom components.

website/src/css/customTheme.scss
Simek: Good catch, they changed the `h5` to `div` with `admonitionHeader` in the class name, this should target the admonition title.

website/src/css/customTheme.scss
Simek: We can remove the font size overwrite, but the margin needs to stay, default `1em` which came from the internal component styles is a bit too little spacing.

huntie (approved):
Thanks for the changes!

PR #3261, opened 2022-08-15
nit: fix typo in new architecture docs

intergalacticspacehighway:
## Issue
Fixes tiny typos in the new architecture docs

### DIFF
--- a/docs/the-new-architecture/pillars-fabric-components.md
+++ b/docs/the-new-architecture/pillars-fabric-components.md
@@ -96,7 +96,6 @@ import type { HostComponent } from 'react-native';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
 export interface NativeProps extends ViewProps {
-  ...ViewProps,
   text: string | null | undefined,
   // add other props here
 }
--- a/docs/the-new-architecture/pillars-fabric-components.md
+++ b/docs/the-new-architecture/pillars-fabric-components.md
@@ -96,7 +96,7 @@ import type { HostComponent } from 'react-native';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
 export interface NativeProps extends ViewProps {
-  text: string | null | undefined,
+  text: string | null | undefined;
   // add other props here
 }
 
--- a/docs/the-new-architecture/pillars-fabric-components.md
+++ b/docs/the-new-architecture/pillars-fabric-components.md
@@ -96,7 +96,7 @@ import type { HostComponent } from 'react-native';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
 export interface NativeProps extends ViewProps {
-  text: string | null | undefined;
+  text?: string;
   // add other props here
 }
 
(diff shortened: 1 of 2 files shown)


docs/the-new-architecture/pillars-fabric-components.md
```diff
@@ -96,7 +96,6 @@ import type { HostComponent } from 'react-native';
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 
 export interface NativeProps extends ViewProps {
-  ...ViewProps,
```
intergalacticspacehighway, line 99: I think this is not valid in typescript, extending the interface with ViewProps should be enough I guess. Let me know if I am wrong though!

Simek: Agree, this also probably should be `type` not `interface` too, as well `text` should be marked as optional and then `undefined` can be removed from the type.

Let's see what's others opinion, CC @cipolleschi, @cortinico

intergalacticspacehighway: yeah, thought it might be required for codegen. I verified the codegen was working after removing the spread syntax.

intergalacticspacehighway: also, this felt a bit weird.  `import type {ViewProps} from 'ViewPropTypes';`
Thought it should be `import type {ViewProps} from 'react-native/Libraries/Components/View/ViewPropTypes';` instead but typescript package is not yet upgraded, so not sure.

cipolleschi: Hi, thanks for the comments!

I'm not a TypeScript expert, but I will do some checks.

Could you update the PR making the `NativeProps` a type and the `text` property optional?

cipolleschi (changes requested):

docs/the-new-architecture/pillars-fabric-components.md
intergalacticspacehighway: @cipolleschi thanks for checking. I could make those changes but I think current codegen requires component spec [to be an interface](https://github.com/facebook/react-native/blob/e7a4dbcefc9e393c41f4a796d522211bc1e60b6f/packages/react-native-codegen/src/parsers/typescript/components/index.js#L136).

I also get an error while performing codegen with below types

```ts
type NativeProps = Readonly<{
    text?: string
}> & ViewProps

```

Above is the flow equivalent in typescript but uses type instead of interface. I think @Titozzz who recently did PR in webview is [using interface ](https://github.com/react-native-webview/react-native-webview/pull/2635/files#diff-c039269b1e20f1eaf19cb406b643be7c06c97a34884a8d0e903510394a539d4bR129) due to same reason. 

cc @acoates-ms @ZihanChen-MSFT

docs/the-new-architecture/pillars-fabric-components.md
```diff
@@ -96,8 +96,7 @@ import type { HostComponent } from 'react-native';
...
-  text: string | null | undefined,
+  text?: string;
```
intergalacticspacehighway, line 99: @cipolleschi made the `text` property optional, verified that codegen works with it

cipolleschi (approved):

PR #3282, opened 2022-08-28
Move the "Using `forwardRef`" and "Using a `getter`" sections to a lower hierarchy level

cipolleschi:
The two section [Using `forwardRef`](https://reactnative.dev/docs/next/new-architecture-library-intro#using-forwardref) and [Using a `getter`](https://reactnative.dev/docs/next/new-architecture-library-intro#using-a-getter-note-the-addition-of-getviewref) are alternative solutions to get rid of `findNodeHandle`, based on the type of component (function or class). 
However, the current structure of the doc leads the reader to think that these are two other API that needs to be migrated: they are at the same level of the `findNodeHandle` one and they also appear in the ToC of the page, on the right.

This PR moves them down one level, reducing the importance of their title and it should make them disappear from the ToC on the right.


Simek (approved):
LGTM! 👍

PR #3284, opened 2022-08-29
Fix color reference in flexbox.md

EvertEt:


Simek (approved):

PR #3287, opened 2022-08-30
basic documentation for new roles and traits

fabriziobertoglio1987:
documentation for https://github.com/facebook/react-native/pull/34477


fabriziobertoglio1987:
- [ ] add disclaimer that role is only supported on android if no iOS [trait](https://developer.apple.com/documentation/uikit/uiaccessibilitytraits?language=objc) availalbe https://github.com/facebook/react-native-website/pull/3226/files#diff-69a54da073e7180ca0a427a159f89255379f695ac1456a924f24cf067d4efa20R164

PR #3300, opened 2022-09-06
Change form to from

tientaidev:
Change log:
- Fix typo `form` to `from`


Simek:
Hello @tientaidev, thank you for the correction, but It looks like this has been already addressed in #3301.

PR #3307, opened 2022-09-08
[Fix] Add Missing Line in Podfile

cipolleschi:
In the guide to migrate an App to the New Architecture, we were missing a line that has to be added in the Podfile by the user.
This PR adds that mising line.


cortinico (approved):

cipolleschi:
Will be superseeded by another PR

PR #3326, opened 2022-09-16
fix _getting-started-macos-ios.md info format

sunnylqm:


Simek (approved):

PR #3329, opened 2022-09-19
Fixing symbolic links

rplankenhorn:
Soft links require an absolute path to work properly.  Before this change, `which /usr/local/bin/gcc` still resolves to `/usr/bin/gcc` even with `/usr/local/bin` being earlier in the `$PATH` variable.


cortinico (approved):
Can you apply the changes also to the versioned docs?

Simek:
Thank you for the contribution @rplankenhorn! 👍 

@cortinico Going to merge this PR and prepare a follow-up for the versioned docs.

PR #3335, opened 2022-09-21
First version of React Native Version and Lifecycle support document

Andjeliko:
Blog post sharing the discussed and agreed version of the RN versions and lifecycle support document.
Also including upgrade to docosaurus 2.1.0


Simek:
Hello @Andjeliko, unfortunately upgrading Docusaurus is not that staring forward, you need to verify that various content and UI with custom styles is still rendering correctly after the bump.

At the quick glance at the deployment I can see a visual regression, like unstyled announce bar. I would recommend pushing an upgrade in the separate PR, so it won't block the blog post you want to ship.

Simek (changes requested):
Also, please remove the `package-lock` file from the changeset. 

(Yarn lock seems to have a lot of changes too, which is a bit unexpected by just bumping the Docusaurus, can you verify that all changes in there are expected?)

Andjeliko:
Good point @Simek reversed docosaurus change now.

Simek:
Thank you! Can you also revert the `yarn.lock` changes which left in the changeset? 

It looks like you have Yarn configured to use NPM registry instead of the default one:
Screenshot 2022-09-21 145404

cipolleschi (approved):

website/blog/2022-09-21-rn-versions-lifecycle-support.md
```diff
@@ -0,0 +1,83 @@
...
+Please note that each cherry-pick request will be assessed and approved individually. Cherry-Pick requests against unsupported versions will be rejected unless they involve security issues.
+
+## Security Issues
+
+Security issues should follow our issue report policy [described here](https://reactnative.dev/contributing/overview#security-bugs).
+Reports for security issues on unsupported versions will handed on case by case basis when deciding if a new version needs to be released.
+
+## Licence
+
+React Native is open-source software distributed under the **MIT license**. Software is provided "as is" without warranty.
```
kelset, line 83: I reckon you won't need this in the blogpost - maybe close it with a call to action to update to the latest versions of the package if folks haven't done it yet?

Simek (approved):
Thank you for the lock revert! 🙏 Approving to do not block this one.

website/blog/2022-09-21-rn-versions-lifecycle-support.md
rickhanlonii: +1

website/blog/2022-09-21-rn-versions-lifecycle-support.md
```diff
@@ -0,0 +1,83 @@
...
+  - The highest version number is defined following the [NPM semver algorithm](https://github.com/npm/node-semver) tagged as "latest".
+- **next version** - The next version that hasn’t been fully released yet, and is currently in development/testing (e.g. 0.69.0-RC1).
+- **minor series** - A collection of versions (either stable or not stable) sharing the same minor version (e.g. **0.68**.0-RC0, **0.68**.0, **0.68**.1 are all part of the **0.68** minor series).
+
+## Which versions are currently supported?
+
+We’re supporting the **latest version**, and the latest versions from the **two previous minor series.**
+We’ll also support the **next version** being developed, which will become stable after its release.
+
+Versions:
```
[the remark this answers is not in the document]
Andjeliko: Agreed moved blog to https://github.com/reactwg/react-native-releases/blob/main/README.md

cipolleschi:
@Andjeliko @cortinico @kelset @rickhanlonii What's the status of this?

kelset:
my understanding is that this could be closed since we've already propagated this information in many other places and ways (and it will be further mentioned in the 0.71 and maven incident post mortem posts)

cortinico:
> my understanding is that this could be closed since we've already propagated this information in many other places and ways (and it will be further mentioned in the 0.71 and maven incident post mortem posts)

I believe we might want to blogpost-it at some point anyway (cc @rickhanlonii @lunaleaps)

lunaleaps:
Have some questions on why this should be a blog post -- would this be better suited to an announcement on the Releases working group?

cortinico:
> Have some questions on why this should be a blog post -- would this be better suited to an announcement on the Releases working group?

A post in the release wg would also work 👍 It will also allow people to comment and follow up if needed

PR #3337, opened 2022-09-22
fix: remove isTurboModuleEnabled check from docs

okwasniewski:
Hello,

This PR removes `isTurboModuleEnabled` check from the docs, since it's no longer needed.

### DIFF
--- a/docs/the-new-architecture/backward-compatibility-turbomodules.md
+++ b/docs/the-new-architecture/backward-compatibility-turbomodules.md
@@ -347,7 +347,7 @@ For a Turbo Native Module, the source of truth is the `Native<MyModule>.js` (or 
 import MyModule from 'your-module/src/index';
 ```
 
-The **goal** is to conditionally `export` the proper object from the `index` file , given the architecture chosen by the user. We can achieve this with a code that looks like this:
+Since `TurboModuleRegistry.get` taps into the old `NativeModules` API under the hood, we need to re-export our module, to avoid registering it multiple times.
 
 <Tabs groupId="turbomodule-backward-compatibility"
       defaultValue={constants.defaultTurboModuleSpecLanguage}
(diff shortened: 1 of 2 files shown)


docs/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -347,7 +347,7 @@ For a Turbo Native Module, the source of truth is the `Native<MyModule>.js` (or
 import MyModule from 'your-module/src/index';
 ```
 
-The **goal** is to conditionally `export` the proper object from the `index` file , given the architecture chosen by the user. We can achieve this with a code that looks like this:
+Since `TurboModuleRegistry.get` taps into the old `NativeModules` API under the hood, we need to re-export our module, to avoid registering it multiple times.
```
cipolleschi, line 350:
```suggestion
Since `TurboModuleRegistry.get` taps into the old Native Modules API under the hood, we need to re-export our module, to avoid registering it multiple times.
```

website/versioned_docs/version-0.70/the-new-architecture/backward-compatibility-turbomodules.md
```diff
@@ -332,7 +332,7 @@ For a Turbo Native Module, the source of truth is the `Native<MyModule>.js` (or
 import MyModule from 'your-module/src/index';
 ```
 
-The **goal** is to conditionally `export` from the `index` file the proper object, given the architecture chosen by the user. We can achieve this with a code that looks like this:
+Since `TurboModuleRegistry.get` taps into the old `NativeModules` API under the hood, we need to re-export our module, to avoid registering it multiple times.
```
cipolleschi, line 335:
```suggestion
Since `TurboModuleRegistry.get` taps into the old Native Modules API under the hood, we need to re-export our module, to avoid registering it multiple times.
```

cipolleschi (changes requested):
Some small changes and then it will be good to go! ;)

PR #3338, opened 2022-09-23
rework the community section of the website

kelset:
Now that the work on the showcase has seen good progress (and that it's hackathon week at MSFT), I decided to go back to another white whale of mine on the website and take a stab at the community section.

This PR is fairly big, taking the existing single page, splitting it into multiple, updating and enhancing its content (used as inspiration[ my community map ](https://lifeiswhat.notion.site/The-React-Native-Community-Map-07f3221c294d4c8ab65a13cf1dee6901)) and adding it back to the header - while also tweaking the footer.

I think this will need to pass a few review rounds but overall I think this is a pretty good place to start a rework for the community section on the website :)

One thing that I deliberately not added to any of the pages, right now, are podcasts and newsletters - but I think that maybe they should be there, in the staying up to date section; let me know your thoughts.


website/community/staying-updated.md
```diff
@@ -0,0 +1,60 @@
...
+
+### News
+
+Twitter is one of the main communication channels used by the team to communicate updates, news and request for participation: follow the [React Native account](https://twitter.com/reactnative) to stay up to date - the [ReactJS account](https://twitter.com/reactjs) covers both React and React Native.
+
+For long form news and announcements, keep an eye on the [blog](blog) to find out what is happening in the world of React Native.
+
+### Discussions and efforts
+
+React Native is a dynamic framework, and its release cycle leaves the door open for discussing how it can evolve at every step of the way.
```
lunaleaps, line 15: Some potential simplification

```suggestion
React Native is a dynamic framework and there is always room for improvement. As an open source framework, large-scale coordination and discussion should remain publicly accessible. Here are some dedicated discussion groups you can participate in: 
```

lunaleaps:
Also there was some discussion about splitting this into separate sections. It might be nice to keep it still as one page for easier discovery cc @rickhanlonii

website/community/communities.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Communities
+description: The React Native Community
+wrapperClassName: 'community-communities-page'
```
Simek, line 4:
```suggestion
```

`wrapperClassName` was required to correct the page appearance, since it was a plain MD page earlier. 

Now, you are using the `content-doc` plugin for the community section pages, which adds the docs pages classes and makes this property redundant.

Simek: You can also remove `.community-page` references from styles:
https://github.com/facebook/react-native-website/blob/72c8bd5f749ec7d3256af651a7ea1ed3ebd5a9c1/website/src/css/customTheme.scss#L382-L386

kelset: thanks! done

website/community/overview.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Overview
+description: The React Native Community
+wrapperClassName: 'community-overview-page'
+---
+
+React Native is a widely used technology, empowered by a massive community of hundreds of thousands of developers.
```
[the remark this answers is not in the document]
kelset: I've added links to the stargazers page of the GH repo, and expanded the phrase to also cover #weekly downloads and linked to npm. LMK if that's closer to what you had in mind

kelset:
> It might be nice to keep it still as one page for easier discovery

I disagree, having smaller pages with focused topics would make it easier to read through and find what you are looking for

website/community/communities.md
Simek: Thank you! 🙏  One more nit, you have left `html[data-theme="dark"].community-page .hash-link` line in styles. 😉

lunaleaps:
@kelset I think the content of each section feels quite short. I can see how `cmd+f` on a single page versus having it split up being a pro. I don't feel super strongly though. 

In addition, should we add the devX discussion group? We could also link the general organizations in case there are new discussion groups created. Maybe we should follow some sort of nomenclature for discussion-only repos under react-native-community?

kelset:
> I can see how `cmd+f` on a single page versus having it split up being a pro.

Again, we have different views - IMHO if a person needs to do a cmd+f on a docs page it's already a loss for the page. (especially if there's a search feature on the website).

That said, I totally understand it's just my personal take so I prepared a second PR that has the same content changes but kept in the monopage style: https://github.com/facebook/react-native-website/pull/3350

Both are updated with the DX WG also linked in, just select whichever you find more correct and let's get it merged :)

> Maybe we should follow some sort of nomenclature for discussion-only repos under react-native-community?

Potentially, I don't think it'd change things much - let's follow up on that separately ;)

lunaleaps:
Wow thanks for doing that! I think I do prefer that as it's easier to scroll through on mobile. @Simek do you have a preference as a tie-breaker?

Simek:
Heh, it's a hard call to made. I have no strong opinion on one approach over other, but I think that introducing docs plugin (multiple pages) includes a future-proofing aspect, because there will be a point in time, when the single page will grow to the size when we will need to split it anyway.

Having those short pages for a while, might also be an incentive for updating and expanding them further for us or for someone from the community.

The mobile UX is not ideal, I agree with that too, but it's inherited from Docusaurus solutions/choices and we can hope that those will improve with time in next releases, or we can propose some change either.

lunaleaps (approved):
Looks good to me!

website/community/communities.md
```diff
@@ -0,0 +1,26 @@
...
+
+The React Native ecosystem is far and wide, and people can be part of it in many forms; here you will find but a partial list of different ways one developer can be part of it. If you know of other, or want to help expand this page, [submit a PR](https://github.com/facebook/react-native-website/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc)!
+
+### Local communities
+
+There are a lot of React Native gatherings that happen around the world. Often there is React Native content in React meetups as well, use tools like [Meetup](http://www.meetup.com/topics/react-native/) and [Eventbrite](https://www.eventbrite.co.uk/d/online/react-native/?page=1) to find out recent events in your area - or start one!
+
+### Company-based communities
+
+Some companies actively involved in the React Native have also their own communication channels focused towards the projects they maintain, like [Callstack.io's Discord server](https://discordapp.com/invite/zwR2Cdh), [Invertase.io's Discord server (e.g. React Native Firebase)](https://discord.gg/C9aK28N), [Infinite Red's Slack Group](http://community.infinite.red/) and [The Expo Discord server](https://chat.expo.dev/).
```
Simek, line 14: Just a readability nit, can we use list here?

kelset: good idea, done - thanks!

kelset:
@lunaleaps / @Simek I've updated both PRs with the company-based comminities list - again, pick one and let's merge it 🚀

Simek (approved):

PR #3359, opened 2022-10-04
Fix codegen iOS scripts paths

krystofwoldrich:
I've noticed a typo in the iOS codegen script paths.


krystofwoldrich:
Thanks. :)

PR #3365, opened 2022-10-07
fix: duplicate header

Naturalclar:
Deleted a duplicated header


Simek:

Simek (approved):

PR #3371, opened 2022-10-12
fix: add back missing docs vote component

Simek:
# Why

@arati-1 spotted that the docs page vote component has disappeared from the documentation, let's fix that.

# How

The problem seems to have its root cause in the change structure of the swizzled components in one of Docusaurus releases, so `DocItemFooter` become `DocItem/Footer`, which stopped to load customized component.

Also, I was not able to use "Wrap" mode again, since it look like wrapper no longer have an access to all children props, like need `content.metadata`, so I had to use full swizzle ("Eject") the component to bring back the voting mechanism.


lunaleaps (approved):
Whoa cool, had to look up what swizzle means -- love that term lol
Thank you @Simek so much for fixing!

Simek:
For the context:
* https://docusaurus.io/docs/next/swizzling

PR #3372, opened 2022-10-12
Update typescript.md

sunnylqm:


docs/typescript.md
```diff
@@ -138,16 +138,16 @@ Out of the box, transforming your files to JavaScript works via the same [Babel
 
 ## What does React Native + TypeScript look like
 
-You can provide an interface for a React Component's [Props](props) and [State](state) via `React.Component<Props, State>` which will provide type-checking and editor auto-completing when working with that component in JSX.
+You can provide an interface for a React Functional Component's [Props](props) via `React.FC<Props>` which will provide type-checking and editor auto-completing when working with that component in JSX.
```
Simek, line 141:
```suggestion
You can provide an interface for a React Function Component's [Props](props) via `React.FC<Props>` which will provide type-checking and editor auto-completing when working with that component in JSX.
```

It's "function" component, not "functional" - https://reactjs.org/docs/components-and-props.html.

docs/typescript.md
```diff
@@ -138,16 +138,16 @@ Out of the box, transforming your files to JavaScript works via the same [Babel
...
 
 ```tsx title="components/Hello.tsx"
 import React from 'react';
 import { Button, StyleSheet, Text, View } from 'react-native';
 
-export type Props = {
+export interface Props {
```
Simek, line 147:
```suggestion
export interface Props {
```

Can you elaborate on this change? I feel like `type` is the correct structure here, `interface` won't be optimized/removed and you cannot use `import type` with it.

sunnylqm: `interface` will be removed by compiler and can be imported for sure and is preferred to use for performance  https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

NickGerleman: That rule seems to be advocating for interface + extends over intersecting types. This example isn't intersecting types.

sunnylqm: Say you will change a type to interface whenever you want a intersecting type? Will that make sense? Compared to use an interface from the beginning?

PR #3381, opened 2022-10-26
Include window reference for iOS swift projects

jerryseigle:
Updated instructions to add React Native to existing iOS app using swift

- Added window reference which is to be included in the AppDelegate
- Updated Podfile instructions 

Correct documentation to include window reference when integrating into existing iOS swift project. If a window reference is not included in iOS swift project the app will crash with error Thread 1: "-[ProjectName.AppDelegate window]: unrecognized selector sent to instance 0x600001f9dd60"

### DIFF
--- a/docs/_integration-with-exisiting-apps-swift.md
+++ b/docs/_integration-with-exisiting-apps-swift.md
@@ -288,6 +288,28 @@ Wire up the new link in the main menu to the newly added event handler method.
 
 > One of the easier ways to do this is to open the view in the storyboard and right click on the new link. Select something such as the `Touch Up Inside` event, drag that to the storyboard and then select the created method from the list provided.
 
+##### 3. Window Reference
+
+Add an window reference to your AppDelegate.swift file. Ultimately, your AppDelegate should look something similar to this:
+
+```
+import UIKit
+
+@main
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+    // Add window reference
+    var window: UIWindow?
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        // Override point for customization after application launch.
+        return true
+    }
+
+    ....
+}
+```
+
 ### Test your integration
 
 You have now done all the basic steps to integrate React Native with your current application. Now we will start the [Metro bundler][metro] to build the `index.bundle` package and the server running on `localhost` to serve it.
--- a/docs/_integration-with-exisiting-apps-swift.md
+++ b/docs/_integration-with-exisiting-apps-swift.md
@@ -101,48 +101,48 @@ Ultimately, your `Podfile` should look something similar to this:
 
 ```
 require_relative '../node_modules/react-native/scripts/react_native_pods'
-require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+ require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 
-platform :ios, '12.4'
-install! 'cocoapods', :deterministic_uuids => false
+ platform :ios, '12.4'
+ install! 'cocoapods', :deterministic_uuids => false
 
-target 'MyReactNativeApp' do
-  config = use_native_modules!
+ target 'MyReactNativeApp' do
+   config = use_native_modules!
 
-  # Flags change depending on the env values.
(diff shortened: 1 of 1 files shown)


docs/_integration-with-exisiting-apps-swift.md
```diff
@@ -288,6 +302,28 @@ Wire up the new link in the main menu to the newly added event handler method.
 
 > One of the easier ways to do this is to open the view in the storyboard and right click on the new link. Select something such as the `Touch Up Inside` event, drag that to the storyboard and then select the created method from the list provided.
 
+##### 3. Window Reference
+
+Add an window reference to your AppDelegate.swift file. Ultimately, your AppDelegate should look something similar to this:
+
+```
```
cipolleschi, line 309: can you add the `swift` language tag here, so that the syntax results highlighted correctly?
Should be something like:
```
 ```swift
```

docs/_integration-with-exisiting-apps-swift.md
```diff
@@ -100,34 +100,48 @@ The `Podfile` will contain a boilerplate setup that you will tweak for your inte
...
-
-  # Third party deps podspec link
-  pod 'DoubleConversion', :podspec => '../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec'
-  pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
-  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
+require_relative '../node_modules/react-native/scripts/react_native_pods'
+require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
+
+platform :ios, '12.4'
+install! 'cocoapods', :deterministic_uuids => false
```
cipolleschi, line 107: We are updating this in the template. the new setup would be:
```ruby
platform :ios, min_ios_version_supported
prepare_react_native_project!
```

Perhaps it could make sense to link to the [most recent `Podfile`](https://github.com/facebook/react-native/blob/main/template/ios/Podfile) we have in the `template`, rather than manually copy and paste it?
So we don't have to come back to this if we make updates, what do you think?

cipolleschi (changes requested):
Thank you so much for taking care of this! :bow: 

I added a couple of suggestions to improve readability.

jerryseigle:
@cipolleschi is there a timeframe where these updates will reflect on the website

cortinico:
> @cipolleschi is there a timeframe where these updates will reflect on the website

They'll be visible as soon as we merge this PR? Does this answer your question?

cipolleschi:
@jerryseigle could you apply the suggestion I sent, please? We won't merge the PR without those. Thank you so much!

jerryseigle:
@cipolleschi I tried to accept changes but the option is disabled. It will only allow me to comment

Screenshot 2022-12-15 at 1 42 51 PM

cipolleschi:
Ah no, you have to make the changes locally on your branch and add another commit.

jerryseigle:
@cipolleschi I just added the recommendation changes you requested. Hope this is what you need. please let me know any additional change or task needed if any.

cipolleschi (changes requested):
Uhm, no. I can't see any changes to the files on github. :( 

The suggestion was to completely remove the code of the Podfile and to add a link to the Podfile in the template: https://github.com/facebook/react-native/blob/main/template/ios/Podfile

And below, just add the word `swift` after the triple ticks ` ``` `.

to apply the changes, you have to open the files in your local environment, apply the changes and then run `git add`, `git commit`, and `git push` with the right parameters so that the PR gets updated.

docs/_integration-with-exisiting-apps-swift.md
jerryseigle: included the link to the Podfile Template.

jerryseigle:
@cipolleschi just completed the update. Sorry took so long please let me know if this is what is expected

cipolleschi (approved):
Perfect, thank you for doing this!

cipolleschi:
There are some linting errors... :( 

Screenshot 2022-12-30 at 16 33 29

jerryseigle:
Do I need to fix linting issues or have they been corrected

cipolleschi:
> Do I need to fix linting issues or have they been corrected 

You have to. The error tells you also how to run a command that fixes them automatically. Then you still have to commit and push as you did earlier.

PR #3382, opened 2022-10-26
Mention snapToInterval doesn't work on web

sreuter:
See [react-native-web code path](https://github.com/necolas/react-native-web/blob/80ef32f6eea4a7950c07b1ea71b45b5c2b0269ee/packages/react-native-web/src/exports/ScrollView/index.js#L166-L209)

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -669,7 +669,7 @@ Use in conjunction with `snapToOffsets`. By default, the end of the list counts 
 
 ### `snapToInterval`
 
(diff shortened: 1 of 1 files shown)


Simek:
Hey @sreuter, thanks for sending the PR! 👍 

The official documentation focuses on native and build-in platforms. Out-of-tree platforms usually have their own docs, in case of web you should propose the change here:
* https://github.com/necolas/react-native-web
* https://necolas.github.io/react-native-web/docs/

PR #3387, opened 2022-10-28
fix: missing paren in publishing

pcasdf:
Adding a missing paren in the sentence under **Publish your own version of React Native**.


Simek (approved):
Thank you for the correction! 👍

PR #3388, opened 2022-10-31
Update clipboard.md

ramirobg94:

### DIFF
--- a/docs/clipboard.md
+++ b/docs/clipboard.md
@@ -31,10 +31,10 @@ const App = () => {
   return (
     <SafeAreaView style={{ flex: 1 }}>
       <View style={styles.container}>
-        <TouchableOpacity onPress={() => copyToClipboard()}>
+        <TouchableOpacity onPress={copyToClipboard}>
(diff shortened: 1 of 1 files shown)


NickGerleman:
Out of curiosity, how did you end up stumbling into this page? I can't seem to find it linked anywhere, and the component was removed a coupe releases ago. Probably best to remove the page.

Simek:
@NickGerleman We retain the pages for the removed/extracted components and APIs to avoid 404 from old resources and avoid defining redirects, but they are not linked in any of sidebars, for example:
* https://reactnative.dev/docs/clipboard
* https://reactnative.dev/docs/imagepickerios

There is an information at the top already, but maybe it's a good idea to make those pages more like stubs, without an Example and propos listed - only removal notice and a short description what it was, wdyt?

Simek:
Hey @ramirobg94, thanks for the contribution! 👍  Since the Clipboard API has been removed the core a while ago, we recently also cleaned up the pages content, leaving only the notice, so the correction should no longer be needed:
* https://reactnative.dev/docs/next/clipboard

PR #3392, opened 2022-11-02
Cleanup mentions of RN versions in Hermes docs

hramos:
Hermes is the default as of React Native 0.70, and bundled Hermes was introduced in 0.69.

The versioned docs for 0.69 and 0.70 have been updated to remove the explicit mention of their own version numbers:

- When looking at versions of the docs for 0.69 and later, we just mention that Hermes is now bundled into RN.
- When looking at versions of the docs for 0.70 and later, we just mention that Hermes is now the default.

Since the docs are versioned, we do not need to call out when something was introduced unless it is directly relevant, e.g. we still make the caveat that Hermes by default is a 0.70+ feature under "Enabling Hermes on Older Versions of React Native" (relevancy) and in "Architecture/Bundled Hermes" (unversioned guide).


hramos:
Closing in favor of https://github.com/facebook/react-native-website/pull/3393

PR #3394, opened 2022-11-03
Don't document VirtualizedList context

NickGerleman:
VirtualizeList internally uses contexts to allow buidling tree of nested VirtualizedList refs. This is an API really meant to only be called by VirtualizedList, but is public so other VirtualizedList instances can call it. The object shape, parameters, etc can change (and did change a bit in 0.71). Remove it from documentation, since folks should not rely on it.

### DIFF
--- a/docs/virtualizedlist.md
+++ b/docs/virtualizedlist.md
@@ -514,25 +514,6 @@ flashScrollIndicators();
 
 ---
 
-### `getChildContext()`
-
-```jsx
-getChildContext () => Object;
-```
-
-The `Object` returned consist of:
-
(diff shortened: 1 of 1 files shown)


Simek (approved):
LGTM, thank you for providing a context for this removal! 👍 

Maybe we should remove this method also from the versioned docs, wdyt?

NickGerleman:
> LGTM, thank you for providing a context for this removal! 👍
> 
> Maybe we should remove this method also from the versioned docs, wdyt?

I wouldn't be against it. Some of the other imperative methods here like recordInteraction() are also a bit private, so it might not hurt for me to dig through everything listed for the component and past revisions.

PR #3396, opened 2022-11-04
feat: Add aria Web props to Accessibility docs

gabrieldonadel:
This adds documentation regarding all aria Web props recently added to react-native due to  https://github.com/facebook/react-native/issues/34424. The description of most of these items was based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes.

Related to https://github.com/facebook/react-native/commit/98d84e571df64b00e1ed3484923a1d8169dcfda1, https://github.com/facebook/react-native/commit/ebdb23c6e060897112b060c25172135d50c78cd3, https://github.com/facebook/react-native/commit/720cdbc658aa0b513d106526eb81d749fa957c99, https://github.com/facebook/react-native/commit/f353119113d6fc85491765ba1e90ac83cb00fd61 and https://github.com/facebook/react-native/commit/7ea54a408734ccde9ee01a12942bf5722e9cf7c1

PR #3404, opened 2022-11-07
feat: Add aria Web props to component docs

gabrieldonadel:
Follow up PR of https://github.com/facebook/react-native-website/pull/3396. 

This adds documentation to `View`, `Text` and `TouchableWithoutFeedback` regarding all aria Web props recently added to react-native due to  https://github.com/facebook/react-native/issues/34424. The description of most of these items was based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes.

Related to https://github.com/facebook/react-native/commit/98d84e571df64b00e1ed3484923a1d8169dcfda1, https://github.com/facebook/react-native/commit/ebdb23c6e060897112b060c25172135d50c78cd3, https://github.com/facebook/react-native/commit/720cdbc658aa0b513d106526eb81d749fa957c99, https://github.com/facebook/react-native/commit/f353119113d6fc85491765ba1e90ac83cb00fd61 and https://github.com/facebook/react-native/commit/7ea54a408734ccde9ee01a12942bf5722e9cf7c1

### DIFF
--- a/docs/touchablewithoutfeedback.md
+++ b/docs/touchablewithoutfeedback.md
@@ -265,11 +265,15 @@ Indicates that an element will be updated, and describes the types of updates th
 - **polite** Accessibility services should announce changes to this view.
 - **assertive** Accessibility services should interrupt ongoing speech to immediately announce changes to this view.
 
+| Type                                     | Default |
+| ---------------------------------------- | ------- |
+| enum(`'assertive'`, `'off'`, `'polite'`) | `'off'` |
+
 ---
 
 ### `aria-modal` <div class="label ios">iOS</div>
 
-Boolean value indicating whether VoiceOver should ignore the elements within views that are siblings of the receiver. Has precedence over the `accessibilityViewIsModal` prop.
+Boolean value indicating whether VoiceOver should ignore the elements within views that are siblings of the receiver. Has precedence over the [`accessibilityViewIsModal`](#accessibilityviewismodal-ios) prop.
 
 | Type    | Default |
 | ------- | ------- |
--- a/docs/touchablewithoutfeedback.md
+++ b/docs/touchablewithoutfeedback.md
@@ -195,6 +195,98 @@ See the [Accessibility guide](accessibility.md#accessibility-actions) for more i
 
 ---
 
+'aria-selected'?: ?boolean,
+
+### `aria-busy`
+
+Indicates an element is being modified and that assistive technologies may want to wait until the changes are complete before informing the user about the update.
+
+| Type    | Default |
+| ------- | ------- |
+| boolean | false   |
+
+---
+
+### `aria-checked`
+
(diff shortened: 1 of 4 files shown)


Simek (approved):
Thank you for the follow up PR! 👍 

I have only one suggestion - can you order the props alphabetically? Definitely needed in case of `View`, where we have some ARIA props already.

gabrieldonadel:
> Thank you for the follow up PR! 👍
> 
> I have only one suggestion - can you order the props alphabetically? Definitely needed in case of `View`, where we have some ARIA props already.

Sure, fixed

docs/touchablewithoutfeedback.md
```diff
@@ -195,6 +195,98 @@ See the [Accessibility guide](accessibility.md#accessibility-actions) for more i
 
 ---
 
+'aria-selected'?: ?boolean,
+
```
Simek, line 199:
```suggestion
```

Looks like an artefact while moving entries.

gabrieldonadel: My bad, removed

docs/touchablewithoutfeedback.md
```diff
@@ -195,6 +195,96 @@ See the [Accessibility guide](accessibility.md#accessibility-actions) for more i
...
+---
+
+### `aria-live` <div class="label android">Android</div>
+
+Indicates that an element will be updated, and describes the types of updates the user agents, assistive technologies, and user can expect from the live region.
+
+- **off** Accessibility services should not announce changes to this view.
+- **polite** Accessibility services should announce changes to this view.
+- **assertive** Accessibility services should interrupt ongoing speech to immediately announce changes to this view.
+
```
Simek, line 267: Can you also add the Type table for this prop and specify the default value? For example like in those props:
* https://reactnative.dev/docs/next/textinput#autocapitalize
* https://reactnative.dev/docs/next/image#imagecacheenum-ios

docs/view.md
```diff
@@ -249,6 +249,113 @@ When `true`, indicates that the view is an accessibility element. By default, al
...
+
+- **off** Accessibility services should not announce changes to this view.
+- **polite** Accessibility services should announce changes to this view.
+- **assertive** Accessibility services should interrupt ongoing speech to immediately announce changes to this view.
+
+---
+
+### `aria-modal` <div class="label ios">iOS</div>
+
+Boolean value indicating whether VoiceOver should ignore the elements within views that are siblings of the receiver. Has precedence over the `accessibilityViewIsModal` prop.
```
Simek, line 343:
```suggestion
Boolean value indicating whether VoiceOver should ignore the elements within views that are siblings of the receiver. Has precedence over the [`accessibilityViewIsModal`](#accessibilityviewismodal-ios) prop.
```

Let's link the other props in the description if possible for the UX.

gabrieldonadel: done

Simek (commented):
Thank you for all of the updates! 🙏

PR #3422, opened 2022-11-18
Lint/Format all the snack examples

NickGerleman:
This adds a script which can run ESLint on each Snack example in Markdown. This is wired to the default RN ESLint configuration. Apart from enforcing consistency, it means copy/pasting examples into a new RN project will avoid linting and formatting errors. 

These are exposed as `yarn format:examples` and `yarn lint:examples` scripts to the website package, but the script to lint is its own package.

I cleaned up the errors, so we can enable this in the `ci:lint` job, but we still have many warnings.


Simek:
In the past we tried the similar approach, but at the end we decided to ditch the idea due to the multiline imports, which in some cases took up to 1/2 of initially visible example, which felt like UX degradation. If we could avoid that it would be great.

Also, what's the reason for crating a lot of alternate configuration files for the snacks link? Cannot we just use current setup and expand it?

I would also move all the added dependencies to the `website` project in monorepo, since those changes/script do not apply to the other packages in the repo.

Simek:
I also feel like we can achieve the similar result just adding https://github.com/eslint/eslint-plugin-markdown to the `website` ESLint stack/config. 😅

NickGerleman:
> In the past we tried the similar approach, but at the end we decided to ditch the idea due to the multiline imports, which in some cases took up to 1/2 of initially visible example, which felt like UX degradation. If we could avoid that it would be great.

Do you have any good examples for this? I think the only control we have here is for max column width.

> Also, what's the reason for crating a lot of alternate configuration files for the snacks link? Cannot we just use current setup and expand it?

The idea is to use the out-of-the-box configuration of a react-native app so that the samples don't create linter errors if copied to a project. Ignoring formatting, we have quite a few of these across the examples. My guess is that we probably don't want to make the website inherit from `@react-native-community/eslint-config`, like we do want to for the examples.
image

> Not sure about Typescript addition tho, is it really needed?

TypeScript examples are coming soon, and the out-of-the-box new app template will run `typescript-eslint`. I think we could move dependencies of the real-app linting setup in its own package instead of the website one if we wanted to isolate these a bit more, but a little more overhead.

> I also feel like we can achieve the similar result by just adding https://github.com/eslint/eslint-plugin-markdown to the `website` ESLint stack/config. 😅

TIL ESLint processors are a thing. It seems like a much better solution if we are just targeting ESLint, but I was also looking to followup with adding `tsc` to typecheck the examples (e.g. to catch out of date examples).

Simek:
> Do you have any good examples for this? I think the only control we have here is for max column width.

AFAIK there is no other option to control this other than max column width, and that's why we opted out earlier, since changing the max allowed width affected other parts of examples negatively.

> The idea is to use the out-of-the-box configuration of a react-native app so that the samples don't create linter errors if copied to a project.

This is a good reason, make sense, thank you for clarifying!

> My guess is that we probably don't want to make the website inherit from @react-native-community/eslint-config, like we do want to for the examples.

It might depends on the ruleset of this plugin, but probably you are right.

> TypeScript examples are coming soon, and the out-of-the-box new app template will run typescript-eslint.

That's great to hear, and it make sense to add TS with that context in mind.

> It seems like a much better solution if we are just targeting ESLint, but I was also looking to followup with adding tsc to typecheck the examples (e.g. to catch out of date examples).

Having in mind the separation idea and goal to validate examples with the templated default rules, it is probably better to to not mix those two lint stacks together. 🙂

NickGerleman:
> AFAIK there is no other option to control this other than max column width, and that's why we opted out earlier, since changing the max allowed width affected other parts of examples negatively.

I can deploy a test of 80 line vs 70 line. 80 lines will cause wrapping, but should avoid the verbose imports more often, and means the formatting is exactly as expected by the preset.

NickGerleman:
> Having in mind the separation idea and goal to validate examples with the templated default rules, it is probably better to to not mix those two lint stacks together. 🙂

The shareable part here is the mechanism of "adapt examples in Markdown to files which could be verified by a relatively stock config, then patch back". So, if we have that mechanism in place, we can use relatively close to stock configs for ESLint, tsc, jest, or whatever other tool comes along.

NickGerleman:
The script driving patching + ESLint should be much faster/more reasonable now as well. Running at 4s locally. 
image

NickGerleman:
Looked closer at `eslint-plugin-markdown`, and it would let us override rules for a filetype of "**/*.md/*.SnackPlayer" (and maybe the string of variables attached to it?). We would not be able to reuse the stock react native ESLint config (since it just overrides ".js/.jsx/.ts/.tsx"). We also lose the names of the Snack in the files being shown during linting, and instead just get an index. Having ESLint be the driver wouldn't actually get us editor integration either, since the website would be using a different version of the dependencies.

NickGerleman:
@Simek @rickhanlonii @yungsters would you be able to take another look?

I have local changes ready to make all of the "core component" examples TypeScript-ready (most are compatible with JS and inferable by TS) and removing some of the lint warnings. I am not able to push TypeScript-specific quite yet (see SnackPlayer refresh bug in #3426 I still am debugging), but am hoping to use this to avoid needing to manually reformat all of the examples, and to add a way to enforce typechecking.

Simek (approved):
Tested locally, LGTM! Thank you for working on this tool and for pushing on this idea to unify examples style! 🙏

PR #3423, opened 2022-11-18
Lint/Format all the snack examples (80 char)

NickGerleman:
Same as https://github.com/facebook/react-native-website/pull/3422 but sizes to the slightly more spacious, and more standard 80 char limit instead of the current snack player's limit of 70 before wrapping.

See 40cea231171d17ed3561c8792504fceb97b699e5 for the diff of moving from 70 to 80.


Simek:
Thank you for making an effort to compare both settings! 🙏 I think that `80` characters limit results in a better formatted examples for our use case.

NickGerleman:
Sounds good, I'll fold it into the other change.

PR #3426, opened 2022-11-19
Allow TypeScript in Snacks (and optional "ext" parameter)

NickGerleman:
This makes a couple changes in preparation for TypeScript examples. Going through 20 pages for core components, most examples can be shifted to be both JavaScript and strict TypeScript. Some are not compatible as both, and need to be separated as JS examples vs TS examples.

**Add parameter to SnackPlayer blocks**

Snack defaults a code block to represent a file named [`App.js`](https://github.com/expo/snack/blob/main/docs/url-query-parameters.md#parameters). This extension effects how the file is parsed during bundling, and the syntax highlighting it is given in the editor. We allow passing the extension, so that specific examples can specifically target JS vs TS. This directive will later be used to determine how the file is used during linting.

**Change default example extension to `.tsx`**

This is needed for Snack to parse an input as TypeScript during bundling and syntax highlighting. Vanilla JS is still allowed, and it seems to create an identical result to before in the snack player.

### DIFF
--- a/plugins/remark-snackplayer/src/index.js
+++ b/plugins/remark-snackplayer/src/index.js
@@ -32,8 +32,16 @@ const processNode = (node, parent) => {
       const description = params.description
         ? decodeURIComponent(params.description)
         : 'Example usage';
-      const sampleCode = node.value;
-      const encodedSampleCode = encodeURIComponent(sampleCode);
+      const ext = params.ext ? decodeURIComponent(params.ext) : 'tsx';
+      const filename = `App.${ext}`;
+      const files = encodeURIComponent(
+        JSON.stringify({
+          [filename]: {
+            type: 'CODE',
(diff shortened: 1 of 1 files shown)


NickGerleman:
This is causing an issue, where the Snack loads on first page visit, but not when switching page.

NickGerleman:
> This is causing an issue, where the Snack loads on first page visit, but not when switching page.

This is due to a bug in SnackPlayer. Submitted a fix with https://github.com/expo/snack/pull/361, but added a workaround in the meantime.

lunaleaps (approved):

NickGerleman:
https://github.com/expo/snack/pull/361 was deployed, and examples on switching works correctly now without the workaround 👍. Removed the the empty `data-snack-code`.

PR #3429, opened 2022-11-22
Update TypeScript Page and References to TypeScript

NickGerleman:
This change updates the main TypeScript page with new instructions for 0.71, where TypeScript is the default language, and RN types are included by default.

This does not yet update samples for TS, which is still a work in progress. I also chose to omit information about the 0.70 to 0.71 migration, since it seems like it might make more sense in either the release announcement or a separate blog post instead of long-term in the documentation, but I would be open to feedback on that.

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -5,91 +5,43 @@ title: Using TypeScript
 
 import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
 
-[TypeScript][ts] is a language which extends JavaScript by adding type definitions, much like [Flow][flow]. While React Native is built in Flow, it supports both TypeScript _and_ Flow by default.
+[TypeScript][ts] is a language which extends JavaScript by adding type definitions. New React Native projects target TypeScript by default, but also support JavaScript and Flow.
+
 
 ## Getting Started with TypeScript
 
-If you're starting a new project, there are a few different ways to get started.
+New projects created by the [React Native CLI](http://localhost:3000/docs/next/environment-setup#creating-a-new-application) or popular templates like [Ignite][ignite] will use TypeScript by default.
 
-You can use the [TypeScript template][ts-template]:
-
-```shell
-npx react-native init MyApp --template react-native-template-typescript
-```
-
-:::note
-
-If the above command is failing, you may have an old version of `react-native` or `react-native-cli` installed globally on your system. To fix the issue try uninstalling the CLI:
-
-- `npm uninstall -g react-native-cli` or `yarn global remove react-native-cli`
-
-and then run the `npx` command again.
-Optionally, you can also use the command given below to get started with your template.
-
-- `npx react-native --ignore-existing init MyApp --template react-native-template-typescript`
-
-:::
-
-You can use [Expo][expo], which maintains TypeScript templates, or will prompt you to automatically install and configure TypeScript when a `.ts` or `.tsx` file is added to your project:
-
-<Tabs groupId="package-manager" defaultValue={constants.defaultPackageManager} values={constants.packageManagers}>
-<TabItem value="npm">
+TypeScript may also be used with [Expo][expo], which maintains TypeScript templates, or will prompt you to automatically install and configure TypeScript when a `.ts` or `.tsx` file is added to your project.
 
 ```shell
 npx create-expo-app --template
 ```
 
-</TabItem>
-<TabItem value="yarn">
-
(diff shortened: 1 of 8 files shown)


docs/typescript.md
```diff
@@ -5,92 +5,44 @@ title: Using TypeScript
 
 import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
 
-[TypeScript][ts] is a language which extends JavaScript by adding type definitions, much like [Flow][flow]. While React Native is built in Flow, it supports both TypeScript _and_ Flow by default.
+[TypeScript][ts] is a language which extends JavaScript by adding type definitions. New React Native projects target TypeScript by default, but also support JavaScript and Flow.
 
-## Getting Started with TypeScript
```
NickGerleman, line 10: This section is overall a lot shorter, since the default new app setup should do this. So it now mostly just links to the environment setup page, with a specific callout for Expo.

docs/typescript.md
```diff
@@ -5,92 +5,44 @@ title: Using TypeScript
...
 
 ```shell
 npx create-expo-app --template
 ```
 
-</TabItem>
-<TabItem value="yarn">
-
-```shell
-yarn create expo-app --template
```
NickGerleman, line 46: The yarn examples here were actually totally wrong and wouldn't work. Should just be `npx` for executing remote dependencies.

docs/typescript.md
```diff
@@ -99,12 +51,14 @@ yarn add -D typescript @types/jest @types/react @types/react-native @types/react
 }
 ```
 
-3. Create a `jest.config.js` file to configure Jest to use TypeScript
```
NickGerleman, line 102: Jest, at least in the current version, supports ts, tsx out of the box, and we don't need to add a new config anymore.

docs/typescript.md
```diff
@@ -114,9 +67,13 @@ module.exports = {
 
 5. Run `yarn tsc` to type-check your new TypeScript files.
 
+## Using JavaScript Instead of TypeScript
+
+React Native defaults new applications to TypeScript, but JavaScript may still be used by default. Files with a `.jsx` extension are treated as JavaScript instead of TypeScript, and will not be typechecked. JavaScript modules may still be imported by TypeScript modules, along with the reverse.
```
NickGerleman, line 72:
```suggestion
React Native defaults new applications to TypeScript, but JavaScript may still be used. Files with a `.jsx` extension are treated as JavaScript instead of TypeScript, and will not be typechecked. JavaScript modules may still be imported by TypeScript modules, along with the reverse.
```

docs/typescript.md
```diff
@@ -5,92 +5,43 @@ title: Using TypeScript
...
-:::
-
-You can use [Expo][expo], which maintains TypeScript templates, or will prompt you to automatically install and configure TypeScript when a `.ts` or `.tsx` file is added to your project:
-
-<Tabs groupId="package-manager" defaultValue={constants.defaultPackageManager} values={constants.packageManagers}>
-<TabItem value="npm">
+TypeScript may also be used with [Expo][expo], which maintains TypeScript templates, or will prompt you to automatically install and configure TypeScript when a `.ts` or `.tsx` file is added to your project.
 
 ```shell
 npx create-expo-app --template
```
lunaleaps, line 17: one question about this -- why do we need to pass the empty `--template` flag? 

By the cli docs it seems to imply it'll guide you through to choose the template but trying out it doesn't seem to?

```
❯ npm create expo-app --template
✔ What is your app named? … my-app
✔ Downloaded and extracted project files.
✔ Installed JavaScript dependencies.

✅ Your project is ready!

To run your project, navigate to the directory and run one of the following npm commands.

- cd my-app
- npm run android
- npm run ios
- npm run web

~ via  v18.7.0 took 36s
❯ cd my-app/
```

Simek: Without the `--template` the default template is used without listing the options.

Simek: Ah, I see the problem. Running `npm create expo-app` seems to do not forward the flag, that's probably why we recommend `npx create-expo-app` over it on this page in the package Readme.

brentvatne: with `npm create` you need to use `npm create expo-app -- --template` in order to differentiate between flags for `npm create` and flags for `create-expo-app`

lunaleaps: Ohoo my mistake! I think I had copied the wrong command over from some other documentation, thanks for clarifying!

PR #3444, opened 2022-12-01
Update TS Instructions for Simplified Template

NickGerleman:
https://github.com/facebook/react-native/pull/35529 simplifies the ESLint config of the template a bit more compared to the previous version, more directly derived from the community template. I expect it will be cherry-picked into 0.71, so this updates the corresponding TS documentation on adding TS to a non-TS project to remove the extra differences.


cortinico (approved):

Simek (approved):
🚀

PR #3450, opened 2022-12-03
Start TypeScriptify-ing the examples

NickGerleman:
This change starts migrating the examples to TypeScript. Every example is now linted as both tsx and js, unless an extension is specified. All of the existing ESLint warnings, minus those for inline styles which I have disabled for now, are addressed, so we can now flag new lint warnings when examples are added. A script `tsc-examples` is added, in addition to ESLint, but I haven't folded it into `lint:examples` yet because not everything has been fixed, and doing so will take multiple PRs.

100/171 examples are clean for both ESLint and TypeScript in strict mode, with an additional in this change opted out of typechecking (Clipboard) since it has been removed and is no longer part of typings. Most examples are already valid as both JSX and strict TSX, but some are separated into tabs (where TypeScript is shown first and by default), require changes so that TypeScript can infer the types (while still being valid JS), or require fixup (e.g. due to changed APIs). Some pages like Systrace are out of date/broken and will need to be opted out of typechecking (and given a warning) while they are in disrepair.

### DIFF
--- a/docs/accessibilityinfo.md
+++ b/docs/accessibilityinfo.md
@@ -23,22 +23,22 @@ const App = () => {
   useEffect(() => {
     const reduceMotionChangedSubscription = AccessibilityInfo.addEventListener(
       'reduceMotionChanged',
-      reduceMotionEnabled => {
-        setReduceMotionEnabled(reduceMotionEnabled);
+      isReduceMotionEnabled => {
+        setReduceMotionEnabled(isReduceMotionEnabled);
       },
     );
     const screenReaderChangedSubscription = AccessibilityInfo.addEventListener(
       'screenReaderChanged',
-      screenReaderEnabled => {
-        setScreenReaderEnabled(screenReaderEnabled);
+      isScreenReaderEnabled => {
+        setScreenReaderEnabled(isScreenReaderEnabled);
       },
     );
 
-    AccessibilityInfo.isReduceMotionEnabled().then(reduceMotionEnabled => {
-      setReduceMotionEnabled(reduceMotionEnabled);
+    AccessibilityInfo.isReduceMotionEnabled().then(isReduceMotionEnabled => {
+      setReduceMotionEnabled(isReduceMotionEnabled);
     });
-    AccessibilityInfo.isScreenReaderEnabled().then(screenReaderEnabled => {
-      setScreenReaderEnabled(screenReaderEnabled);
+    AccessibilityInfo.isScreenReaderEnabled().then(isScreenReaderEnabled => {
+      setScreenReaderEnabled(isScreenReaderEnabled);
     });
 
     return () => {
@@ -76,7 +76,10 @@ export default App;
 </TabItem>
 <TabItem value="classical">
 
-```SnackPlayer name=AccessibilityInfo%20Class%20Component%20Example&supportedPlatforms=android,ios
+<Tabs groupId="language" defaultValue={constants.defaultSnackLanguage} values={constants.snackLanguages}>
+<TabItem value="javascript">
+
+```SnackPlayer name=AccessibilityInfo%20Class%20Component%20Example&supportedPlatforms=android,ios&ext=js
 import React, {Component} from 'react';
 import {AccessibilityInfo, View, Text, StyleSheet} from 'react-native';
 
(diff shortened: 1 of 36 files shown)


docs/animated.md
```diff
@@ -49,6 +49,7 @@ const App = () => {
     Animated.timing(fadeAnim, {
       toValue: 1,
       duration: 5000,
+      useNativeDriver: true,
```
NickGerleman, line 52: `useNativeDriver` is required by the TS typings, and if not specified will lead to a runtime warning saying it is required.

docs/backhandler.md
```diff
@@ -152,69 +155,16 @@ export default App;
 ```
 
 </TabItem>
-</Tabs>
-
-`BackHandler.addEventListener` creates an event listener & returns a `NativeEventSubscription` object which should be cleared using `NativeEventSubscription.remove` method.
-
-Additionally `BackHandler.removeEventListener` can also be used to clear the event listener. Ensure the callback has the reference to the same function used in the `addEventListener` call as shown the following example ﹣
```
NickGerleman, line 159: This method of removing event handler subscriptions has been deprecated for a while and we shouldn't recommend it as an alternative to the right way to.

docs/accessibilityinfo.md
```diff
@@ -23,22 +23,22 @@ const App = () => {
   useEffect(() => {
     const reduceMotionChangedSubscription = AccessibilityInfo.addEventListener(
       'reduceMotionChanged',
-      reduceMotionEnabled => {
-        setReduceMotionEnabled(reduceMotionEnabled);
+      isReduceMotionEnabled => {
```
NickGerleman, line 26: A lot of these renamings throughout the PR are to avoid shadowing (so we can get clean of ESLint warnings)

docs/accessibilityinfo.md
```diff
@@ -143,6 +146,83 @@ const styles = StyleSheet.create({
 export default AccessibilityStatusExample;
 ```
 
+</TabItem>
+<TabItem value="typescript">
```
NickGerleman, line 150: General note: even though the TS tab comes after in the document, it is still shown first on the actual webpage (controlled by the TabConstants ordering).

docs/handling-touches.md
```diff
@@ -26,11 +26,11 @@ Go ahead and play around with the `Button` component using the example below. Yo
 
 ```SnackPlayer name=Button%20Basics
 import React, {Component} from 'react';
-import {Button, StyleSheet, View} from 'react-native';
+import {Alert, Button, StyleSheet, View} from 'react-native';
```
NickGerleman, line 29: RN's current ESLint config enables `no-alert`, which bans the raw `alert()` API. But using the RN `Alert` API is seen as fine.

docs/layout-props.md
```diff
@@ -167,7 +167,7 @@ const Square = () => (
 
 const randomHexColor = () => {
   return '#000000'.replace(/0/g, () => {
-    return (~~(Math.random() * 16)).toString(16);
+    return Math.round(Math.random() * 16).toString(16);
```
NickGerleman, line 170: This was previously doing two's complement negation twice, but we can also just round for a similar effect (tested this out). The ESLint config bans bitwise operators.

scripts/lint-examples/src/lintExamples.js
```diff
@@ -11,41 +10,70 @@
...
+ * Process arguments to be forwarded to the linter
+ */
+const processArgs = process.argv.slice(2);
 
 /**
- * Main function
+ * Valid extensions for snack examples
  */
-(async () => {
+const validExtensions = ['js', 'tsx'];
```
NickGerleman, line 33: It seems kind of odd to use `tsx`, but not `jsx`, but Snack seems to not support `jsx`, and the RN 0.71 ESLint config also no-oped on JSX. So instead we check that it is either `js` or `tsx` to avoid the inevitable confusion if someone adds `jsx` or `ts`.

scripts/lint-examples/tsconfig.json
```diff
@@ -1,3 +1,4 @@
 {
-  "extends": "@tsconfig/react-native/tsconfig.json"
+  "extends": "@tsconfig/react-native/tsconfig.json",
+  "include": ["./out"]
```
NickGerleman, line 3: This is needed to avoid tsc seeing the scripts directory and getting confused.

scripts/lint-examples/package.json
```diff
@@ -3,14 +3,17 @@
...
+    "tsc-examples": "./bin/tsc-examples.js"
   },
-  "dependencies": {
+  "devDependencies": {
     "@babel/core": "^7.12.9",
     "@babel/preset-env": "^7.14.0",
     "@babel/runtime": "^7.12.5",
     "@react-native-community/eslint-config": "^3.0.0",
     "@tsconfig/react-native": "^2.0.2",
+    "@types/jest": "^29.2.3",
```
NickGerleman, line 16: Needed because @tsconfig/react-native will try to add global types from @types/jest, even if we don't use Jest ourselves.

scripts/lint-examples/bin/tsc-examples.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @format
+ */
+
+const lintExamples = require('../src/lintExamples');
+
+lintExamples({
+  command: 'tsc',
+  extension: 'tsx',
+  writeBack: true,
```
NickGerleman, line 16:
```suggestion
  writeBack: false,
```

docs/animated.md
lunaleaps: Should we default this to `false`? Is there a preference?

docs/animated.md
```diff
@@ -311,7 +315,7 @@ Config is an object that may have the following options:
 - `velocity`: Initial velocity. Required.
 - `deceleration`: Rate of decay. Default 0.997.
 - `isInteraction`: Whether or not this animation creates an "interaction handle" on the `InteractionManager`. Default true.
-- `useNativeDriver`: Uses the native driver when true. Default false.
+- `useNativeDriver`: Uses the native driver when true. Required.
```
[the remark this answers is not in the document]
NickGerleman: If it will issue a warning saying it is required, the type system enforcing it is present seems good to me. My guess is that there were too many existing usages to do the same in Flow.

docs/layout-props.md
lunaleaps: TIL!

lunaleaps (approved):

PR #3452, opened 2022-12-04
Updated the installation instructions

Smyja:
The guide needs to mention how to install the expo cli else the `expo start` command won't run, i have added it.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -17,12 +17,13 @@ This page will help you install and build your first React Native app.
 <Tabs groupId="guide" defaultValue={constants.defaultGuide} values={constants.guides}>
 <TabItem value="quickstart">
 
-Run the following command to create a new React Native project called "AwesomeProject":
+Run the following command to install expo cli and create a new React Native project called "AwesomeProject":
 
 <Tabs groupId="package-manager" defaultValue={constants.defaultPackageManager} values={constants.packageManagers}>
 <TabItem value="npm">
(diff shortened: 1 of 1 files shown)


Simek:
Hello @Smyja, currently using `npx expo start` is recommended instead of installing global Expo CLI:
* https://blog.expo.dev/expo-sdk-46-c2a1655f63f7#ddca

Could you alter the changes, to match this recommendation? Also, you will need to accept the CLA, so we can merge those changes in..

Simek:
Thank you for the quick update! 👍  

The `npx expo start` is mention in the code block in a later step, and I think we should stick to that order, since you should not need CLI to initiate the template. So, in other word, we should remove the added install CLI mention and just make the comments from the code block the recommended way, so:
```sh
npm start # you can also use: npx expo start
``` 
can be now just:
```sh
npx expo start
``` 
And the same with Yarn instructions.

Smyja:
> Thank you for the quick update! 👍
> 
> The `npx expo start` is mention in the code block in a later step, and I think we should stick to that order, since you should not need CLI to initiate the template. So, in other word, we should remove the added install CLI mention and just make the comments from the code block the recommended way, so:
> 
> ```shell
> npm start # you can also use: npx expo start
> ```
> 
> can be now just:
> 
> ```shell
> npx expo start
> ```
> 
> And the same applies to the Yarn instructions.

I have updated it.

PR #3457, opened 2022-12-06
Update ignite blurb on more resources

KrastanD:


KrastanD:
@cortinico looks like recreating the PR worked

PR #3460, opened 2022-12-07
Typescriptify Examples H-J

NickGerleman:
This makes examples in documents starting from H-J valid for TypeScript. This leaves 36 examples left. Verified these pages locally after the change.

### DIFF
--- a/docs/image-style-props.md
+++ b/docs/image-style-props.md
@@ -81,7 +81,7 @@ const DisplayAnImageWithStyle = () => {
 const styles = StyleSheet.create({
   container: {
     display: 'flex',
-    flexDirection: 'vertical',
+    flexDirection: 'column',
     justifyContent: 'space-around',
     alignItems: 'center',
     height: '100%',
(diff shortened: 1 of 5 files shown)


docs/intro-react.md
```diff
@@ -355,8 +469,8 @@ import React, { useState } from 'react';
 
 Then you declare the component’s state by calling `useState` inside its function. In this example, `useState` creates an `isHungry` state variable:
 
-```jsx
-const Cat = (props) => {
+```tsx
```
NickGerleman, line 472: Going to take a separate pass through inline jsx later to change to tsx. Though they cannot be linted/typechecked since these are not full modules.

docs/image-style-props.md
```diff
@@ -81,7 +81,7 @@ const DisplayAnImageWithStyle = () => {
 const styles = StyleSheet.create({
   container: {
     display: 'flex',
-    flexDirection: 'vertical',
```
lunaleaps, line 84: This isn't even valid value is it?

lunaleaps (approved):

docs/image-style-props.md
NickGerleman: Nope lol

NickGerleman: Column is already the default thief which is probably why it wasn’t surfaced before.

PR #3461, opened 2022-12-08
Typescriptify Examples K-P

NickGerleman:
This makes examples in documents starting from K-P valid for TypeScript. This leaves 23 examples left. Verified these pages locally after the change.

### DIFF
--- a/docs/linking.md
+++ b/docs/linking.md
@@ -114,7 +114,10 @@ You can handle these events with `Linking.getInitialURL()` - it returns a Promis
 
 ### Open Links and Deep Links (Universal Links)
 
-```SnackPlayer name=Linking%20Function%20Component%20Example&supportedPlatforms=ios,android
+<Tabs groupId="language" defaultValue={constants.defaultSnackLanguage} values={constants.snackLanguages}>
+<TabItem value="javascript">
+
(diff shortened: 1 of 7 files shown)


docs/linking.md
```diff
@@ -274,7 +441,8 @@ const App = () => {
         action="android.settings.APP_NOTIFICATION_SETTINGS"
         extras={[
           {
-            'android.provider.extra.APP_PACKAGE': 'com.facebook.katana',
+            key: 'android.provider.extra.APP_PACKAGE',
+            value: 'com.facebook.katana',
```
NickGerleman, line 445: Confirmed the example here was wrong, rather than the typings

docs/panresponder.md
```diff
@@ -93,15 +93,9 @@ const App = () => {
...
-      onPanResponderGrant: () => {
-        pan.setOffset({
-          x: pan.x._value,
-          y: pan.y._value,
-        });
-      },
       onPanResponderMove: Animated.event([null, {dx: pan.x, dy: pan.y}]),
       onPanResponderRelease: () => {
-        pan.flattenOffset();
+        pan.extractOffset();
```
NickGerleman, line 98: This has the same effect as before without relying on private member variables from Animated.value() like the previous example did.

lunaleaps (approved):

PR #3463, opened 2022-12-08
Fix rendering of snacks on route update on mobile

NickGerleman:
Switching pages on mobile currently causes a blank snack example to show.

image

This is root caused to a workaround on the `onRouteUpdate` callback to read the DOM after mutation, which seems to work on desktop, but not on the mobile version of the site. This change moves to a newer lifecycle method called after the document is rendered to the DOM, which fixes the issue.

### DIFF
--- a/website/snackPlayerInitializer.js
+++ b/website/snackPlayerInitializer.js
@@ -74,7 +74,7 @@ export default (() => {
 
   return {
     onRouteDidUpdate({location}) {
-      // console.log('onRouteUpdate', {location});
+      // console.log('onRouteDidUpdate', {location});
 
       initSnackPlayers();
       setupTabPanelsMutationObservers();
--- a/website/snackPlayerInitializer.js
+++ b/website/snackPlayerInitializer.js
@@ -73,15 +73,11 @@ export default (() => {
   setupThemeSynchronization();
 
   return {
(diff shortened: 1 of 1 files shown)


Simek (approved):
Nice catch, LGTM!

cortinico (approved):

leotm:
awesome ^

> called after the document is rendered to the DOM

just learning docusaurus, it deffo called after?

https://docusaurus.io/docs/advanced/client

> ...
> 4. The new location's route component gets rendered to DOM.
> `onRouteUpdate` will be called at event (2), and `onRouteDidUpdate` will be called at (4)

otherwise thought the lifecycle fn's called at start or during the events

NickGerleman:
> awesome ^
> 
> > called after the document is rendered to the DOM
> 
> just learning docusaurus, it deffo called after?
> 
> https://docusaurus.io/docs/advanced/client
> 
> > ...
> > 4. The new location's route component gets rendered to DOM.
> > `onRouteUpdate` will be called at event (2), and `onRouteDidUpdate` will be called at (4)
> 
> otherwise thought the lifecycle fn's called at start or during the events

I pulled it from the thread in the code comment about the issue, mentioning the "DidUpdate" variant is called after the content is rendered to the DOM.

PR #3274, opened 2022-08-22
Update Folly pod version in backward compatibility and pillars new arch docs

leotm:
Context: https://github.com/reactwg/react-native-new-architecture/discussions/68#discussioncomment-3398169

Close: #3273

TODO: <s>also update these?</s>
- [x] https://github.com/facebook/react-native-website/blob/main/docs/the-new-architecture/backward-compatibility-fabric-components.md
- [x] https://github.com/facebook/react-native-website/blob/main/docs/the-new-architecture/pillars-turbomodule.md
- [x] https://github.com/facebook/react-native-website/blob/main/docs/the-new-architecture/pillars-fabric-components.md
- [x] Check more refs
- [x] Consider reverting pre-0.70 changes

PR #3471, opened 2022-12-13
fix: remove explicit folly version as it's not needed anymore

okwasniewski:
Hey, 

This PR removes explicit folly versions as specifying them is not needed anymore.


leotm:
good spot ^ i did similar
- https://github.com/facebook/react-native-website/pull/3274

but don't think it's needed anymore

> we make some major changes in the doc and I think that this is not relevant anymore.

cc @cipolleschi

cipolleschi (approved):
Hi @okwasniewski, thanks a lot for doing this!

PR #3473, opened 2022-12-13
Typescriptify Examples Q-S

NickGerleman:
This makes examples in documents starting from Q-S valid for TypeScript. This leaves 9 examples left. Verified these pages locally after the change.

### DIFF
--- a/docs/stylesheet.md
+++ b/docs/stylesheet.md
@@ -194,7 +194,7 @@ const App = () => (
     <View style={styles.box1}>
       <Text style={styles.text}>1</Text>
     </View>
-    <View style={styles.box2}>
+    <View style={[styles.box2, StyleSheet.absoluteFill]}>
       <Text style={styles.text}>2</Text>
     </View>
     <View style={styles.box3}>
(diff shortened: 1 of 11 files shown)


docs/stylesheet.md
```diff
@@ -334,9 +333,3 @@ export default App;
 This constant will always be a round number of pixels (so a line defined by it can look crisp) and will try to match the standard width of a thin line on the underlying platform. However, you should not rely on it being a constant size, because on different platforms and screen densities its value may be calculated differently.
 
 A line with hairline width may not be visible if your simulator is downscaled.
-
----
-
-## `absoluteFill` vs. `absoluteFillObject`
-
-Currently, there is no difference between using `absoluteFill` vs. `absoluteFillObject`.
```
NickGerleman, line 342: So... technically this is true. But from code the intention is that `absoluteFillObject` is the style object and `absoluteFill` is the registered/created style. So it is supposed to be able to be opaque, but `StyleSheet.create()` was changed a while back to itself return an object. So it is dicey and we should not document that they are the same (and the TS typings act as if they are not).

lunaleaps (approved):

PR #3474, opened 2022-12-13
Typescriptify Examples T-Z (and enforce)

NickGerleman:
This makes examples in documents starting from T-Z valid for TypeScript, and enables typechecking as part of linting to enforce new examples are clean. Because of that, this change must be merged after https://github.com/facebook/react-native-website/pull/3473


docs/toastandroid.md
```diff
@@ -67,58 +67,6 @@ const styles = StyleSheet.create({
...
-      message,
-      ToastAndroid.LONG,
-      ToastAndroid.BOTTOM,
-      25,
-      50,
-    );
-    return null;
-  }
-  return null;
-};
```
NickGerleman, line 90: I removed this whole section because it is 100% not kosher for a render function to hav e side effects like this.

docs/virtualizedlist.md
```diff
@@ -92,9 +168,9 @@ Inherits [ScrollView Props](scrollview.md#props).
 
 ---
 
-### <div class="label required basic">Required</div> **`data`**
+### `data`
 
-The default accessor functions assume this is an array of objects with shape `{key: string}` but you can override `getItem`, `getItemCount`, and `keyExtractor` to handle any type of index-based data.
+Opaque data type passed to `getItem` and `getItemCount` to retrieve items.
```
NickGerleman, line 173: There isn't a default implementation of `getItem`, and `data` is optional.

PR #3479, opened 2022-12-14
Update using-a-listview.md

anand-codes:
Update keyExtractor to use a valid key


NickGerleman:
https://github.com/facebook/react-native-website/pull/3474/files makes the same update as well.

anand-codes:
Duped in https://github.com/facebook/react-native-website/pull/3474/files

PR #3483, opened 2022-12-15
Make C++ TM example work with C++17

christophpurrer:
The example given uses C++20 syntax which MSVC does not support yet.
It works in clang with some special compiler flags. 
Updated sample also works in C++17 with MSVC.
Verified that example compiles in Xcode


NickGerleman (approved):
LGTM, though MSVC has supported designated initializers in C++20 mode for a while now (But clang/gcc would allow them in earlier language versions).

cortinico (approved):

PR #3486, opened 2022-12-22
Update Linting to RN 0.71.0-rc.5

NickGerleman:
Updates to a package with newer RN typings, which no longer requires patching.


Simek (approved):
That's great news, LGTM! 👍

PR #3489, opened 2022-12-22
Remove `NativeDeprecated` from "Direct Manipulation" page

yungsters:
This removes the deprecation notice at the top of "Direct Manipulation".

For the foreseeable future, `setNativeProps` will continue providing value in high frequency, performance-sensitive scenarios such as animations. Until we have a reasonable alternative, there is no reason for the deprecation notice to be displayed on this page.

This is additionally influenced by our ongoing work to implement `setNativeProps` in the New Architecture.

Screenshot of the "Direct Manipulation" documentation page without the deprecation notice

### DIFF
--- a/docs/direct-manipulation.md
+++ b/docs/direct-manipulation.md
@@ -3,9 +3,7 @@ id: direct-manipulation
 title: Direct Manipulation
 ---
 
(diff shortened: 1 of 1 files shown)


Simek (approved):
LGTM! 👍

cipolleschi (approved):

PR #3490, opened 2022-12-27
fix: improve examples for Android Native Modules addListener/removeListeners

mdjastrzebski:
Improve Android Native Module examples for using `addListener`  and `removeListeners` with pattern that mimics the one in RCTEventEmitter on iOS. The current examples are not self explanatory, as `removeListeners` is called with not-obvious `count` parameter. I've sent 1 hr figuring out how all of these should work and want to save other readers from having to do the same research.


cipolleschi:
@mdjastrzebski Thank you so much for taking the time to add these details in all the versioned docs!

PR #3492, opened 2022-12-28
adding typing

guilhermexpc:
Updating the structure of passing parameters by "props" to the typescript's typing structure.

### DIFF
--- a/docs/props.md
+++ b/docs/props.md
@@ -36,7 +36,11 @@ Your own components can also use `props`. This lets you make a single component 
 import React from 'react';
 import {Text, View} from 'react-native';
 
-const Greeting = props => {
+type props= {
+  name: string;
+}
+  
+const Greeting = (props: props) => {
   return (
     <View style={{alignItems: 'center'}}>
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hello @guilhermexpc, this example has the extension set to JS manually, this is probably for the simplicity, since this is an introduction page.

Personally I would leave this example as-is, but if you want to proceed with the change, you will need to:
* remove the explicit extension defined for Snack
* rename type, so it uses te pascal case (`Props`)
* ensure that there are no other missing types or problem with code after switch to TS.

Additionally, we cannot merge the PR without the CLA signed.

NickGerleman:
Yeah, the tab below for Typescript does have this already, so the default example will have the type, but not the JS-specific one.

NickGerleman:
Though I did manage to name it "GreetingsProps" for "Greeting" (non-plural) which I should fix...

Simek:
Closing this one in favour of Nick change, also looks like the "Props" example have a correct type definition already:

Screenshot 2025-08-08 at 16 18 08

PR #3493, opened 2022-12-28
feat: Add inset logical properties to layout props

gabrieldonadel:
This PR updates the Layout Props documentation to include the following new `inset` logical properties: 

- `inset` 
- `insetBlock` 
- `insetBlockEnd` 
- `insetBlockStart` 
- `insetInline` 
- `insetInlineEnd` 
- `insetInlineStart` 

Related to https://github.com/facebook/react-native/commit/9669c10afceef65626c82149210afc07d47df98b


gabrieldonadel:
Hi @NickGerleman, who should I ping for a review on this?

NickGerleman:
Right now the "next" version of the documents targets React Native 0.71. These are cut as part of a react-native release, and after, "next" will target 0.72. Since this change targets 0.72, we need to wait until that cut before merging it.

Content-wise, mostly LGTM. Could we add the clarification that these only work on Fabric?

gabrieldonadel:
> Could we add the clarification that these only work on Fabric?

Sure, I completely missed that, should we have some sort of tag for `Fabric only` props? Just like we have for `Android and iOS`? e.g.
image

NickGerleman:
I think something like that would be great. Last time I asked around, I think these props and the ones for pointer events are the first ones that are Fabric only, so I'm not sure we've made anything yet.

Simek:
Nope, we do not have the dedicated label for Fabric yet, but it can quite easily created as a variant of platform tag if needed, as Gabriel suggested:
* https://github.com/facebook/react-native-website/blob/main/website/src/css/customTheme.scss#L1418

gabrieldonadel:
Sounds good, thanks for the feedback @NickGerleman and @Simek, I've just pushed a commit adding a `Fabric only` tag

NickGerleman:
https://deploy-preview-3493--react-native.netlify.app/docs/next/layout-props#insetblock-fabric

LGTM

image

NickGerleman (approved):

Simek (approved):
Thank you for the docs updates and bringing up the Fabric-only props discussion, LGTM! 👍

rickhanlonii (changes requested):
Let's discuss this a bit more, I think it's a bit confusing for users that they need to now know what architecture a prop works on, and that we're exposing codenames like Fabric in the API docs.

NickGerleman:
Cc @rickhanlonii, from offline discussion, can we go ahead and merge?

gabrieldonadel:
any updates on this? cc @NickGerleman

NickGerleman:
> any updates on this? cc @NickGerleman

This change ended up sparking a lot of discussion internally on how we should be framing the New Architecture and Fabric. I think there's a good concern here, about the majority user not yet being on Fabric, or more casual users not knowing what Fabric is. So UI-wise, I think the uncertainty is coming down to how we represent, and what prominence we want to give to Fabric-only props.

Should be able to share more on Monday. Apologies for this stretching out, I am really glad you made this change to add the documentation.

NickGerleman:
The latest update, we still don't have great agreement on how we should handle this case. So we're continuing to refine that (for future cases where we want to do something similar to this).

In this specific case, it sounded like there may also be a natural Paper implementation for aliasing layout props. @gabrieldonadel is that something you think you would be willing to take on for the 0.72 timeframe?

I do also want to apologize my role in this, advocating for switching from a JS-side shim to native Fabric implementation. Part of the reason I did that was knowing the Fabric only implementation was not much more expensive to build, and I still think it's likely the better alternative in that case, but it also landed the change into some unknown territory we're still figuring out.

gabrieldonadel:
No worries, thanks for the update @NickGerleman, I can definitely implement this on Paper within the 0.72 timeframe

PR #3498, opened 2023-01-03
Add @latest when creating library to ensure up-to-date template

satya164:
When running a command with `npx`, there's no guarantee that it'll use the latest version of the package. If the package is already cached, it will use the cached version instead (see https://github.com/npm/cli/issues/4108). This can result in users getting a project based on an outdated template which might be missing any fixes.

This change adds `@latest` to the command to ensure that the latest version is always used. We already do this in Bob's documentation: https://github.com/callstack/react-native-builder-bob


cipolleschi (approved):
@satya164 Thank you for this fix!

PR #3505, opened 2023-01-06
Update diffs in TypeScript path alias instructions

NickGerleman:
Noticed the diffs used here are out of date compared to the RN template (even the previous TS one). Update based on the current out-of-the-box files.

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -145,19 +145,21 @@ You can explore the syntax more in the [TypeScript playground][tsplay].
 
 ## Using Custom Path Aliases with TypeScript
 
-To use custom path aliases with TypeScript, you need to set the path aliases to work from both Babel and TypeScript. Here's how:
+To use custom path aliases with TypeScript, you need to set the path aliases to work from both Metro and TypeScript. Here's how:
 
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -145,19 +145,21 @@ You can explore the syntax more in the [TypeScript playground][tsplay].
 
 ## Using Custom Path Aliases with TypeScript
 
-To use custom path aliases with TypeScript, you need to set the path aliases to work from both Babel and TypeScript. Here's how:
+To use custom path aliases with TypeScript, you need to set the path aliases to work from both Metro and TypeScript. Here's how:
```
NickGerleman, line 148:
```suggestion
To use custom path aliases with TypeScript, you need to set the path aliases to work from both Babel and TypeScript. Here's how:
```

docs/typescript.md
```diff
@@ -183,7 +185,8 @@ yarn add --dev babel-plugin-module-resolver
 
 ```diff {3-13}
```
NickGerleman, line 186:
```suggestion
```diff
```

PR #3506, opened 2023-01-06
Add section on updating linting tools

NickGerleman:
Adds high level instructions to update the versions/config files used by `@react-native-website/lint-examples` as part of the release process.


README.md
```diff
@@ -112,6 +112,14 @@ Docusaurus keeps track of the list of versions for the site in the `website/vers
 
 #### Cutting a new version
 
+##### After RC
```
NickGerleman, line 115: Theoretically we can also add examples which target not-yet-RC versions using the same steps, but in practice Snack won't be able to use them yet.

PR #3508, opened 2023-01-06
default for removeClippedSubviews is True on Android and False on iOS…

RRaideRR:
…. The documentation did not reflect that (it said the default is always false)

### DIFF
--- a/docs/optimizing-flatlist-configuration.md
+++ b/docs/optimizing-flatlist-configuration.md
@@ -25,9 +25,9 @@ Here are a list of props that can help to improve `FlatList` performance:
 
 ### removeClippedSubviews
 
-| Type    | Default |
-| ------- | ------- |
-| Boolean | False   |
+| Type    | Default                     |
+| ------- |-----------------------------|
+| Boolean | False (iOS), True (Android) |
 
 If `true`, views that are outside of the viewport are detached from the native view hierarchy.
 


docs/optimizing-flatlist-configuration.md
```diff
@@ -25,9 +25,9 @@ Here are a list of props that can help to improve `FlatList` performance:
 
 ### removeClippedSubviews
 
-| Type    | Default |
-| ------- | ------- |
-| Boolean | False   |
+| Type    | Default                     |
+| ------- | --------------------------- |
+| Boolean | False (iOS), True (Android) |
```
Simek, line 30:
```suggestion
| Type    | Default                              |
| ------- | ------------------------------------ |
| Boolean | `true` on Android, otherwise `false` |
```

Let's make this description more inline with the current code:
* https://github.com/facebook/react-native/blob/main/Libraries/Lists/FlatList.js#L158

Simek (commented):
Hello @RRaideRR, thank you for catching this! With the suggested changes applied it should be ready to ship.

It would be nice to also update component docs, not only the guide, but this can be done in the separate PR, or just someone else could follow on that.

Screenshot 2023-01-07 at 10 15 00

Simek:
Hey @RRaideRR, thanks for pointing this out! 🙏

I have tried to merge it in, but looks like the CI checks got stuck, so I had to make an another PR with the fixes. So closing this one.

PR #3517, opened 2023-01-12
Update getting-started.md

aeorge:
Starting with 0.71, TypeScript is used by default and instead of an `App.js` file an `App.tsx` file gets created.


Simek (commented):

Can you also update the versioned file for 0.71, so the change will be visible in the latest docs?
* https://github.com/facebook/react-native-website/blob/main/website/versioned_docs/version-0.71/getting-started.md

Simek (approved):
Thank you for the change backport, LGTM! 🚀

PR #3529, opened 2023-01-23
Fix bundle install on M1

SamiKammoun:
On M1, bundle install will produce an error that could be mitigated by specifying the architecture (arm64)
The solution was found [here](https://github.com/CocoaPods/CocoaPods/issues/11641#issuecomment-1323173766)

## How to reproduce:

following the tutorial step by step and running `npx react-native init AwesomeProject` (installing with gem or brew produces the same error)

### DIFF
--- a/docs/_getting-started-macos-ios.md
+++ b/docs/_getting-started-macos-ios.md
@@ -90,10 +90,16 @@ This is not necessary if you are integrating React Native into an existing appli
 
 If you are having trouble with iOS, try to reinstall the dependencies by running:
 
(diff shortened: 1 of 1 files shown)


Simek:
Hey @SamiKammoun, since the dependencies and package architecture for Apple Silicon has been stabilized, I think that those additional notes are no longer required. 

Tested this on fresh macOS installation lately and there were no hiccups. Let's create a new issue or PR if this persist for you for some reason, happy to help you debug that!

PR #3535, opened 2023-01-25
Adding an info box on NVM and zsh to MacOS + iOS

kraenhansen:
For users of NVM ([a very popular Node.js version manager](https://2022.stateofjs.com/en-US/other-tools/#utilities)) and zsh (which is now the default shell on macOS), I want to add an info box to help workaround a common issue of the current React Native template, to help Xcode resolve the correct path for Node.js when invoking shell scripts.

Screenshot 2023-01-25 at 14 21 17

On a related note, I've commented on an issue in the NVM repository to update the install.sh script to have this happen automatically for new users: https://github.com/nvm-sh/nvm/issues/338#issuecomment-1403559575


cipolleschi (approved):
Looks good to me. I'm not an expert of NVM, but adding those lines in the doc won't hurt and may help people!

PR #3536, opened 2023-01-25
docs: update correct usage of LayoutAnimation.create

hslee2008:
# Update

A long time ago, I had opened a pull request about a bug in `LayoutAnimation.create()`.
Upon inspecting it a bit more, I found out it was not a bug.
The properties were not set and the documentation had no table to explain the parameters requirements.

https://github.com/facebook/react-native/issues/33740

The issue can be closed now.

# Changes to the documentation

1. Rename `creationProp` to `property`. Since the original function's parameter's name is `property`
2. Add table parameters with all of the parameters set to required w
3. Update Snack Example with updated API Example (both function and dclass example)

The problem was the example and the table which was inaccurate. Not sure why but even without some of the parameters, iOS worked and android didn't. But with all of the parameters defined in the example, it works in both platforms.

### DIFF
--- a/docs/layoutanimation.md
+++ b/docs/layoutanimation.md
@@ -237,12 +237,9 @@ class App extends Component {
   };
 
   toggleBox = () => {
-    LayoutAnimation.configureNext({
-      duration: 500,
-      create: {type: 'linear', property: 'opacity'},
-      update: {type: 'spring', springDamping: 0.4},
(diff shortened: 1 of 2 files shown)


HyunseungLee-Travis:
@Simek Please have a look at the changes. I think it the changes to the documentation is important because the `create()` api doesn't even work in the example and doesn't work even when following the documentation.

PR #3544, opened 2023-01-29
Fix "tsconfig.json" and "babel.config.js" example code

kadiraydinli:
I noticed that `compilerOptions` is not used in the `tsconfig.json` sample code.
It should be like here; https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

I also made a minor appearance fix for the `babel.config.js` file.


NickGerleman:
LGTM 👍

PR #3545, opened 2023-01-30
[RFC] Add New Arch version docs

rickhanlonii:
## Overview

I'm proposing a change to the New Arch docs, which moves all of the New Arch docs to a separate version in the docs. In this PR, I'll explain the benefits I think this brings, some common concerns about the move, and how I think they're mitigated.

[Live Preview](https://deploy-preview-3545--react-native.netlify.app/docs/experimental/the-new-architecture/landing-page)

### Benefits

- **Consistent docs**: Currently, the docs for a version includes both the old arch and the new arch. This means for a given version, you have to a) know what architecture you're on and b) know which parts of the docs are relevant to the new architecture. As a result, we have a bunch of references all over the docs to the new architecture, which is also confusing because users should not be opting into the new arch yet except to test and provide feedback.
- **Clearer expectations**: We're hearing feedback that users don't understand our expectations for the new arch (i.e. should they upgrade now or not). My thesis is that the docs are a part of this - the docs are documented as if it's landed in the stable channel in available to use, when in reality, for all non-testing users, it's going to land in a future version. 
- **Fabric-only props**: There is a proposal to add some fabric-only props, and that brings up the issue of how to document this. One solution is to add inline-flags in the docs to tell users the props are fabric only. My thesis is that this exasperates the above two problems. With this change, we could document those props only in the New Architecture version, since they're only available in the New Arch.
- **Simplifying the docs**: After this change, we can start finalizing what the docs will look like in a post-new-arch world. For example, instead of having a docs page for Native Modules, and Turbo Modules, we can update the Native Modules doc to explain how to implement Turbo Modules only (referencing the legacy modules where needed).
- **Un-versioning the New Arch docs**: See the "concerns" section below, but I see this as a benefit - this change un-versions the docs so that there is one version of the new architecture, the latest version, and we are free to update and change the version/docs during this phase. This sets the expectation that to test the new arch and give us feedback, you need to update to the lastest version before testing.

Moving the top-level "architecture" section: finally, I moved the Architecture top-level nav item to the New Architecture version. This page only documents the new architecture, so until that lands, it's not accurate or relevant to most users/

### Concerns

Please let me know what your concerns are, here are some that I have heard:
- **Loss of versioning for the New Arch**: As mentioned above, I see this as a benefit. With versioning, users on older versions would have clear instructions for how to upgrade to the New Architecture, and we're losing that. But the question is: Why should users use the New Arch on older versions? Right now, we only want users to enable the New Arch to test it and give us feedback. Although it's possible that we could get useful feedback on old versions, more often then not, issues users experience will be fixes in later versions anyway. Note: from our website data, essentially 0 users are currently visiting old versions of the New Arch docs.
- **Discoverability**: there is some concern that this buries the New Arch docs so users can't find it. There's some truth to this, and although I think that the power users who we know are testing the New Arch will find the docs, maybe there is something we can do to make these docs more discoverable without disrupting the experience of our old arch users.
- **Harder to change the docs**: Since this change forks the docs, there is some additional overhead in maintaining the New Arch docs. For example, we no longer can use the "next" channel to "stage" New Arch docs (so we may need to feature branch them until the next release).
- More? Let me know!

Overall, I think this is making a tradeoff. We lose versioning and add a little more work to maintaining the docs, but we get a much simpler story, with clearer expectations, and a lot more freedom to change the docs for the New Architecture.


NickGerleman:
Instructions may and have changed per-version. Moving all of the new architecture out of version management breaks that, along with any of the benefits we get from being able to document different behavior on different versions.

I am strongly against breaking versioning for anything "experimental", especially since this is public, opt-in, and people to frequently use and make edits to this documentation.

If we want to downplay the new architecture compared to current prominence, I think that is fine, but let's please not break versioning to do that.

cortinico (approved):
Thanks for looking into this @rickhanlonii 
This will make it easier for folks to distinguish between new arch and old arch 👍 Let's iterate on the setup if we find some rough edges

lunaleaps:
Not a blocker but these sidebar categories should probably be clarified -- the titles are kind of confusing between what the difference is
CleanShot 2023-02-21 at 09 26 49@2x

Simek:
In general, this is a good idea, but in the other hand I feel like we introduce another level of docs complexity, which even before, with versioned docs and "Next" could confuse people, but with another select it might be even more problematic for some users. So I think that we should focus on making the distinction between both docs versions and ensure that we do not list twice content or navigation unless it's beneficial, to keep the nav structure clear and sidebar streamlined.

Even now, I'm sure why "Stable.Next" docs include the Architecture sections (but in unusual order) and they to not differ from "Experimental.Next", but this also might be a mistake:
Screenshot 2023-02-22 081814

Additionally docs for latest releases docs differs in the section order to the both "Next" versions, which also seems like an unintended change.
Screenshot 2023-02-22 082938

Maybe for the clarity, and distinction, we should include the permanent banner while browsing "Experimental" docs, similar to the one, which pop out in "Next" docs:
Screenshot 2023-02-22 082657

kelset:
I agree with the general feedback that has been given so far:
* we should really not lose versioning (that seems be have been addressed, at least somewhat? but the "all other versions" doesn't link to a dedicated versions page for new arch? see screenshot)
Screenshot 2023-02-22 at 11 25 20

* I think fully duplicating the docs is gonna lead to massive pains for contributors, we are basically in a similar situation of a long living branch were changes need to be manually done twice every time. We are increasing the risk of the docs misaligning (ex. fixing a typo in the alert component page needs to be done twice, and since there's no way to automatically replicate either the reviewer remembers to tell the person to do it in both places, or we'll end up with things misaligned) (so, it's bigger than just "additional overhead in maintaining the New Arch docs") (as Simek and Luna have shown in the screenshot, it's very easy to end in an unsynced state)
* the versions shown in the submenus should always match [the support window](https://github.com/reactwg/react-native-releases#which-versions-are-currently-supported) IMO
* I really don't think we should let props be fabric-only - the user shouldn't even have to ask themself if a feature they want is available on their arch: everything should work, the change should be transparent for the avg user
* this solution increases docs complexity for the user
* I agree with the sentiment of wanting to set clearer expectations and I think it's going to benefit people a lot to have clarity in that sense

I'm passing this PR to more folks at MSFT so that they can provide their own thoughts

acoates-ms:
I'm glad versioning came back for the new docs.

> I really don't think we should let props be fabric-only - the user shouldn't even have to ask themself if a feature they want is available on their arch: everything should work, the change should be transparent for the avg user

Unfortunately, that is already a reality.  And at this point new features in core are often going to only get implemented on fabric unless it is trivial to add it to paper too.  Otherwise, you are often going to double the amount of work to add a new feature, which just slows everything down.

I look forward to the "simplifying the docs" next step, which I think is really where this plan will shine.  Showing users the "Native Modules" and "Native Components" sections currently is just confusing new users.  The only place those should be mentioned is under the back-compat section.

kelset:
> Unfortunately, that is already a reality. And at this point new features in core are often going to only get implemented on fabric unless it is trivial to add it to paper too. Otherwise, you are often going to double the amount of work to add a new feature, which just slows everything down.

right - it sounds like this is an architectural conversation on how the new arch should have been structured/delivered/planned more than a feedback to give to the docs, so I'll retire that bullet point from my feedback

NickGerleman:
> > Unfortunately, that is already a reality. And at this point new features in core are often going to only get implemented on fabric unless it is trivial to add it to paper too. Otherwise, you are often going to double the amount of work to add a new feature, which just slows everything down.
> 
> right - it sounds like this is an architectural conversation on how the new arch should have been structured/delivered/planned more than a feedback to give to the docs, so I'll retire that bullet point from my feedback

This has been contentious, and there are Fabric only props introduced in 0.72 that are (if all goes well) also getting a paper version because of it.

There will be new Fabric only properties (e.g. pointer events) and components (e.g. StrictLayout), but how these will be staged/released is not certain yet.

I do personally tend to still think it's the best use of engineer-time to ship Fabric features when ready (add to typings and documents with some disclaimer) but it does create confusion.

NickGerleman:
I must confess I looked through the site preview, and not the source changes, and missed the amount of duplication here. I also agree that it will be an issue. Apart from effort, leading to documents getting out of sync.

As a more extreme example, it means that, per API, we may now have up to 8 of the same code example needing updates for a change. (JS/TS, Class/Function component, Paper vs Fabric, = 2^3). Though we should be able to remove the class component examples soon (the new React docs heavily reduce visibility to them).

rickhanlonii:
The blending of docs is super confusing but based on feedback I’m not going to merge this so
I’m going to close.

PR #3552, opened 2023-02-05
change the use of incorrect word

ahmadaliyu:


Simek:
Hello @ahmadaliyu, nice catch, thank you for the correction! 👍 

Can you also update the `unversioned` doc page, which is a source for newest version, so the change is retained in newer docs?
* https://github.com/facebook/react-native-website/blob/main/docs/native-modules-android.md

ahmadaliyu:
Yes it is now handled accordingly

Simek (approved):
Thank you for the updates, LGTM.

PR #3571, opened 2023-02-15
Correct spelling: exisiting -> existing

blakef:
This looks like a historical artifact.


Simek (approved):
Nice catch, thank you for applying the correction back all the way, including the oldest docs! 👍

PR #3575, opened 2023-02-17
Fix Flow/TypeScript Mismatches in New Architecture Library Intro

### LINKED ISSUE #3515: Fix requireNativeComponent() and codegenNativeCommand() examples post-TS
https://reactnative.dev/docs/next/new-architecture-library-intro#migrating-off-dispatchviewmanagercommand

The document describes doing codegen using Flow, but I changed the example filenames to TypeScript. The content needs some TS specific updates in general, but we should either do that or just revert the filenames back to Flow.

NickGerleman:
Fixes #3515

This fixes a few cases where Flow-specific examples are represented as TypeScript. Some of these were present in the original guide, and others I made worse during my mass conversion of inline code examples to TypeScript.

I think some of these were made to be `.ts` to appease syntax highlighting/formatting, but it looks like adding `@flow` is enough to inform both prettier over markdown, and prism syntax highlighting to render correctly. We also change some `.ts` to `.tsx` when components are used.

image

### DIFF
--- a/docs/new-architecture-library-intro.md
+++ b/docs/new-architecture-library-intro.md
@@ -397,12 +397,12 @@ export default RNTMyNativeViewNativeComponent;
 If `requireNativeComponent` is not typed, you can temporarily use the `mixed` type to fix the Flow warning, for example:
 
 ```js
-// @flow strict
+// @flow strict-local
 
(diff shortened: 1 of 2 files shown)


NickGerleman:
Prism is set to support Flow, so I expect inline Flow examples will be colorized correctly, but will check on the live site.

lunaleaps (approved):

PR #3589, opened 2023-02-23
Updating dead link for label-action workflow

leabs:
This fixes a dead link on `/contributing/bots-reference` in the `label-actions` section.


Simek (approved):

PR #3591, opened 2023-02-26
fix: add vertical line in project structure

antonkesy:
For me, there is no clear reason to not keep the same style for the complete example project structure
Therefore, I added the missing vertical line


Simek (approved):
Thank you for the correction! 👍

PR #3593, opened 2023-02-27
Update Flatlist to Allow ArrayLike data

NickGerleman:
Committed with https://github.com/facebook/react-native/commit/c03de97fb44f0aecbec8f930bd99fe26f37f9648.

This is technically probably going to be backported to 0.71 as well, but there is a breaking bit of the change coming in 0.72, and not all 0.71 versions will permit this, so it probably doesn't make sense in 0.71 versioned docs.

### DIFF
--- a/docs/flatlist.md
+++ b/docs/flatlist.md
@@ -426,11 +426,11 @@ Example usage:
 
 ### <div class="label required basic">Required</div> **`data`**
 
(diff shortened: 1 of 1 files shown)


Simek (approved):
Thank you for the context on this change, LGTM! 👍

PR #3595, opened 2023-02-27
fix: Cocoapods -> CocoaPods

blakef:
Also added it to antfu/case-police#148 so our tooling will fix this in the future.

### DIFF
--- a/website/react-native-dict.json
+++ b/website/react-native-dict.json
@@ -1,0 +1,3 @@
+{
+  "Cocoapods": "CocoaPods"
+}
--- a/website/versioned_docs/version-0.60/navigation.md
+++ b/website/versioned_docs/version-0.60/navigation.md
@@ -35,7 +35,7 @@ The second step is to install the required peer dependencies. You need to run di
   npm install react-native-reanimated react-native-gesture-handler react-native-screens react-native-safe-area-context @react-native-community/masked-view
   ```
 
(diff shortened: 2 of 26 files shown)


Simek (approved):
LGTM. Nice idea to propose that upstream. 👍 

Btw., we can also define the custom dict which can extend the included in `case-police` checks. We do this in Expo repo currently:
* https://github.com/expo/expo/blob/main/docs/expo-dict.json

blakef:
> LGTM. Nice idea to propose that upstream. 👍
> 
> Btw., we can also define the custom dict for more niche terms which can extend the included in `case-police` checks. We do this in Expo repo currently:
> 
> * https://github.com/expo/expo/blob/main/docs/expo-dict.json

That's a great callout, so I'll add it here.  I'd anticipate this isn't the last time we'll hit something similar.

Simek:
You also need to pass the dict path via param in the command:
* https://github.com/expo/expo/blob/main/docs/package.json#L15

It is also good to verify that case-police works as expected, IIRC the order of prams matters, for example, for rules overwrites.

cortinico (approved):

blakef:
> You also need to pass the dict path via param in the command:
> 
> * https://github.com/expo/expo/blob/main/docs/package.json#L15
> 
> It is also good to verify that case-police works as expected, IIRC the order of prams matters, for example, for rules overwrites.

Thanks for pointing this out @Simek , I was in too much of a hurry and didn't commit my `package.json` changes.  I've thrown those up here: #3597.

PR #3596, opened 2023-02-27
docs: remove references to Ruby Version Management

blakef:
We now support Ruby versions >= 6.8.10, <= 3.2.0 (facebook/react-native#36281). This reduces some of the noise in our docs by removing error prone steps on setting up a Ruby Version Manager.


cipolleschi (approved):
Nice! I like that: more streamlined and less confusion for our users!

PR #3618, opened 2023-03-10
chore: upgrade docusaurus to 2.3.1 & react 18.2

blakef:
Upgrade these dependencies, so we can support deep-linking in [Tabs](https://deploy-preview-8225--docusaurus-2.netlify.app/docs/markdown-features/tabs/?current-os=ios#query-string).

### DIFF
--- a/website/src/pages/showcase.js
+++ b/website/src/pages/showcase.js
@@ -14,13 +14,15 @@ import ThemedImage from '@theme/ThemedImage';
 import {Section} from './index';
 import IconExternalLink from '../theme/Icon/ExternalLink';
 
-const renderApp = (app, i) => {
-  const imgSource = !app.icon.startsWith('http')
-    ? useBaseUrl('img/showcase/' + app.icon)
-    : app.icon;
+const renderApp = (app, i) => <AppBox app={app} key={`app-${app.name}-${i}`} />;
+
+const AppBox = ({app}) => {
+  const imgSource = useBaseUrl(
+    app.icon.startsWith('http') ? app.icon : 'img/showcase/' + app.icon
+  );
 
   return (
-    <div className="showcase" key={`app-${app.name}-${i}`}>
+    <div className="showcase">
       <div className="iconBox">
         <img src={imgSource} alt={app.name} className="iconBackground" />
         <img src={imgSource} alt={app.name} className="icon" />
@@ -71,7 +73,7 @@ const renderLinks = app => {
 };
 
 const randomizeApps = apps =>
-  apps.filter(app => !app.group).sort(() => 0.5 - Math.random());
+  [...apps].filter(app => !app.group).sort(() => 0.5 - Math.random());
 
 const Showcase = () => {
   const {siteConfig} = useDocusaurusContext();
--- a/website/package.json
+++ b/website/package.json
@@ -45,18 +45,18 @@
     ]
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.1",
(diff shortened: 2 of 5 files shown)


Simek:
Thanks for looking at Docusaurus upgrade! 👍 

There are several thing which need to be done,, beside bumping the version in `package.json`:
* we need to regenerate swizzled components, and re-apply our custom modifications, files are located in https://github.com/facebook/react-native-website/tree/main/website/src/theme, and there is a guide how to do this - https://docusaurus.io/docs/next/swizzling
* we need to verify if custom classes used in our SCSS files still exist, there is no other way than going through various pages and looking for regressions in comparison to the currently deployed PROD instance

Broken thing I have spotted on quick check:
* loading showcase page crashes Docusuarus - https://deploy-preview-3618--react-native.netlify.app/showcase
* announcement bar (at top) looks unstyled now

blakef:
## Push for the Showcase issue coming soon.

I wrote some visual diff tooling to give me a better chance of catching these things.
CleanShot 2023-03-13 at 20 47 06@2x

blakef:
@Simek I think this LGTM now.  When you have a moment would you mind looking over it again.  I've incorporated the swizzle changes and fixed the issue with showcase.

website/src/pages/showcase.js
```diff
@@ -77,13 +77,13 @@ const Showcase = () => {
...
-  useEffect(() => {
-    setRandomizedApps(randomizeApps(others.filter(app => !app.pinned)));
-    setPinnedRandomizedApps(randomizeApps(others.filter(app => app.pinned)));
-  }, []);
+  const [pinnedRandomizedApps] = useState(
+    randomizeApps(others.filter(app => !app.pinned))
+  );
+  const [randomizedApps] = useState(
+    randomizeApps(others.filter(app => app.pinned))
+  );
```
Simek, line 86: Removing `useEffect` wrap for entries randomization, leads to the rendering issue, where app labels do not match descriptions. This was written that way to address those issues.

 
Screenshot 2023-03-14 090331

Simek:
Thank you for all the work put into the upgrade! 🙏 Besides the comment about Showcase issue, this LGTM! 👍 

Also, it looks like we have broken content container on versions page - https://deploy-preview-3618--react-native.netlify.app/versions - but surprisingly, this already seems to be an issue for PROD deployment, so we can address this in a separate PR.

blakef:
I'm spamming netifly atm because the issue doesn't reproduce on my locally hosted version.  Apologies for the noise.

website/src/pages/showcase.js
```diff
@@ -14,13 +14,15 @@ import ThemedImage from '@theme/ThemedImage';
...
 import IconExternalLink from '../theme/Icon/ExternalLink';
 
-const renderApp = (app, i) => {
-  const imgSource = !app.icon.startsWith('http')
-    ? useBaseUrl('img/showcase/' + app.icon)
-    : app.icon;
+const renderApp = (app, i) => <AppBox app={app} key={`app-${app.name}-${i}`} />;
+
+const AppBox = ({app}) => {
+  const imgSource = useBaseUrl(
```
blakef, line 20: This was necessary to keep the # of hook calls consistent.  [`useBaseUrl` handles FQDNs](https://github.com/facebook/docusaurus/blob/2f75979bc5b4fd971225098e1c122efdd71d6e6f/packages/docusaurus/src/client/exports/__tests__/useBaseUrl.test.tsx#L41).

PR #3619, opened 2023-03-10
feat: tabs now support deep-linking

blakef:
Adds support for deep-linking in tabs.  This will only work *after* #3618 goes it (note: it won't break anything if it does land before).

I updated all of these using this:
```bash
git grep -lE '<Tabs .*groupId' | xargs -J echo gsed -ri 's/<Tabs(.*)groupId="([^\"]*)"/<Tabs\1groupId="\2" queryString/'
```

https://user-images.githubusercontent.com/49578/224362123-8126614d-71e5-4a6f-a219-511d0ed73160.mp4

### DIFF
--- a/docs/_integration-with-existing-apps-objc.md
+++ b/docs/_integration-with-existing-apps-objc.md
@@ -39,7 +39,7 @@ Next, make sure you have [installed the yarn package manager](https://yarnpkg.co
 
 Install the `react` and `react-native` packages. Open a terminal or command prompt, then navigate to the directory with your `package.json` file and run:
 
-<Tabs groupId="package-manager" defaultValue={constants.defaultPackageManager} values={constants.packageManagers}>
+<Tabs groupId="package-manager" queryString defaultValue={constants.defaultPackageManager} values={constants.packageManagers}>
 <TabItem value="npm">
 
 ```shell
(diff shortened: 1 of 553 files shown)


motiz88:
1. I don't think we generally update old versioned snapshots of the docs unless necessary. Reverting those changes can help with the size of this diff.
2. This doesn't fix *existing* inbound links that only include fragment references (without the new query params), right?

blakef:
Answering out of order @motiz88:
> This doesn't fix existing inbound links that only include fragment references (without the new query params), right?

No it doesn't fix that, just enables updating the query string when interacting (or the page if directly linked with the query string parameters set).  Interestingly there's some other work that I have to follow up with here to scroll after load & update the tabs (some of the `id`s aren't visible on load.  That'd also be a wider docusaurus contrib (but we can swizzle it for now).

FYI our CLI has a draft PR using this react-native-community/cli#1870

> I don't think we generally update old versioned snapshots of the docs unless necessary. Reverting those changes can help with the size of this diff.

The change is automated and pretty low risk (IMHO).  The CLI change means we'd want to update all of our [supported versions](https://github.com/reactwg/react-native-releases#releases-support-policy) (0.71, 0.70, 0.69).

motiz88 (approved):
lgtm, assuming you've done a spot check that the old versions still work after this change.

PR #3621, opened 2023-03-11

ThakurBallary:


cortinico (approved):
Thanks for sending this. Coudl you update also the docs for 0.71 which are also affected?

cortinico (approved):

PR #3626, opened 2023-03-16
feat: jumps to anchor in tabbed page

blakef:
We use tabs to nest whole pages. This means deep links might not aleady have the anchor referenced id on the page. For example:

>  https://reactnative.dev/docs/environment-setup#ruby

Only applies if the 'React Native CLI Quickstart' -> macOS -> iOS tabs have been selected. There's no guarantee this is the case until the document has loaded and the application have updated the DOM.  Once a bunch of PRs have landed (), this type of link will work as expected:

> https://reactnative.devdocs/environment-setup?os=macos&platform=ios&guide=native&package-manager=npm#rubys-bundler

This jump-to behaviour should only run on the initial load.

## Dependencies
This can go in before, it'll just be a noop until it lands.
- #3619 

## Demo

https://user-images.githubusercontent.com/49578/225635783-174c382c-8b2e-4813-a908-9dae98d8e8f7.mp4


website/modules/jumpToFragment.js
```diff
@@ -0,0 +1,20 @@
...
+export default (() => {
+  const isReady =
+    ExecutionEnvironment.canUseDOM && ExecutionEnvironment.canUseEventListeners;
+  if (isReady) {
+    window.addEventListener('load', () => {
+      if (!window.location.hash) {
+        return;
+      }
+
+      const ref = document.getElementById(window.location.hash.slice(1));
```
motiz88, line 12: As long as we're implementing this functionality in JS (boo! that we need to do that), what would it take to also support [Text Fragments](https://web.dev/text-fragments/)? There's an official [polyfill](https://github.com/GoogleChromeLabs/text-fragments-polyfill) from Chrome we might be able to reuse.

motiz88 (approved):

motiz88:
> There's no guarantee this is the case until the document has loaded and the application have updated the DOM. 

We should *probably* migrate to a Next.js server-rendered site at some point...

website/modules/jumpToFragment.js
motiz88: Also, is there a way we can retrigger the browser's native implementation (by re-navigating to the current location?) instead of emulating it in JS?

blakef: This is a really nice idea.  I'll add this as a followup idea worth investigating.

motiz88:
@Simek Just to clarify: I'm aware that the site currently uses Docusaurus. I'm saying we might want to consider moving to a different stack.

Simek:
@motiz88 It would be a quite a big and complex task, probably even bigger than migration to V2 (which took few months and leverage the dedicated MHL group help), so not sure if it is worth the effort if only SSR is a gain, and we will need to reimplement a lot of features like versioning, MDX handling, remark plugins etc., but we definitely can discuss that.

PR #3634, opened 2023-03-21
Fix documentation links to source code

1mike12:
I originally found one broken link in the docs and made a quick patch, but @cortinico mentioned it would be nice to backport to older versions of the documentation https://github.com/facebook/react-native-website/pull/3633 . That's when I realized there were a _lot_ of broken links. Most of them have to do with the pattern of the folders moving into the `packages/react-native/` subdirectory. 

For instance

`https://github.com/facebook/react-native/blob/master/React.podspec`
to
`https://github.com/facebook/react-native/blob/main/packages/react-native/React.podspec`

The `master`/`main` url is auto forwarded by github, but while fixing the links I figured it would make sense to just rewrite any masters to main

All the changes were done via regex replace everywhere. I spot tested various links before writing the various regex's and tried to write the patterns to be fairly targeted. I then randomly sampled some links manually to make sure nothing out of the ordinary was going on. Although with so many links I guess there _could_ be a mistake somewhere. Inside the commit messages, I wrote down the exact patterns used

There are still some broken links after this sweep. An example is the following links that only exist in older documentation relating to the custom webviews. I cant figure out where this is ( or if already deprecated since 0.63+) and doesn't really exist anymore
https://github.com/facebook/react-native/blob/master/React/Views/RCTWebViewManager.m
https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/views/webview/ReactWebViewManager.java


website/versioned_docs/version-0.60/_integration-with-existing-apps-objc.md
```diff
@@ -280,7 +280,7 @@ First `import` the `RCTRootView` header.
 }
 ```
 
-> Note that `RCTRootView initWithURL` starts up a new JSC VM. To save resources and simplify the communication between RN views in different parts of your native app, you can have multiple views powered by React Native that are associated with a single JS runtime. To do that, instead of using `[RCTRootView alloc] initWithURL`, use [`RCTBridge initWithBundleURL`](https://github.com/facebook/react-native/blob/master/React/Base/RCTBridge.h#L93) to create a bridge and then use `RCTRootView initWithBridge`.
+> Note that `RCTRootView initWithURL` starts up a new JSC VM. To save resources and simplify the communication between RN views in different parts of your native app, you can have multiple views powered by React Native that are associated with a single JS runtime. To do that, instead of using `[RCTRootView alloc] initWithURL`, use [`RCTBridge initWithBundleURL`](https://github.com/facebook/react-native/blob/main/packages/react-native/React/Base/RCTBridge.h#L93) to create a bridge and then use `RCTRootView initWithBridge`.
```
NickGerleman, line 283: For all of these versioned documents, maybe we should keep the existing organization, but point to e.g. `0.60-stable` instead of master/main?

NickGerleman (approved):

website/versioned_docs/version-0.60/_integration-with-existing-apps-objc.md
1mike12: yeah that makes the most sense, but it would be a lot harder and require a bit manual intervention. I don't think I would have the time to do that right now unfortunately, but what I could do is at least get the current docs pointed at the right place and save the older ones for later?

Simek (changes requested):
We were discussing replacing those link with @cortinico in one of the other PRs, and we decided that in versioned docs we should use release branches instead of `main`, for example `stable-0.60`.

NickGerleman:
> > We were discussing replacing those link with @cortinico in one of the other PRs, and we decided that in versioned docs we should use release branches instead of `main`, for example `stable-0.60`.
> 
> yup this would be the preferred solution

@1mike12 had mentioned above he may not have the time to contribute this. https://github.com/facebook/react-native-website/pull/3634#discussion_r1145528721

Maybe we should merge this, to bring back parity to the old state, than iterate if we want to maybe script something to do this?

Simek:
> Maybe we should merge this, to bring back parity to the old state, then iterate if we want to maybe script something to do this?

The problem with this approach is that this PR has changed not only the branch, but also path of old, versioned links paths, which would make switching to the release branch in the future more difficult for the person who want to tackle that.

Saying that, I can prepare a PR which addresses all of the problematic links in the way as we want this weekend, so we can solve issues related to the repository structure changes, nad move forward on that.

Simek:
Hey @1mike12, recently we have introduced the helpers for repository link based on the current docs version which user browse:
* https://github.com/facebook/react-native-website/pull/4817

Would you be interested in updating the PR using them?

Simek:
Hey @1mike12, looks like there is a lot of conflicts and it would be a bit of messy task to resolve them after the changes mentioned in the comment above.

If you still want to help us alter the link please create a new PR using the mentioned utils instead a hardcoded links per version.

PR #3641, opened 2023-03-23
Fix missing typescript template information

ssysm:


Simek:
Hello @ssysm, leaving the template parameter empty is desired and expected, since then user can select template to use from the list, where there are two options for TS project:
Screenshot 2023-03-23 104736

ssysm:
Ah I see, thanks for the info!

PR #3661, opened 2023-03-30
fix broken links in versioned docs for 71-66 releases

Simek:
# Why

It's a continuation of efforts to get rid of broken links started in #3660

# How

This PR updates all of the `main/master` links used in versioned documentation for releases from 0.66 to 0.71, with a links pointing to the specific release branches.

It would be nice to establish a pattern, a do the links replacement after cutting a new docs version.


cortinico (approved):
Same as the other pr, I've skimmed through the diff and it looks good to me 👍

If there is any meaningful change to review please let me know

PR #3662, opened 2023-03-30
fix broken links in versioned docs for 65-60 releases

Simek:
# Why

It's a continuation of efforts to get rid of broken links started in #3660

# How

This PR updates all of the `main/master` links used in versioned documentation for releases from 0.60 to 0.65, with a links pointing to the specific release branches.

There will be on more, small PR with various small fixes for the links I have spotted, but it need to be stacked on earlier PRs, to avoid messing up the versions split.


cortinico (approved):
I've skimmed through the diff and it looks good to me 👍

PR #3666, opened 2023-04-03
Update style.md

Badestrand:
Hi

I think the documentation about styling currently is simply wrong as it states that the same CSS as on the web can be used which is simply not true. I expanded a bit although it would be even better if there was an extensive documentation about styles and which CSS properties can be used in which way.


Badestrand:
I will change the wording a bit,.

PR #3676, opened 2023-04-08
Update pillars-fabric-components.md

hmhdworks:
android java class name(CenteredText.java) is CenteredText but in CenteredTextManager.java file we used as RTNCenteredText. it should be CenteredText. I am trying to create fabric component. it's getting build failed when I update the classname. it's working fine


hmhdworks:
@CLA SIGNED

PR #3680, opened 2023-04-11
feat: redirect current version urls

blakef:
Allows our CLI tool to user the latest version in the path, and gracefully handles this server side:

https://reactnative.dev/docs/0.71/X -> https://reactnative.dev/docs/X

## Why

Our CLI tooling will generate direct links based on OS, platform and version of React Native.  If you're running the latest version, our current docs don't support a version number infix.  For more details of why see [here](https://github.com/react-native-community/cli/pull/1902#issuecomment-1503094396) where the issue was originally raised.

Options that weren't better (IMHO):
- using the Docusaurus plugin-client-redirects → doesn't like non-existent paths (/0.71/) and would generate a bunch of placeholder html files with client side redirects.
- CLI checks the current version (failing which just default to top-level docs) → a bunch of version checks we shouldn't really need, but this was my fallback if we couldn't find a better way
- some kind of configuration in the `docusaurus.config.js` → I couldn't find anything that was baked in to handle this

## Testing
~~I'd need this to deploy to netlify to confirm, but I've run the `yarn build` locally and confirmed that correct changes are made to the `_redirects` file.~~

https://user-images.githubusercontent.com/49578/231149619-436671ce-dcee-4579-ade6-0e0102311c26.mp4

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -11,7 +11,7 @@
   },
   "scripts": {
     "start": "docusaurus start",
-    "build": "docusaurus build",
+    "build": "docusaurus build && yarn run update-redirect ./build/_redirects ./versions.json",
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
     "serve": "docusaurus serve",
@@ -58,6 +58,7 @@
   "devDependencies": {
     "@docusaurus/types": "^2.3.1",
     "@react-native-website/lint-examples": "0.0.0",
+    "@react-native-website/update-redirects": "0.0.0",
     "alex": "^10.0.0",
     "fs-extra": "^10.1.0",
     "glob": "^8.0.3",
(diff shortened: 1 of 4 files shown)


website/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "scripts": {
     "start": "docusaurus start",
-    "build": "docusaurus build",
+    "build": "docusaurus build && yarn run update-redirect ./build/_redirects ./versions.json",
```
NickGerleman, line 14: Does this work for “docusaurus start”?

NickGerleman (approved):
Makes sense to me. Thank you for the due diligence on alternatives to this.

website/package.json
blakef: Good question.  It doesn't work gracefully with `docusaurus start`.   I don't see a path to modify the webpack to support this.

I don't think this is blocking.

PR #3702, opened 2023-04-25
Update Debugging docs to no longer recommend Remote debugging

gabrieldonadel:
## Summary:

As a follow-up of https://github.com/facebook/react-native/pull/36754, which removes the remote debugging option from the dev menu, this PR updates the Debugging documentation, deprecating Remote debugging and recommending the usage of Direct debugging with Safari as an alternative. 

This also adds a quick section about Flipper as it currently is the hard-coded way of debugging when pressing `Open Debugger`

image


kelset:
sidenote: this should not be merged until the PR referenced in merged and ported to 0.72 (if it will).

huntie (changes requested):
Thanks @gabrieldonadel! It's high time we documented Flipper! My main suggestion is not to remove what we have already around Chrome DevTools (instead inserting this deprecation notice under this section), and to change the documented timeline to 0.73.

The React Native team is actively re-evaluating our debugging options ahead of enabling the New Architecture by default. We are looking at replacement direct debugging approaches, and Chrome DevTools should very much remain part of the picture. (Plans may change.)

As such, happy to approve this (after feedback), to merge after 0.72 is out. We're likely to come back to this page and do a second pass once our plans are solidified for 0.73.

huntie (approved):
Thanks! Approving, tied to the result of https://github.com/facebook/react-native/pull/36754 (still under evaluation).

This PR is held by the `Wait for future release` label.

PR #3705, opened 2023-04-28
[blog] Add Package Exports blog post

huntie:
Announcement blog post for [Package Exports support](https://github.com/react-native-community/discussions-and-proposals/blob/main/proposals/0534-metro-package-exports-support.md). This aims to complement the React Native 0.72 release post, given this is a larger and ecosystem-impacting feature. Communicates most important changes for app devs and package maintainers.

**To-dos**

- [x] Depends on incoming guide in the Metro docs — https://github.com/facebook/metro/commit/b2facd9e14c6e68b3b7bc1f4ca5bb0d90f897c8d
- [x] Date should be updated to align with the 0.72 release blog post.

Related to:

- https://github.com/facebook/metro/issues/670


website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+# Package Exports Support in React Native
+
+With the release of React Native 0.72, Metro — our JavaScript build tool — now includes experimental support for the `package.json` [`"exports"` field](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports). [When enabled](#enabling-package-exports-experimental), it adds the following functionality:
+
+- [React Native projects will work with more npm packages out-of-the-box](#for-app-developers)
+- [New capabilities for packages to define their API and target React Native](#for-package-maintainers-preview)
+- [Some breaking changes to package resolution (in edge cases)](#breaking-changes)
+
+In this post we'll cover how Package Exports works, and what these changes mean for you as a React Native app developer or package maintainer.
+
```
Simek, line 17:
```suggestion

<!--truncate-->

```

Small Docusaurus related nit, let's add truncate tag for example here, or somewhere else in the document, so we do not display full post content on the blog posts list.

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+date: 2023-05-01
+---
+
+# Package Exports Support in React Native
+
+With the release of React Native 0.72, Metro — our JavaScript build tool — now includes experimental support for the `package.json` [`"exports"` field](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports). [When enabled](#enabling-package-exports-experimental), it adds the following functionality:
+
+- [React Native projects will work with more npm packages out-of-the-box](#for-app-developers)
+- [New capabilities for packages to define their API and target React Native](#for-package-maintainers-preview)
+- [Some breaking changes to package resolution (in edge cases)](#breaking-changes)
```
Simek, line 14:
```suggestion
- [React Native projects will work with more npm packages out-of-the-box](/blog/2023/05/01/package-exports-support#for-app-developers)
- [New capabilities for packages to define their API and target React Native](/blog/2023/05/01/package-exports-support#for-package-maintainers-preview)
- [Some breaking changes to package resolution (in edge cases)](/blog/2023/05/01/package-exports-support#breaking-changes)
```

Oh, and those post links need to included full URL, so they lead to the post section when clicked from the list view.

Example from earlier post:

https://github.com/facebook/react-native-website/blob/aba3a883819a425cf2e6ced0d49557894e842232/website/blog/2023-01-03-typescript-first.md?plain=1#L12-L14

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+</details>
+
+### Breaking changes
+
+We decided on an implementation of Package Exports in Metro that is spec-compliant (necessitating some breaking changes), but backwards compatible otherwise (helping apps with existing imports to migrate gradually).
+
+The key breaking change is that when `"exports"` is provided by a package, it will be consulted first (before any other `package.json` fields) — and a matched subpath target will be used directly.
+
+- Metro will not expand [`sourceExts`](https://facebook.github.io/metro/docs/configuration/#sourceexts) against the import specifier.
+- Metro will not resolve [platform-specific extensions](/docs/platform-specific-code) against the target file..
```
Simek, line 144:
```suggestion
- Metro will not resolve [platform-specific extensions](/docs/platform-specific-code) against the target file.
```

Double dot.

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+
+For existing projects, we recommend that early adopters follow these steps to see if resolution changes occur after enabling `unstable_enablePackageExports`. This is a one-time process. It's likely that there will be no changes at all, but we'd like developers to opt in with certainty.
+
+<details>
+<summary>💡 Validating changes in your project</summary>
+
+1. Get all resolved dependencies (before changes):
+
+   ```sh
+   yarn metro get-dependencies index.js --platform android --output before.txt
```
brentvatne, line 115:
```suggestion
   # Replace index.js with your entry file if needed, such as App.js.
   yarn metro get-dependencies index.js --platform android --output before.txt
```

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+   ```
+
+   - **Expo only**: First, run `npx expo customize metro.config.js` (the generated file can be deleted later).
+   - **Expo only**: Substitute `index.js` for `App.js` (or an alternative entry file).
```
brentvatne, line 119:
```suggestion
```

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+```
+  "exports": {
+    "browser": "./dist/index-browser.js",
+    "react-native": "./dist/index-react-native.js",
+    "default": "./dist/index.js"
+  }
+```
+
+:::note
+We chose not to introduce `"android"` and `"ios"` conditions, due to the prevalence of other existing platform selection methods, and the complexity of how this behaviour might work across frameworks. Please use the [`Platform.select()`](https://reactnative.dev/docs/platform#select) API instead.
```
Simek, line 214:
```suggestion
We chose not to introduce `"android"` and `"ios"` conditions, due to the prevalence of other existing platform selection methods, and the complexity of how this behaviour might work across frameworks. Please use the [`Platform.select()`](/docs/platform#select) API instead.
```

WE can use internal link here.

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+<details>
+<summary>💡 Validating changes in your project</summary>
+
+1. Get all resolved dependencies (before changes):
+
+   ```sh
+   yarn metro get-dependencies index.js --platform android --output before.txt
+   ```
+
+   - **Expo only**: First, run `npx expo customize metro.config.js` (the generated file can be deleted later).
```
brentvatne, line 118:
```suggestion
   - **Expo CLI**: Run `npx expo customize metro.config.js` if your project doesn't have a custom Metro config yet.
```

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
+---
+title: 'Package Exports Support in React Native'
+authors: [huntie]
+tags: [announcement, metro]
+date: 2023-05-01
+---
+
+# Package Exports Support in React Native
+
+With the release of React Native 0.72, Metro — our JavaScript build tool — now includes experimental support for the `package.json` [`"exports"` field](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports). [When enabled](#enabling-package-exports-experimental), it adds the following functionality:
```
Simek, line 10:
```suggestion
With the release of React Native 0.72, Metro — our JavaScript build tool — now includes experimental support for the `package.json` [`"exports"` field](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports). [When enabled](/blog/2023/05/01/package-exports-support#enabling-package-exports-experimental), it adds the following functionality:
```

This link also need to use full URL, since it will be before the truncate tag.

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+- [`unstable_conditionsByPlatform`](https://facebook.github.io/metro/docs/configuration/#unstable_conditionsbyplatform-experimental) — The additional condition names to assert when resolving for a given platform target. By default, this matches `'browser'` when the platform is `'web'`.
+
+:::tip
+**Remember to use the React Native [Jest preset](https://github.com/facebook/react-native/blob/main/template/jest.config.js#L2)!** Jest includes support for Package Exports by default. In tests, you can override which `customExportConditions` are resolved using the [`testEnvironmentOptions`](https://jestjs.io/docs/configuration#testenvironmentoptions-object) option.
+:::
+
+#### Validating changes in your project
+
+For existing projects, we recommend that early adopters follow these steps to see if resolution changes occur after enabling `unstable_enablePackageExports`. This is a one-time process. It's likely that there will be no changes at all, but we'd like developers to opt in with certainty.
```
lunaleaps, line 107: What's the advice if there are changes?

website/blog/2023-05-01-package-exports-support.md
```diff
@@ -0,0 +1,229 @@
...
+## For app developers
+
+Package Exports can be enabled today, experimentally.
+
+- Imports against packages that depend on Package Exports features (such as [**Firebase**](https://www.npmjs.com/package/firebase) and [**Storybook**](https://www.npmjs.com/search?q=%40storybook)) should now work as designed.
+- React Native for Web projects using Metro will now be able to use the `"browser"` conditional export, removing the need for workarounds.
+
+Enabling Package Exports brings a few [edge-case breaking changes](#breaking-changes) that may affect specific projects, and which you can [test today](#validating-changes-in-your-project).
+
+**In a future React Native release, Package Exports will be enabled by default**. In a chicken-and-egg situation, React Native apps were previously a holdout for some packages to migrate to `"exports"` — or used our `"react-native"` root field escape hatch. Supporting these features in Metro will allow the ecosystem to move forward.
```
lunaleaps, line 81: What's the incentive for someone to enable Package Exports? Is it simply that we're going to make this default eventually and it's better to prepare earlier?

website/blog/2023-05-01-package-exports-support.md
huntie: The reason do this validation is for users to be certain that no resolutions have changed, particularly for large projects — it won't at this phase reveal any inaccessible imports.

e.g. Things like a `"browser"` conditional import now being resolved correctly. This might cause a subtle, but investigable change:

```diff
- node_modules/some-pkg/src/index.js
+ node_modules/some-pkg/dist/index.cjs
```

For brevity, I haven't included advice beyond "see if resolution changes occur", given the above output (should be both 🤞🏻) 1/ rare, and 2/ instructive.

website/blog/2023-05-01-package-exports-support.md
huntie: Mostly. It immediately enables the benefits presented on lines 76-77, but admittedly for most common packages there is already a happy React Native path or packages haven't moved to `"exports"` yet. (For cases where instead workarounds in apps were previously required, see https://github.com/facebook/metro/issues/670).

When some React Native libraries start migrating to `"exports"` later this year, they should be able to point back to this post to explain why RN 0.72/0.73 or above is required to consume a breaking package release that has switched to `"exports"`.

website/blog/2023-06-12-package-exports-support.md
```diff
@@ -0,0 +1,231 @@
...
+
+<details>
+<summary>💡 Validating changes in your project</summary>
+
+1. Get all resolved dependencies (before changes):
+
+   ```sh
+   # Replace index.js with your entry file if needed, such as App.js
+   yarn metro get-dependencies index.js --platform android --output before.txt
+   ```
```
Pranav-yadav, line 119: We should also provide the `npx` (or `npm *` equivalent) commands since not everyone is using `yarn` and we're suggesting creating the new RN projects using `npx` in the first place.
So, something like:

```sh
# using npx
npx metro get-dependencies index.js --platform android --output before.txt

# OR using yarn
yarn metro get-dependencies index.js --platform android --output before.txt
```

Or a note at the start about how to switch to `npx` commands would be great.

P.S.: I know some previous blog posts have been using `yarn` only, which also somewhat conflicts with rest of the decisions made for the docs.

Cc: @NickGerleman

website/blog/2023-06-12-package-exports-support.md
```diff
@@ -0,0 +1,231 @@
...
+
+This replaces the previous `"react-native"` root field. The priority order for how this was previously resolved was determined by projects, [which created ambiguity when using React Native for Web](https://github.com/expo/router/issues/37#issuecomment-1275925758). Under `"exports"`, _packages concretely define the resolution order for conditional entry points_ — removing this ambiguity.
+
+```
+  "exports": {
+    "browser": "./dist/index-browser.js",
+    "react-native": "./dist/index-react-native.js",
+    "default": "./dist/index.js"
+  }
+```
```
Pranav-yadav, line 213: Add language name for a better formatting 👍

website/blog/2023-06-14-package-exports-support.md
```diff
@@ -0,0 +1,235 @@
...
+};
+```
+
+Metro exposes two further resolver options which configure how conditional exports behave:
+
+- [`unstable_conditionNames`](https://facebook.github.io/metro/docs/configuration/#unstable_conditionnames-experimental) — The set of [condition names](https://nodejs.org/docs/latest-v19.x/api/packages.html#community-conditions-definitions) to assert when resolving conditional exports. By default, we match `['require', 'import', 'react-native']`.
+- [`unstable_conditionsByPlatform`](https://facebook.github.io/metro/docs/configuration/#unstable_conditionsbyplatform-experimental) — The additional condition names to assert when resolving for a given platform target. By default, this matches `'browser'` when the platform is `'web'`.
+
+:::tip
+**Remember to use the React Native [Jest preset](https://github.com/facebook/react-native/blob/main/template/jest.config.js#L2)!** Jest includes support for Package Exports by default. In tests, you can override which `customExportConditions` are resolved using the [`testEnvironmentOptions`](https://jestjs.io/docs/configuration#testenvironmentoptions-object) option.
```
NickGerleman, line 104: We are doing something similar in the RN 0.73 template `tsconfig`, but I did actually walk back the change I made in RN 0.72 to move to `moduleResolution: 'nodenext'`, because it didn't make sense to resolve by default when Metro wasn't yet, and the version of TypeScript shipped with 0.72 had some limitations.

So, TypeScript users may need to change their `moduleResolution` setting to enable package exports as well.

huntie: Good point! I've extended this note like so: 

image

Should only affect edge cases, and yeah we'll have this aligned by default for RN 0.73.

NickGerleman: `moduleResolution: 'bundler'` does have a caveat that it is not supported by TypeScript < 5, so the version of TypeScript we ship out of the box does not support it.

The closest thing supported in older versions to enable following exports is `moduleResolution: "nodenext"`, but it has some restrictions I'm not sure is consistent with Metro.

huntie:
@Pranav-yadav I don't think we need to over-highlight this, as it should be an extreme edge case for `"exports"` to locate modules that have differing TypeScript APIs (strong antipattern). Functionally, we don't expect anything to break in regular React Native projects with the default TypeScript config.

PR #3706, opened 2023-05-01
Update TextInput.md to add quirk with long-press selection on Android

hsource:
## Motivation

While debugging an issue on Android with text selection, I noticed this issue: 
[Long-press to select and press to move cursor doesn't work on `position: 'absolute'` TextInput outside parent bounds on Android #37181](https://github.com/facebook/react-native/issues/37181)

## Fix

Given that there's no easy fix, I thought the best thing to do is to document the workaround.

### DIFF
--- a/docs/textinput.md
+++ b/docs/textinput.md
@@ -81,9 +81,59 @@ const MultilineTextInputExample = () => {
 export default MultilineTextInputExample;
 ```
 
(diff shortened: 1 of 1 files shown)


lunaleaps:
Thanks @hsource for the PR!

PR #3707, opened 2023-05-02
Remove bad dependencies from useEffect

timur-svoboda:
Using ref current values in a dependency array is considered a bad practice. You can read more about it here: https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/. If you want to rerun `useEffect` when a new `Animated` object is created you need to use `useState`.

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -33,7 +33,7 @@ const FadeInView = props => {
       duration: 10000,
       useNativeDriver: true,
     }).start();
-  }, [fadeAnim]);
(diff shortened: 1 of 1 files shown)


NickGerleman (changes requested):
This fails the React "Rules of Hooks" ESLint rule.
```
 error  React Hook useEffect has a missing dependency: 'fadeAnim'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

PR #3713, opened 2023-05-05
Update shadow-props.md for Sets the drop shadow zIndex for Android.

rjeyasundharam:
Sets the drop shadow zIndex for Android.

### DIFF
--- a/docs/shadow-props.md
+++ b/docs/shadow-props.md
@@ -130,6 +130,7 @@ const App = () => {
   const [shadowOffsetHeight, setShadowOffsetHeight] = useState(0);
   const [shadowRadius, setShadowRadius] = useState(0);
   const [shadowOpacity, setShadowOpacity] = useState(0.1);
+  const [zIndex, setZIndex] = useState(0);
 
   return (
     <View style={styles.container}>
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hello @rjeyasundharam, when we talk about shadow props we refer to the types structure from React Native core types:
* https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts#L117-L122

In this context, `zIndex` is a Layout Prop (also cross-platform), and we already list it in there (https://reactnative.dev/docs/layout-props#zindex), so probably if we really want, we can add example for that prop it should be there, but I do not see the reason to update shadows example, which also atm only allow to run Snack example on iOS anyway.

rjeyasundharam:
Hi @Simek,
I am a beginner in React Native. Recently, I have been facing a zIndex issue while trying to add shadows to my card element. However, the solution mentioned in the official documentation (https://reactnative.dev/docs/shadow-props) did not work for me. So, I searched and found a solution on Stack Overflow (https://stackoverflow.com/a/48123658).

Issue #3715, opened 2023-05-08
Remove usages of `<Slider />` in our examples


NickGerleman:
React Native 0.72 removes the deprecated `Slider` component. Several of the Expo Snack based examples still use it, which we need to clean up to be able to upgrade the documents to 0.72.

### Usages:
- [ ]  `docs/shadow-props.md`
- [ ]  `docs/text-style-props.md`

PR #3716, opened 2023-05-08
docs: replace Slider usages to use @react-native-community/slider

tarunrajput:
Resolves: #3715

React Native 0.72 removes the deprecated Slider component. Several of the Expo Snack based examples still use it, which we need to clean up to be able to upgrade the documents to 0.72.
This PR updates all the usages of Slider to use @react-native-community/slider community package

### DIFF
--- a/docs/shadow-props.md
+++ b/docs/shadow-props.md
@@ -8,10 +8,10 @@ import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import con
 <Tabs groupId="language" queryString defaultValue={constants.defaultSnackLanguage} values={constants.snackLanguages}>
 <TabItem value="javascript">
 
-```SnackPlayer name=Shadow%20Props&supportedPlatforms=ios&ext=js
+```SnackPlayer name=Shadow%20Props&supportedPlatforms=ios&ext=js&dependencies=@react-native-community/slider
 import React, {useState} from 'react';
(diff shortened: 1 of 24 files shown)


NickGerleman:
I think it might be tricky to add `@react-native-community/slider'` as a dependency. Looking at the [deployed example](https://deploy-preview-3716--react-native.netlify.app/docs/next/text-style-props?language=typescript) we see this:

image

I would recommend instead tackling this change so that the example UIs do not rely on a slider, or relies on some thin helper components for one inline in the example. The flow to do this would be to:
1. Visit the example on the website, like https://reactnative.dev/docs/shadow-props
1. Open up each Snack example
1. Make the edits there, and look at the example
1. Copy the edits back to just `docs`once it example looks good

Simek (changes requested):
To make third-party dependencies work in inline Snacks out of the box you need to add them to the `dependencies` param in the URL (need to be encoded):
* https://github.com/facebook/react-native-website/blob/main/plugins/remark-snackplayer/src/index.js#L45-L45

tarunrajput:
@Simek @NickGerleman I've updated the PR by adding `@react-native-community/slider` to dependencies param.
Test link: https://deploy-preview-3716--react-native.netlify.app/docs/shadow-props

NickGerleman (approved):
LGTM

PR #3720, opened 2023-05-12
chore: add react-native-keys in security section to secure keys in react native

numandev1:
Just added a [react-native-keys](https://github.com/numandev1/react-native-keys) package in the security section because this package can secure and manage ENVs than all other available solutions

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -13,7 +13,7 @@ In this guide, you will learn about best practices for storing sensitive informa
 
 ## Storing Sensitive Info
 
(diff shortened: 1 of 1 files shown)


numandev1:
@Simek ?

Simek:
Hey @numandev1,

Security is a quite sensitive page, we usually do not add new statements or alter it content significantly, unless it's really needed. 

In general, in React Native docs we try to avoid recommending one library, over others, it can be seen for example on the removed API/Component pages, where we link to the directory:
* https://reactnative.dev/docs/imagepickerios

Saying that, we might want to review and probably rewrite the paragraph with packages links on the Security page in near future, to match the pattern mentioned above.

So in case of your package, I would recommend adding it to the React Native directory instead, to gain more users and traction in the community, which also should yield more marketing value than mention on the security page:
* https://reactnative.directory

PR #3721, opened 2023-05-15
Add onStartReached and sync onEndReached doc

janicduplessis:
- Add docs for `onStartReached` for FlatList and SectionList
- Sync up docs with source code for `onEndReached` for FlatList and SectionList


Simek (commented):
Refs:
* https://github.com/facebook/react-native/pull/35321

Hey @janicduplessis, thank you for the followup on core PR! 👍 

After looking at the code, I think it would be nice to clarify where exactly those propos are defined. By that I mean that the props should not be a part of `FlatList` and `SectionList`, but be listed in `VirtualizedList` props, which are than extended but listed earlier components. We already show the parent props link at the top of API documentation.

There were some concerns and benefits around duplicating propos of parent in child components, it's fine to leave it as is for now since is a broader topic, but no matter which approach we choose, the `VirtualizedList` props also need to be updated.

janicduplessis:
Ping @NickGerleman @cortinico

Simek:
I have asked around, and it seems that people do not have a strong preference around that, let's ship this as is. We can always follow up with changes, if someone will find this pattern confusing/cumbersome. 🙂

PR #3726, opened 2023-05-19
feature: Added Messenger Desktop to showcase

blakef:
Also added support for Desktop application links + additional padding to align long application titles.

## Desktop Messenger
CleanShot 2023-05-19 at 11 52 39@2x

## Added support for Desktop application links
Here's an example (not committed) showing layout in the extreme case:
CleanShot 2023-05-19 at 11 56 32@2x

## Added min-height to title to correct for long title offsetting links

**Before:**
CleanShot 2023-05-19 at 11 59 57

**After:**
CleanShot 2023-05-19 at 11 52 53@2x


Simek:
Tested the changes locally, on desktop and mobile, LGTM! 👍 

On thing that we might want to consider changing (but it's not a bloacker) is removing one entry from `pinned` list, or adding a few more apps in there, since now the icon grid on the home page looks a bit off:

Screenshot 2023-05-19 at 18 10 39

Simek (approved):

blakef:
Thanks @Simek, I've followed this up with the fix for the pinned list funk: #3727

PR #3727, opened 2023-05-19
fix: added pinned showcase entries

blakef:
In #3726, adding another pinned entry put the landing page showcase off grid.  Added 4 others to fix this.

## Now
CleanShot 2023-05-19 at 17 26 26@2x


Simek:
Can we also update the mattermost app icon to one taken from AppStore (so it fills the icon shape better)?

* https://apps.apple.com/us/app/mattermost/id1257222717

Simek (approved):
LGTM, thank you for follow up on that! 🙏

PR #3738, opened 2023-05-30
Change instructions due to menu renaming in macOS Ventura and Xcode 14

chousheng:
macOS Ventura and Xcode 14 rename some menus:

- **System Preferences** -> **System Settings** in **Finder**
- **Preferences** -> **Settings** in **Xcode**
- **Components** -> **Platforms** in **Xcode >  Settings**

This PR renames all related references while retaining the old ones (to support the case that older macOS or Xcode versions is used).

### DIFF
--- a/docs/_getting-started-macos-ios.md
+++ b/docs/_getting-started-macos-ios.md
@@ -27,13 +27,13 @@ If you have already installed Xcode on your system, make sure it is version 10 o
 
 #### Command Line Tools
 
-You will also need to install the Xcode Command Line Tools. Open Xcode, then choose "Preferences..." from the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
+You will also need to install the Xcode Command Line Tools. Open Xcode, then choose **Settings... (or Preferences ...)** from the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
 
 ![Xcode Command Line Tools](/docs/assets/GettingStartedXcodeCommandLineTools.png)
 
 #### Installing an iOS Simulator in Xcode
 
-To install a simulator, open <strong>Xcode > Preferences...</strong> and select the <strong>Components</strong> tab. Select a simulator with the corresponding version of iOS you wish to use.
+To install a simulator, open **Xcode > Settings... (or Preferences...)** and select the **Platforms (or Components)** tab. Select a simulator with the corresponding version of iOS you wish to use.
 
 #### CocoaPods
 
--- a/docs/_integration-with-existing-apps-swift.md
+++ b/docs/_integration-with-existing-apps-swift.md
@@ -75,7 +75,7 @@ Assume the [app for integration](https://github.com/JoelMarcey/swift-2048) is a 
 
 ### Command Line Tools for Xcode
 
-Install the Command Line Tools. Choose "Preferences..." in the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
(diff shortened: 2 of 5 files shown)


NickGerleman (approved):

docs/_integration-with-existing-apps-swift.md
```diff
@@ -117,7 +117,7 @@ Sending stats
 Pod installation complete! There are 3 dependencies from the Podfile and 1 total pod installed.
 ```
 
-> If this fails with errors mentioning `xcrun`, make sure that in Xcode in **Preferences > Locations** the Command Line Tools are assigned.
+> If this fails with errors mentioning `xcrun`, make sure that in Xcode in **Settings.. (or Preferences...) > Locations** the Command Line Tools are assigned.
```
Simek, line 120:
```suggestion
> If this fails with errors mentioning `xcrun`, make sure that in Xcode in **Settings... (or Preferences...) > Locations** the Command Line Tools are assigned.
```

Small nit, missing dot.

cortinico (approved):

docs/debugging.md
```diff
@@ -120,7 +120,7 @@ Custom debugger commands executed this way should be short-lived processes, and
 You can use Safari to debug the iOS version of your app without having to enable "Debug JS Remotely".
 
 - On a physical device go to: `Settings → Safari → Advanced → Make sure "Web Inspector" is turned on` (This step is not needed on the Simulator)
-- On your Mac enable Develop menu in Safari: `Preferences → Advanced → Select "Show Develop menu in menu bar"`
+- On your Mac enable Develop menu in Safari: `Settings (or Preferences) → Advanced → Select "Show Develop menu in menu bar"`
```
Simek, line 123:
```suggestion
- On your Mac enable Develop menu in Safari: `Settings... (or Preferences...) → Advanced → Select "Show Develop menu in menu bar"`
```

Label in Safari also include ellipsis, so let's add the dots here too. 🙂 

Btw. We probably also might want to reformat steps in the whole list, so they use font weight instead of code block, but this can be done in the future, as a separate change.

docs/_getting-started-macos-ios.md
```diff
@@ -27,13 +27,13 @@ If you have already installed Xcode on your system, make sure it is version 10 o
 
 #### Command Line Tools
 
-You will also need to install the Xcode Command Line Tools. Open Xcode, then choose "Preferences..." from the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
+You will also need to install the Xcode Command Line Tools. Open Xcode, then choose **Settings... (or Preferences ...)** from the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
```
Simek, line 30:
```suggestion
You will also need to install the Xcode Command Line Tools. Open Xcode, then choose **Settings... (or Preferences...)** from the Xcode menu. Go to the Locations panel and install the tools by selecting the most recent version in the Command Line Tools dropdown.
```

Whitespace nit.

docs/_integration-with-existing-apps-swift.md
chousheng: Good catch!

chousheng:
Thank you for the suggestions, @Simek! I've updated the PR.

Please let me know if you would like me to squash the two commits.

Simek:
Thank you for the update! :+1:

No squash need, but it looks like you have applied only one suggestion, but resolved all of them. With all suggestions in this should be ready to merge.

chousheng:
Thank you, @Simek! You are right. I missed applying two suggestions in the batch.

I have updated the PR.

Simek (approved):

PR #3740, opened 2023-05-31
 change mode option with configuration in 0.71 version

aymane-missouri-nw:
In the documentation's "Publishing to App Store" section, there is a mistake in the "Build app for release" step. The error lies in mentioning the "--mode" option in the "run-ios" command, which should instead be "--configuration".

image


cortinico (changes requested):
The correct flag is `--mode`:
https://github.com/react-native-community/cli/blob/main/packages/cli-platform-ios/README.md#build-ios

PR #3748, opened 2023-06-05
Updating docs to reflect bumping node from 16 to 18

arushikesarwani94:
React Native is increasing the minimum Node JS requirement from 16 to 18 in https://github.com/facebook/react-native/pull/37709 since Node 16's EOL is 2023-09-11 https://nodejs.org/en/blog/announcements/nodejs16-eol/

Updating documentation to reflect the same.


cortinico (approved):
LGTM 👍 We need to wait till 0.72 gets released to merge this otherwise the Docs for 72 will be affected

NickGerleman:
@arushikesarwani94 @cortinico looks like we should merge this now.

PR #3752, opened 2023-06-09

zenmasterjobo:
zulu11 isn't the name anymore, looks like the cask is just called zulu now

### DIFF
--- a/docs/_getting-started-macos-android.md
+++ b/docs/_getting-started-macos-android.md
@@ -25,10 +25,10 @@ We recommend installing the OpenJDK distribution called Azul **Zulu** using [Hom
 
 ```shell
 brew tap homebrew/cask-versions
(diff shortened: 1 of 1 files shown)


cortinico:
That's not correct. We actually ask users to install `zulu11` otherwise `zulu` will just pull the latest JDK version which might not always work with React Native

zenmasterjobo:
@cortinico Ah ok, is there a different way I can go about installing it? 

image


I probably should have been more clear on the issue that the provided instructions don't seem to work, but perhaps I have my homebrew misconfigured? 

Looking on homebrew's website there isn't a cask called zulu11 either.

zenmasterjobo:
I found my mistake, in the docs my brain skipped past that first line that says to tap the `cask-versions`

Apologies!

Issue #3751, opened 2023-06-08
Fix out-of-bounds iOS/Android emulators in examples


NickGerleman:
Some of the examples on the React Native documentation are missing part of their content when viewed on Android/iOS.

https://reactnative.dev/docs/image-style-props

<img width="480" alt="image" src="https://user-images.githubusercontent.com/835219/236763258-adb87df3-9a3b-44d7-a64c-fa77ec62765d.png">

<img width="480" alt="image" src="https://user-images.githubusercontent.com/835219/236768374-c3640e40-2f4d-40d8-a46b-9b74acdd8d82.png">

This is because the website sets a fixed height for the player, and the rendering of the device we were showing has changed.

https://github.com/facebook/react-native-website/blob/d67805884a7c0920dcefc1af5894580984918239/website/src/css/customTheme.scss#L1717

@byCedric did great work in #3724 to allow the React Native website to customize the frame the emulator is shown in. We should leverage that, and change our styling + embed to a new value which avoids being cut off.

spencerlwahlstrom:
Hello, I attempted to fix this issue with #3753  by adjusting the height, tested this on multiple pages of the site.

PR #3753, opened 2023-06-10
Out of boundsfix

spencerlwahlstrom:
Fixed out-of-bounds issue #3751, for embedded snack player cutting off android, and ios devices. Changed snack players height so view is no longer cutoff and there is sufficient space surrounding it.

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1714,7 +1714,7 @@ html[data-theme="dark"] .docsRating {
 /* Snack Player */
 
 .snack-player {
-  height: 544px;
+  height: 740px;
   width: 100%;
   overflow: hidden;
   margin-bottom: 24px;
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1714,7 +1714,7 @@ html[data-theme="dark"] .docsRating {
 /* Snack Player */
 
 .snack-player {
-  height: 740px;
+  height: 711px;
   width: 100%;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


NickGerleman:
I think this should be stable now that we are specifying the Appetize Frame.

I think we could further optimize a bit, like asking Appetize not to show the frame, so we can take back a little bit of space, but viewing this on the website, it doesn't look too tall to my eyes (though that is of course subjective). I do notice in the iOS example we end up having a larger gap below the device than above. Is this needed to fit the current Android emulator?

Since this is kinda stylistic though, I want to give other reviewers an opportunity to share their thoughts before hitting the merge button.

Simek:
Setting a static height is still a sort of workaround, since the viewport height of Android and iOS screens differs. Like Nick, I also would like to see the player version with device frames disabled. Maybe then we can guarantee same dimensions? 

From the quick test on the deploy previe we can reduce the base height to `691px` which is the size of Snack with Android screen and frame (iOS one height is `661px`).

A quick comparison shot:
Screenshot 2023-06-12 215754

Snack player with new height looks nice in complex examples, but in short ones, it would be nice if we can spare a bit of height viewport, not sure ATM exactly how. Maybe max height could be an additional param, but still, with device frames on we might not able to do much.

Simek:
After dropping the device frame, the simulator content cropping should no longer occurs, thus wrapper height change is no longer needed. Closing this PR for now, it there are still some issues with Snacks rendering let's create a new issues/PR for them. Thanks @spencerlwahlstrom so much for bringing this problem to our attention! 🙏

PR #3732, opened 2023-05-24
Default to `yarn` commands

Pranav-yadav:
### Summary

We're steering away from `npx` due to it's unreliability and switiching to `yarn` instead, at least for below commands:

| Current 👴 | New 🧑 |
| --- | --- |
| `npx react-native start` | `yarn start`
| `npx react-native run-android` | `yarn android`
| `npx react-native run-ios` | `yarn ios`

### Context

This was discussed here: https://github.com/facebook/react-native/pull/37521#discussion_r1200782476

PR #3755, opened 2023-06-12
Refactor: Switch to `let` & `const` from `var` in `JS`, `JSX`, `TS` & `TSX` code-blocks

Pranav-yadav:
### Summary

This is diff update the number of `JS`, `JSX`, `TS` & `TSX` code-blocks to switch to `let` & `const` from `var`.
And of course some markdown formatting touchups like code formatting.

P.S.: I came across these `var` references when working on #3732 😇

### DIFF
--- a/docs/_integration-with-existing-apps-java.md
+++ b/docs/_integration-with-existing-apps-java.md
@@ -178,7 +178,7 @@ const HelloWorld = () => {
     </View>
   );
 };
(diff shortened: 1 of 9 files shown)


cortinico (approved):

Simek (approved):
thank you for those fixes, LGTM! 👍

PR #3773, opened 2023-06-27
Update _getting-started-linux-android.md : SDK manager paths

jeromeabel:
Add more precisions on selecting SDK manager


cortinico:
@jeromeabel why is this change needed?

jeromeabel:
> @jeromeabel why is this change needed?

It's just a proposal.

I find it useful to update the path to access to SDK Manager. In my setup (Xubuntu 22.04), it seems that I have a different window configuration in Android Studio. 

android-studio-on-xubuntu-sdk-manager

Also, the "Preferences" dialog is named "Settings". I wonder if sorting steps with bullets points might be easier to follow. Or even adding an image. See also, the name of "Google API", there is a "_64" added. 

Capture d’écran_2023-06-27_15-46-58

Also, it could be useful to add an extra note about "Create a new device". It seems important to choose a device with the PlayStore column checked to use it with Expo.

Capture d’écran_2023-06-27_15-48-34

PR #3778, opened 2023-06-29
Upgrade to Docusaurus v2.4.1

slorber:
Hey, Docusaurus maintainer here 👋 

I'll soon attempt an upgrade of this site to Docusaurus v3.

As a first step, I'd like to upgrade the site to the latest stable v2 version, so that the other PR can focus only on the v2 => v3 breaking changes.

---

I'm also adding some convenient package scripts that you'll eventually find useful to run from the repo root. 

If you don't want that, I can remove them but would appreciate to at least keep the ability to run `yarn docusaurus xyz` from the repo root.

For example, this is a good way to test a production build of the website locally, using a limited number of versions like on Netlify previews (builds much faster):

```bash
yarn website:build:fast
yarn website:serve
```

---

I'm also using this PR to introduce you to a tool called [Argos CI](https://argos-ci.com/) that I have used successfully on the Docusaurus repository to prevent visual regressions during package upgrades. I'll edit this PR once the setup is ready to be demonstrated.

Sometimes when upgrading Docusaurus from one version to another, some UI elements and custom CSS breaks, and we notice it after merge. A tool like Argos can compare visually the main site to the deployment preview, and tell you how they differ. 

My setup on Docusaurus involves taking screenshots of all the URLs found in `sitemap.xml`. Here's the kind of link each new docs PR could generate: https://app.argos-ci.com/meta-open-source/docusaurus/builds/134/50874925

(new tool, their UI is kind of rough but it's working nicely)

---

Note, due to a commit from yesterday (https://github.com/facebook/react-native-website/pull/3776), your site builds with errors because it uses both Docusaurus 2.4.1 and 2.3.1 package, while official package versions are supposed to be aligned. This little upgrade also fixes the issue ;)

CF the Netlify build: https://app.netlify.com/sites/react-native/deploys/649c06ab24479c000817b0e2

CleanShot 2023-06-29 at 17 21 30@2x

PR #3780, opened 2023-06-30
Upgrade React-Native website to Docusaurus v3

slorber:
WIP PR to upgrade to Docusaurus v3 canary

I will let you know when it's ready to review

---

Deploy preview: https://deploy-preview-3780--react-native.netlify.app/

Argos visual tests 
- Reports: https://app.argos-ci.com/slorber/rnw-visual-tests
- PR: https://github.com/slorber/rnw-visual-tests/pull/2

---

Note to myself: we can revert usage of `｜` in favor of `\|`, see change made in https://github.com/facebook/react-native-website/pull/3807#discussion_r1297318844

---

Follow-up of:
- v2.4.1 upgrade: https://github.com/facebook/react-native-website/pull/3778
- v3 docs preparation - iteration 1: https://github.com/facebook/react-native-website/pull/3807
- v3 docs preparation - iteration 2: https://github.com/facebook/react-native-website/pull/3830
- v3 docs preparation - iteration 3: https://github.com/facebook/react-native-website/pull/3843


slorber:
@Simek there are docs that I had to modify so that they can compile with MDX 2.

Those are mostly isolated in this commit, only updating the last version:
[`ad86759` (#3780)](https://github.com/facebook/react-native-website/pull/3780/commits/ad86759c72f9363e50696a007c9d9ae4691cb770)
(older versions are probably also affected by the same problems)

If you use expressions like `{x: number, y: number}`, `version <5` or `<COMMIT_HASH>` in MD, you either have to escape them with `\` or find another way to create the same text.

I used escaping here, but IMHO you should rather use inline code blocks for many of these things. I also noticed that some API table types are already in a code block, and some others are not, so this is not super consistent, and MDX 2 catched that.

```md

Before, doesn't work with MDX 2
| ({nativeEvent: [PressEvent](pressevent)}) => boolean |

Escaped:
| (\{nativeEvent: [PressEvent](pressevent)}) => boolean |

Inline code block:
| `({nativeEvent: [PressEvent](pressevent)}) => boolean` |
```

My suggestion is to use inline code blocks consistently everywhere for such cases.

If you can do this right now on existing docs and merge it, that would be a nice preliminary work for this PR, and would help reduce the diff so that this PR remains easy to review.

plugins/remark-snackplayer/package.json
```diff
@@ -21,11 +21,10 @@
   "dependencies": {
     "dedent": "^0.7.0",
     "object.fromentries": "^2.0.3",
-    "unist-builder": "^2.0.3",
     "unist-util-visit-parents": "^3.1.1"
   },
   "devDependencies": {
-    "remark": "^12.0.1",
+    "remark": "^11",
```
cipolleschi, line 27: why this has been downgraded? 🤔

plugins/remark-snackplayer/package.json
```diff
@@ -21,11 +21,10 @@
   "dependencies": {
     "dedent": "^0.7.0",
     "object.fromentries": "^2.0.3",
-    "unist-builder": "^2.0.3",
```
cipolleschi, line 24: we are removing this because we don't need it?

website/docusaurus.config.js
```diff
@@ -20,6 +20,9 @@ const commonDocsOptions = {
   remarkPlugins: [require('@react-native-website/remark-snackplayer')],
 };
 
+const isDev = process.env.NODE_ENV === 'development';
```
cipolleschi, line 23: I think that If this is not used, we can remove it.

website/package.json
```diff
@@ -48,10 +48,10 @@
     ]
   },
   "dependencies": {
-    "@docusaurus/core": "^2.4.1",
-    "@docusaurus/plugin-google-gtag": "^2.4.1",
-    "@docusaurus/plugin-pwa": "^2.4.1",
-    "@docusaurus/preset-classic": "^2.4.1",
+    "@docusaurus/core": "0.0.0-5658",
```
cipolleschi, line 51: Are these version numbers correct? 🤔 
when are we going to move to 3.y.z?

cipolleschi (commented):
Hi @slorber, thank you so much for this PR. I left a few comments, mainly for clarification.
The only significant piece I am not confident to review are the Snack updates: at first glance, they looks good to me, but  let's wait for another pair of eyes to double check them.

plugins/remark-snackplayer/package.json
slorber: This is because it is used for the remark snackplayer code but it only works for MDX v1 and its usage has been removed.

CleanShot 2023-09-12 at 15 01 12@2x


MDX v1 used to be more based on "string concatenation".

With MDX v2, JSX elements have first-class MDX AST nodes that you can construct manually.

This code has been replaced by: 

CleanShot 2023-09-12 at 15 02 55@2x

website/package.json
slorber: @cipolleschi those are canary Docusaurus versions: https://docusaurus.io/community/canary

As soon as this PR is merged, I'll release Docusaurus v3-beta.0 and we can replace those with the first beta.

Or if you agree to merge, I can also publish v3-beta.0 first, and update this PR before merging.

My goal here is to be able to say to the community that a few major Docusaurus sites are already running on Docusaurus v3 beta (Docusaurus + Jest + React-Native), showing it is safe, and encourage adoption. 

I don't consider those canaries to be unstable myself, and could eventually release this canary as v3 as GA today. This is mostly a way to reassure the Docusaurus community and help increase adoption so that we can catch the few minor bugs that we might have missed. Similarly to when RN releases a RC and is looking for community feedback ;)

plugins/remark-snackplayer/package.json
slorber: Hmm, can't remember and looks suspicious. Reverting.

website/docusaurus.config.js
slorber: yes 👍 removed

website/core/DocsRating.js
```diff
@@ -5,11 +5,12 @@
...
  */
 
-import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
 import React, {useState} from 'react';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 
 const DocsRating = ({label}) => {
-  if (!ExecutionEnvironment.canUseDOM) {
+  const isBrowser = useIsBrowser();
+  if (!isBrowser) {
```
slorber, line 13: This fixes a little React hydration issue. The issue was already there but is now logged thanks to React 18 new hydration callback `onRecoverableError(error)` (https://react.dev/blog/2022/03/29/react-v18)

The SSR/SSG render and the first CSR render outputs should match, so things such as `typeof window !== "undefined"` etc should be avoided.

slorber:
> Hi @slorber, thank you so much for this PR. I left a few comments, mainly for clarification. The only significant piece I am not confident to review are the Snack updates: at first glance, they looks good to me, but let's wait for another pair of eyes to double check them.

Thanks @cipolleschi

Regarding the SnackPlayer, unless you know well the differences between MDX v1/v2 it would be hard to review technically. IMHO you'd rather focus on the outcome: is the snack player still working on all the pages in the deploy preview, or do you see any regression?

website/package.json
cipolleschi: Thanks for the explanation. I think that they are fine as they are.

cipolleschi:
> Regarding the SnackPlayer, unless you know well the differences between MDX v1/v2 it would be hard to review technically. IMHO you'd rather focus on the outcome: is the snack player still working on all the pages in the deploy preview, or do you see any regression?

Not really, I was just mentioning it in case someone among @cortinico @TheSavior or @Simek wants to double-check that code! :D

website/src/css/customTheme.scss
```diff
@@ -1865,3 +1865,13 @@ html[data-theme="light"].blog-wrapper {
...
+
+/*
+Prevent useless bottom margin for multiline <figcaption> tags in MDX:
+<figcaption>
+  some paragraph text
+</figcaption>
+ */
+figcaption > p:last-child {
+  margin-bottom: 0;
+}
```
TheSavior, line 1877: This needs a new line at the end of the file it seems

website/blog/2018-03-05-AWS-app-sync.md
```diff
@@ -94,7 +94,7 @@ return <S3Album track/>
 
 AWS Amplify favors a convention over configuration style of development, with a global initialization routine or initialization at the category level. The quickest way to get started is with an [aws-exports file](https://aws.amazon.com/blogs/mobile/enhanced-javascript-development-with-aws-mobile-hub/). However, developers can also use the library independently with existing resources.
 
-For a deep dive into the philosophy and to see a full demo, check out the video from [AWS re:Invent](https://www.youtube.com/watch?v=vAjf3lyjf8c).
+For a deep dive into the philosophy and to see a full demo, check out the video from [AWS re\:Invent](https://www.youtube.com/watch?v=vAjf3lyjf8c).
```
TheSavior, line 97: Why does this need to be escaped?

website/blog/2015-03-26-react-native-bringing-modern-web-techniques-to-mobile.md
```diff
@@ -18,9 +18,7 @@ Together with the rapid iteration cycle of the web, we've been able to build som
 <footer>
   <a
     href="https://code.facebook.com/posts/1014532261909640/react-native-bringing-modern-web-techniques-to-mobile/"
-    className="btn">
-    Read more
-  </a>
+    className="btn">Read more</a>
```
TheSavior, line 21: Was this change made by a formatter or manually?

TheSavior (approved):
The snack changes seem fine, and the snack player seems to work in the preview

website/src/css/customTheme.scss
slorber: I added a new line but not sure why it's needed? CI is passing. Is there linter I don't see?

website/blog/2018-03-05-AWS-app-sync.md
slorber: Docusaurus v3 now uses https://github.com/remarkjs/remark-directive to provide support for admonitions, and other things using syntax such as `:textDirective`, `::leafDirective` and `:::containerDirective`.

`Re:invent` is parsed as `Re` + `:invent` text directive.

This is what happens in the MDX playground if you turn the remark-directive option on:

CleanShot 2023-09-14 at 13 06 03@2x


This regression was captured by the visual regression tests:

CleanShot 2023-09-14 at 12 51 24@2x

---

Technically it should probably be possible to add some code in Docusaurus v3 so that this escaping becomes un-necessary, and the unhandled AST directive nodes get serialized back as raw text. For now it's not implemented, and not even sure who should implement that 😅. Will think about it. I hope in the meantime it's not a blocker for you to merge this PR.

slorber:
Thanks for the review @cipolleschi @TheSavior

Let me know if I can do anything else ;)

cipolleschi (approved):

cipolleschi (approved):

website/src/css/customTheme.scss
TheSavior: I mentioned it because GitHub showed this icon presumably because there wasn't an extra new line, whereas all other files have that extra new line. 
CleanShot 2023-09-15 at 17 48 25

And yeah, it's normally enforced with a linter. See react-native's main eslint config: https://github.com/facebook/react-native/blob/2be409ff55b50563a5f123f7823fa7cdb72dbef9/packages/eslint-config-react-native/index.js#L264

PR #3820, opened 2023-08-17
[website] add VerticalTable, improve new arch appendix

Simek:
# Why

Currently, on the New Architecture Appendix page the Typescript section table breaks the desktop and mobile display, also the Flow section readability is not that great.

Screenshot 2023-08-17 at 14 23 22

# How

This PR adds vary basic `VerticalTable` component, with data structure which allows us to avoid HTML entities (used due to MDX parse limitations/problems) and unifies the data structure and appearance between Flow and TypeScript sections.

The changes have also been backported to the two latest versioned docs.

~~One issue I can see is that relative component import would need to be updated manually after cutting new version. Would be nice to have a app root import alias, or something allowing to avoid relative import paths in docs (CC @slorber).~~

> This has been resolved by using [`@site` alias](https://docusaurus.io/docs/versioning#use-absolute-import-within-the-docs) in import paths.

# Preview

Screenshot 2023-08-17 at 14 06 51

### DIFF
--- a/docs/new-architecture-appendix.md
+++ b/docs/new-architecture-appendix.md
@@ -4,6 +4,7 @@ title: Appendix
 ---
 
 import NewArchitectureWarning from './\_markdown-new-architecture-warning.mdx';
+import VerticalTable from '../website/core/VerticalTable';
 
 <NewArchitectureWarning/>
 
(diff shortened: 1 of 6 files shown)


cortinico (approved):
Thanks for polishing this @Simek I've checked the preview and it looks great 👌

cipolleschi (approved):

PR #3824, opened 2023-08-21
Fix native module type conversion tables for iOS and Android

flokol120:
Within the native module documentations there were two errors in the type conversion tables.

For Android the Promise type was missing from the conversion table.

For iOS two NSNumber conversions were present in the type conversion table. One of which stated NSNumber would be converted into a ?boolean. This table entry was removed as this seems to be not true. The other NSNumber entry which states NSNumber is converted to number? was left untouched.


flokol120:
> Thanks for sending this @flokol120 Could you accept our CLA here: https://code.facebook.com/cla otherwise we won't be able to accept your code

Strange, I already accepted it 🤔. But in this case I am going to do it again.

Issue #3637, opened 2023-03-22
Ambiguity on why the index property in ViewToken is marked optional


jobpaardekooper:
## Description

On [the ViewToken page in the docs](https://reactnative.dev/docs/viewtoken) it shows that the `index` property is optional. This is not explained in any further detail by the docs.

## What is the problem?
It is unclear to me in what situation this property would be `null`. This object is used as a parameter for the `onViewableItemsChanged` callback in `VirtualizedList` so it is in turn also used in both `FlatList` and `SectionList`. This makes me think there wouldn't be a situation where `index` would be `null`.

## How can we address it?

Update the description of the `index` property on `ViewToken` to clarify in what situation it could be `null`.

If anyone knows a resource like an issue in the react-native repo or a from post that discusses this or can explain it to me I can also make a PR to clarify the description on the property.

## Why is it important?

It is currently ambiguous why the `index` property on `ViewToken` is marked as optional.

## Who needs this?

Anyone using the `onViewableItemsChanged` prop and relies on the indexes of the callback.

jobpaardekooper:
No response yet so it would not consider this "fixed". (Just commenting so the issue will stay open, as it should)

pedropbazzo:
https://github.com/facebook/react-native-website/pull/3835 @jobpaardekooper

jobpaardekooper:
@pedropbazzo please see my comment on your PR. I don't think it solves this issue.

PR #3835, opened 2023-08-26
settings doc upgrade issue #3637

pedropbazzo:


jobpaardekooper:
In my opinion this does not clear up the confusion om _why_ the property would be null.

The text added by this PR is just a different way of saying index is optional. The docs already mention it can be optional so the change does not add more value over what is already there.

It should be cleared up and explained in what situation the index property would be null.

PR #3842, opened 2023-09-01
docs: transform origin

intergalacticspacehighway:
Adds documentation for [transformOrigin](https://github.com/facebook/react-native/pull/38559#issuecomment-1702758720) prop.

### DIFF
--- a/docs/transforms.md
+++ b/docs/transforms.md
@@ -164,9 +164,7 @@ export default App;
 
 # Reference
 
-## Methods
-
-### `transform()`
+## `Transform`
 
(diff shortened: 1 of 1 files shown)


javache:
Can you also provide examples of the array syntax? While it's not web-compatible, it should be more efficient to use in RN, as it avoids having to do string parsing in the update path.

javache:
Can you also fix the structure of this page? We shouldn't have `methods`, and `Transform` can be its own top-level section.

intergalacticspacehighway:
@javache updated

PR #3847, opened 2023-09-06
Remove out dated caveat about using setTimeout from a headless js task

jkells:
Hey Team,

I think this caveat is outdated.

This was fixed in https://github.com/facebook/react-native/pull/33044 and it's now possible to use setTimeout from a HeadlessJsTask.

See:
https://github.com/facebook/react-native/pull/33043
https://github.com/facebook/react-native/pull/33044

### DIFF
--- a/docs/headless-js-android.md
+++ b/docs/headless-js-android.md
@@ -193,7 +193,6 @@ If you wish all errors to cause a retry attempt, you will need to catch them and
 
 ## Caveats
 
-- The function passed to `setTimeout` does not always behave as expected. Instead the function is called only when the application is launched again. If you only need to wait, use the retry functionality.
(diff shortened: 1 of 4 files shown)


Simek (commented):
Hello @jkells, thank you for catching that! From the context you have provided it looks like we can indeed remove that note. 

Can you also backport this change to the versioned docs located in:

Since according to the commit tags, it has been shipped with `0.69` release:
* https://github.com/facebook/react-native/commit/dac56ce0776e0e4d23ed4f8b324f2e2432aefa6a

jkells:
No probs @Simek that's done.

Simek (approved):
Thanks for the updates, LGTM! 🚀

PR #3850, opened 2023-09-16
change file extension to ts in turbo modules creation guide

groot007:
Minor fix of file extension cause for someone it can be confusing to have .js file in final result when typescript code was used at the beginning


Simek (approved):

PR #3854, opened 2023-09-22
Fix link to redirect to standalone react-devtools

Randall71:


Simek (approved):

PR #3861, opened 2023-09-28
Draft: Packaging guide for iOS

michalchudziak:
This PR introduces a comprehensive guide on packaging React Native modules as `.framework` for iOS:

- Describes the motivation to package React Native modules for integration with native apps.
- Provides step-by-step instructions, from creating a new React Native app to building the framework.
- ... TODO

The goal of this new documentation is to equip developers with the knowledge needed for seamless integration of React Native modules into native iOS projects.


cipolleschi:
Hi @michalchudziak, thanks for this guide, but I have a few questions:
1. how can I see the preview from netlify? I see that the guide is not connected with the sidebar and if I navigate to `docs/packaging-ios`, I can't see it.
2. There are some linting error that should be fixed
3. What's the goal with this guide? From which need does it come from? I am not sure that this is what we actually want to promote as a way to package libraries: we already have a guide to package iOS (and Android) modules use NPM and that should be the suggested way to ship those libraries. On top of that, packaging the code as XCFramework is not that straightforward if we think about the New Architecture and the code generated by Codegen: the framework you'd like to ship will not build as it needs the files generate by codegen that lives in the app space and not in the framework space.
So, as of today, I don;t think that we can/should ship a guide like this: it is quite advanced and it can create confusion in the majority of React Native users.

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+## Creating a 2048 High Scores Package
+
+In this guide, you'll craft a High Scores screen, similar to the one made in the [guide to integrate with native apps](https://reactnative.dev/docs/integration-with-existing-apps?language=objc). You'll package it as a standalone `.xcframework` file. Afterward, you can use it directly in a native app like any native library, eliminating the need for installing any JavaScript dependencies.
+
+### Starting a New React Native App
+
+Begin by generating a fresh React Native project:
+
+```bash
+npx react-native init HighScores2048
```
thymikee, line 32:
```suggestion
npx react-native@latest init HighScores2048
```

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+
+4. **Linking Dependencies**: It's crucial to link all dependencies of the React Native module accurately. This ensures that when the module is activated from a native app, it operates with all the required code and resources.
+
+## Prerequisites
+
+1. Follow the [standard native environment setup](https://reactnative.dev/docs/environment-setup?guide=native).
+
+## Creating a 2048 High Scores Package
+
+In this guide, you'll craft a High Scores screen, similar to the one made in the [guide to integrate with native apps](https://reactnative.dev/docs/integration-with-existing-apps?language=objc). You'll package it as a standalone `.xcframework` file. Afterward, you can use it directly in a native app like any native library, eliminating the need for installing any JavaScript dependencies.
```
keisyrzk, line 25: What if I like to use only a certain component not a whole "screen"? I understand the doc "how can I integrate a component" so a single view not exactly whole screen. I would like to see guidance for this case + the whole screen if the implementation paths are different for them.

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Starting a New React Native App
+
+Begin by generating a fresh React Native project:
```
keisyrzk, line 29: Why do I need to create a whole project if it is said we can integrate components? I would like to see explanation if this is the only way.

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+Adjust the export in the `index.js` file using the `AppRegistry`. This serves as our React component tree's entry point. You can register several components with `AppRegistry` and link them to the native UI tree via `RCTRootView`.
+
+```js
+import {AppRegistry} from 'react-native';
+import HighScores from './HighScores';
+
+AppRegistry.registerComponent('HighScores', () => HighScores);
+```
+
+Finally, modify the code in `AppDelegate.mm` so it references the correct component. Also, feed mocked data to the view for development testing.
```
keisyrzk, line 95: I have not noticed when you created an iOS project. Besides, you assume there is `AppDelegate`. Not sure about ObjC but using Swift there is more likely you see `SceneDelegate` and no `AppDelegate`. The description seems to be incomplete.

docs/packaging-ios.md
michalchudziak: The guide covers that case, but perhaps we should make it more explicit. Simply register the component in AppDelegate, and you should be good to go. The guide how to export a UIView instance, not a ViewController or anything that would make it work only as a screen.

docs/packaging-ios.md
michalchudziak: Right. In that case we can register multiple root views and duplicate the binding files. I think it would be good to explain that explicitly in the docs tho.

docs/packaging-ios.md
michalchudziak: I'm reusing the project generated by the RN template, I haven't accounted for creating a fresh project form the scratch. Also it only serves as a playground, so I don't consider it crucial for the whole integration.

cipolleschi:
@michalchudziak I agree that XCFrameworks can bring benefit, but we can have this page in the official documentation, unfortunately.

This can make for an amazing post in your company blogpost, but for the official documentation we want to have something that is battle-tested and work in all the scenarios, including the New Architecture, and that integrates well with the current tooling (e.g. autolinking).
At the moment, none of these requirements are true:
- We are not using packaged Xcframeworks internally (except from hermes-engine, which required a non trivial amount of work to have it working properly) => Not battle tested
- It doesn't work well in all the scenarios => the New Architecture is not supported
- It does not integrates well with the current tooling => it doesn't work with autolinking (as you have to reference the target manually and to modify the `AppDelegate.mm`) and it requires multiple steps to integrate from our users.

Also considering that now we can build dependencies with `use_frameworks!` and dynamic linking, the pros of packaging the library as XCFramework are obtainable in another way and with much less effort.

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+... (no changes here)
+```
+
+Run the app to verify its functionality. At this stage, you should see the described screen.
+
+![Screenshot](/docs/assets/PackagingPlaygroundApp.png)
+
+### Setting Up a New Package Target
+
+Now, create a separate Xcode target to build our `.xcframework` package. Open the `ios/HighScores2048.workspace` file. In Xcode, navigate to the targets section and select **Add new target (+)**.
```
keisyrzk, line 107: You create in in Xcode (your XCode project) or this is a RN project's iOS part? Why do you create a framework? It is said we will integrate RN code in to XCode project. The vast majority of cases user will have an XCode app project not a framework. There is no explanation why you do so.
MY goal and understanding:
Integrate some RN components into my own standalone XCode app project. I do not see this here.

docs/packaging-ios.md
keisyrzk: `The guide how to export a UIView instance, not a ViewController` - perhaps I misunderstood sth but I believe we want to export RN code to work in XCode project not UIView into RN project.

docs/packaging-ios.md
keisyrzk: But wait, we would like to use RN components inside MY OWN XCODE PROJECT. I understand we need to create a framework using RN project to finally use it inside XCode project. I miss more explicit explanation what is the goal and steps to achieve this.

docs/packaging-ios.md
```diff
@@ -0,0 +1,228 @@
...
+
+![Screenshot](/docs/assets/PackagingAddNewTarget.png)
+
+From the options, pick `Framework` and name it **HighScore2048Module**. Ensure the **Embed in Application** option remains off.
+
+![Screenshot](/docs/assets/PackagingCreateNewTarget.png)
+
+If done correctly, you'll see a `HighScore2048Module` folder on the left sidebar and a new yellow suitcase icon in the targets section.
```
jbinda, line 115: typeo in name of the module (missing `s`). Should be: `HighScores2048Module`

PR #3869, opened 2023-10-03
Update scrollview.md

adriancuadrado:
`pagingEnabled` **is** supported on Android. I tested it and it just seems to work.

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -515,8 +515,6 @@ Possible values:
 
 When true, the scroll view stops on multiples of the scroll view's size when scrolling. This can be used for horizontal pagination.
 
-> Note: Vertical pagination is not supported on Android.
-
 | Type | Default |
 | ---- | ------- |
 | bool | `false` |


Simek (commented):
Hey @adriancuadrado, thank you for spotting that! 👍 

It looks like this comment has been added a while ago in https://github.com/facebook/react-native/pull/14844 but it is still present in the source code:
* https://github.com/facebook/react-native/blob/353d8d3980e04c32c233f5eebbe4ef3bc9bce460/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L545-L551

Can you prepare the PR removing it from source, ideally with test/demo in RNTester? Alternatively it would be nice to find PR/commit adding the support. With the fact confirmed, and changes in, we can follow up on this docs PR.

adriancuadrado:
@Simek
> Hey @adriancuadrado, thank you for spotting that! 👍
> 
> It looks like this comment has been added a while ago in [facebook/react-native#14844](https://github.com/facebook/react-native/pull/14844) but it is still present in the source code:
> 
> * https://github.com/facebook/react-native/blob/353d8d3980e04c32c233f5eebbe4ef3bc9bce460/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L545-L551
> 
> Can you prepare the PR removing it from source, ideally with test/demo in RNTester? Alternatively it would be nice to find PR/commit adding the support. With the fact confirmed, and changes in, we can follow up on this docs PR.

This is the PR: https://github.com/facebook/react-native/pull/39801

I can't add a test/demo in RNTester because I am getting some errors when I follow [these instructions](https://github.com/facebook/react-native/blob/main/packages/rn-tester/README.md). Mind doing it yourself?

PR #3872, opened 2023-10-07
chore: remove references to StyleSheetRegistry from StyleSheet docs

alecmarcus:
Based on some digging, it seems as though `StyleSheet` no longer uses a registry of `ID`s: 

https://github.com/react-native-community/discussions-and-proposals/issues/400
https://github.com/facebook/react-native-website/pull/629#issuecomment-442501414

This PR updates the docs to reflect that change.

### DIFF
--- a/docs/stylesheet.md
+++ b/docs/stylesheet.md
@@ -3,7 +3,7 @@ id: stylesheet
 title: StyleSheet
 ---
 
-A StyleSheet is an abstraction similar to CSS StyleSheets
+A StyleSheet is an abstraction similar to CSS StyleSheets.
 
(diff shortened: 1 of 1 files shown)


javache:
This is absolutely true, but we do want to keep the option of optimizing this data structure in the future. Users should also not assume a style object is something they can mutate or introspect (without going through flatten), so it'd be great to acknowledge that in the docs.

rikur:
Now I'm even more confused. The `StyleSheet` could really use an update so that users know the best practices.

I've been living in a make-believe world of StyleSheet IDs moving through the bridge for years 😕

PR #3875, opened 2023-10-10
refactor: removed mention of js polyfills that are not used any more

ecreeth:
### Summary

Both `JSC` and `hermes` implements these methods, so we aren't using a polypill version of them.

- Array.from
- Array.prototype.includes
- Array.prototype.{find, findIndex}
- Object.{entries, values}
- Object.assign
- String.prototype.{startsWith, endsWith, repeat, includes}

See [string-functions](https://github.com/facebook/hermes/blob/main/test/hermes/string-functions.js), [object-functions](https://github.com/facebook/hermes/blob/main/test/hermes/object-functions.js) and [array-functions](https://github.com/facebook/hermes/blob/main/test/hermes/array-functions.js) for reference.


cortinico (commented):
I'm not the best person to review this one. @NickGerleman do you happen to know?

PR #3878, opened 2023-10-11
Improve instructions for enabling app transport security

elibroftw:
As someone who uses Windows, it isn't ideal to have to do extra work after buying a Macbook. This PR will save time as not everyone is familiar with setting a value to false in Info.plist.


cipolleschi (approved):

cipolleschi:
nit: it would be nice to add an image of Xcode to show the Allow-Arbitrary Load field
Screenshot 2023-10-16 at 14 06 31

elibroftw:
Sounds reasonable. I believe I have allowed maintainers to edit. Either way just let me know where images go and how to reference images and I can do it.

cipolleschi (approved):

PR #3881, opened 2023-10-11
fix: default value of textBreakStrategy prop of TextInput component

aswin-s:
This PR updates the default value of `textBreakStrategy` prop of `TextInput` component to 'highQuality' instead of 'simple'. This has been the default behaviour for quite some time now and it is tripping off developers. More details regarding this mismatch is available at https://github.com/facebook/react-native/issues/38244


Simek (approved):
After digging a bit through the code it looks like `highQuality` has been a default value for a while. Nice catch, thank you for the updates! 👍

PR #3883, opened 2023-10-12
Add export compliance information

elibroftw:
Just uploaded to Apple Store Connect and got met with an export compliance form. This is what Apple told me to do to bypass it. Would've helped if it was in the docs beforehand.


cipolleschi (approved):

cipolleschi:
Thanks for taking the time to improve the documentation!

PR #3884, opened 2023-10-12
Add App Store screenshot requirements

elibroftw:
Need this as well when publishing to App Store.


Simek:
@cipolleschi Can we marge this PR, or are there any blockers?

PR #3889, opened 2023-10-18
[docs] Create "Debugging Release Builds" guide, consolidate existing pages

huntie:
- Replaces the `symbolication` and `sourcemaps` pages with a new "Debugging Release Builds" guide, now located under Debugging. This minimally describes how to enable source maps and symbolicate release builds. Code references have been updated.
- `/docs/sourcemaps` and `/docs/symbolication` are redirected to the new page.


Simek:
Thanks for updating the guides! 🙏 New content looks really nice, just have  one nit about redirect.

Currently we use the static, serve side redirects for most of the moved pages:
* https://github.com/facebook/react-native-website/blob/8fc23770911514fc8bad2b99c2015e5cca28353b/website/static/_redirects#L3

If there is no strong reason why to use client side approach in those cases, we probably should stick to the solution above.

huntie:
@Simek Thanks, was hoping we had something like that. Updated

PR #3899, opened 2023-10-24
Fix a backlink on hermes-as-the-default

Huxpro:
Hi ;) 

Came across and figured by "last Oct" it was intended to link to my old blog post 😛


Huxpro:
Seems like the CLA is signed according to https://github.com/facebook/jsx/pull/158 

Probably need sometime to get the tag added

cortinico (approved):

PR #3906, opened 2023-11-01
docs: fix fragment for AlertOptions

la55u:
The `options` parameter is currently not linkable because it should point to the #alertoptions fragment. This PR fixes that.


Simek (commented):
Nice catch @la55u, thank you for the fix! 👍 

Can you port those changes to the versioned docs too?

la55u:
@Simek I changed those too except in 0.70 where the heading is named `Options` so the anchors work correctly there.

Simek (approved):
Thank you for the updates, LGTM! 👌

PR #3908, opened 2023-11-01
Added missing xmlns to FrameLayout

scarlac:
Added missing `xmlns:...` to `<FrameLayout ...` on the guide for adding RN to Fragments instead of activities.
Without specifying xmlns in newer versions:

image


Simek (commented):
Thank you for spotting that! 👍 

Can you also update the versioned files for that doc?
* https://github.com/facebook/react-native-website?tab=readme-ov-file#versioned-docs

Simek (approved):
Thank you for the updates, let's ship it! 🚀

PR #3913, opened 2023-11-03
[Text] refactor: useSelect listed under text props instead of text style props

tarunrajput:
`userSelect` is text style prop but it is listed under text props.


lunaleaps:
@tarunrajput Thank you for fixing!

PR #3918, opened 2023-11-07
fix example code prop name

sarvarrose:
Based on https://github.com/react-native-community/cli/blob/main/docs/dependencies.md#dependency, the field is `dependency` not `dependencies`.


Simek (approved):
Hello @sarvarrose, nice catch! Thanks for the context and fix, LGTM! 👍

PR #3919, opened 2023-11-08
Remove Background thread from the threading model documentation

rshest:
This modifies threading model docs to exclude mentioning of the background thread, which is not relevant anymore.


sammy-SC:
looks good, thanks for correcting this.

PR #3926, opened 2023-11-21
include `pdf` in list of supported image source formats

koenpunt:
PDFs work as an image source on iOS.

Edit: I used pdfs as image source using a `.xcassets` catalog, not sure if using a pdf directly works as well.


koenpunt:
Nevermind, pdfs apparently only work when in an asset catalog.

PR #3936, opened 2023-12-04
add correct type of Pressable style prop

sarvarrose:


Simek (approved):
Hey @sarvarrose, nice catch! Thanks for updating the types in tables to match description and actual code state. 👍 

#### Refs
* https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/Pressable/Pressable.js#L165

PR #3937, opened 2023-12-04
Change 'use' to 'lose' to signify that functions are still accessible

ksshp:
The statement at the end says that since the generated abstract class itself extends ReactContextBaseJavaModule, all the functions are still accessible. But the sentence reads "should not use", which does not make sense, and "should not lose" should instead be used.


NickGerleman (approved):

ksshp:
@NickGerleman

PR #3942, opened 2023-12-07
fixed peer dependency issues

Shantanugupta43:
Fixed peer dependency issues 

<h3> Before fix </h3>

issue

<h3> After fix </h3>

solved


Shantanugupta43:
Ready for merge

NickGerleman:
I missed that these dependencies weren't all already covered, but that CI did not complain that the change would need a lockfile change.

I reverted for now, since it looks like it might introduce a new version of major depencies into the three like webpack instead of resolving to the previous one.

PR #3946, opened 2023-12-07
Lint and typecheck examples against RN 0.73

NickGerleman:
Updates the harness for linting snack examples to the newest version of RN.


NickGerleman:
Netlify Deploy is using Node 16. Need to update.

PR #3949, opened 2023-12-08
Change 'use' to 'lose' to signify that functions are still accessible

KartikShankhavaram:
The statement at the end says that since the generated abstract class extends ReactContextBaseJavaModule, all the functions are still accessible. But the sentence reads "should not use" which does not make sense, and "should not lose" should be used instead.

### DIFF
--- a/docs/new-architecture-library-android.md
+++ b/docs/new-architecture-library-android.md
@@ -115,4 +115,4 @@ class NativeAwesomeManager(reactContext: ReactApplicationContext) :
 </TabItem>
 </Tabs>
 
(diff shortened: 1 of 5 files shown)


Simek (commented):
Thanks for the correction and backporting the changes! 👍 

Can you resolve the conflict before we can ship? Looks like you have done PR in a small window when 0.73 docs were generated including minor version. The path should be just `version-0.73` after rebase

Simek (approved):
Thanks for the updates, LGTM! 👌

PR #3955, opened 2023-12-13
Fix(doc/native-modules-ios): Replaced old term CalendarManager

githubjeka:


githubjeka:
Hi

When you start reading about [native-modules-ios](https://reactnative.dev/docs/native-modules-ios) **CalendarModule** will created. But **CalendarManager** appears in the middle of the text


CalendarManager was in [v0.60 doc](https://reactnative-archive-august-2023.netlify.app/docs/0.60/native-modules-ios) , 
Most likely CalendarManager changed to the CalendarModule during some commit, and it was not done well.

Now reading the documentation, it is impossible to write the code correctly, according to the examples. This fix brings clarity to the examples.

cipolleschi (approved):

cipolleschi:
Thanks for the clarification! I didn't see the misalignment before! 😮

PR #3964, opened 2023-12-24
Fixed a mistake in the code example in CalculatorPackage.kt in Turbo Native Modules guide.

### LINKED ISSUE #3963: Turbo Native Modules - error in code example in CalculatorPackage.kt
## Description

In the `CalculatorPackage.kt` file, there is a mistake in the code example. The `getModule` method of the `CalculatorPackage` class has an argument named "name" that is nullable. However, this argument in the extended class `BaseReactPackage` is non-null. This inconsistency causes an error in the Gradle build process.

extended class: 
https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/BaseReactPackage.java

## Documentation version

React Native v0.73 https://reactnative.-dev/docs/the-new-architecture/pillars-turbomodules

photo_2023-12-24 17 20 02

Franson33:
Description: 
In the CalculatorPackage.kt file, there is a mistake in the code example. The getModule method of the CalculatorPackage class has an argument named "name" that is nullable. However, this argument in the extended class BaseReactPackage is non-null. This inconsistency causes an error in the Gradle build process.

extended class:
https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/BaseReactPackage.java

Fixes #3963

photo_2023-12-24 17 20 02
Screenshot 2023-12-24 at 17 54 59
-

### DIFF
--- a/docs/the-new-architecture/pillars-turbomodule.md
+++ b/docs/the-new-architecture/pillars-turbomodule.md
@@ -707,7 +707,7 @@ import com.facebook.react.module.model.ReactModuleInfoProvider
 
 class CalculatorPackage : TurboReactPackage() {
 - override fun getModule(name: String?, reactContext: ReactApplicationContext): NativeModule? = null
-+ override fun getModule(name: String?, reactContext: ReactApplicationContext): NativeModule? =
++ override fun getModule(name: String, reactContext: ReactApplicationContext): NativeModule? =
 +   if (name == CalculatorModule.NAME) {
 +     CalculatorModule(reactContext)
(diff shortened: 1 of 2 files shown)


cortinico (changes requested):
Thanks for the fix @Franson33 

Could you update also the versioned file here:
```
website/versioned_docs/version-0.73/the-new-architecture/pillars-turbomodule.md
```

cortinico (approved):

PR #3972, opened 2024-01-03
Add CircleCI job running `yarn build`

NickGerleman:
We have a job for linting, but the only thing running the build is the netlify deploy, whose logs aren't public. Add CircleCI build check that is public.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,14 @@ jobs:
       - run_yarn
       - run: yarn ci:lint
 
+  build:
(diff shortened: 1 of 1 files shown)


NickGerleman:
This keeps on saying it is getting killed... Need to investigate.

NickGerleman:
Error code 137 is OOM. So... the website build is OOMing I guess?

NickGerleman:
I think that fixed things.

cortinico (approved):

PR #3976, opened 2024-01-05
Give the CircleCI machines more memory

NickGerleman:
Give it 8GB of memory since Docusaurus is liking to OOM during build, even when we restrict Node heap size.


cortinico (approved):

NickGerleman:
From https://github.com/facebook/docusaurus/issues/4765 it looks like Docusaurus can be very hungry. Going to bump this to xlarge, just in case.

PR #3980, opened 2024-01-07
Update handling-text-input.md

Sivac07:
I have set the text input padding as 100. Because the text input box will show in the center of device instead of top of the device. So, it will be more convenient to user to test the input box .

### DIFF
--- a/docs/handling-text-input.md
+++ b/docs/handling-text-input.md
@@ -14,7 +14,7 @@ import {Text, TextInput, View} from 'react-native';
 const PizzaTranslator = () => {
   const [text, setText] = useState('');
   return (
-    <View style={{padding: 10}}>
+    <View style={{padding: 100}}>
       <TextInput
         style={{height: 40}}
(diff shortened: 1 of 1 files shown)


docs/handling-text-input.md
```diff
@@ -14,7 +14,7 @@ import {Text, TextInput, View} from 'react-native';
 const PizzaTranslator = () => {
   const [text, setText] = useState('');
   return (
-    <View style={{padding: 10}}>
+    <View style={{padding: 100}}>
```
Simek, line 17: Hello @Sivac07, thanks for the update! 👍 

Let's use the different layout to active the same effect instead of modifying the padding value.

```suggestion
    <View style={{flex: 1, justifyContent: 'center'}}>
```

We can also add the border to text input to increase its visibility.

Screenshot 2024-02-16 at 13 45 17

Simek:
Closing in favour of:
* #4848

Issue #3977, opened 2024-01-05
The docs mention a hook `useDimensions` from @react-native-community/hooks. This hook does not exist.


kimthompson:
## Description

The [React Native documentation around `useWindowDimensions()`](https://reactnative.dev/docs/usewindowdimensions) contains the following line:

> The [useDimensions](https://github.com/react-native-community/hooks#usedimensions) hook from the community [React Native Hooks](https://github.com/react-native-community/hooks) library aims to make handling screen/window size changes easier to work with.

That hook doesn't exist, and was apparently [removed from the library](https://github.com/react-native-community/hooks/blob/1b19bc04159665434f29410e070b7e4f867fe181/CHANGELOG.md?plain=1#L14) in February of last year. 

## What is the problem?

I attempted to download `useDimensions` when I ran into a weird issue using `useWindowDimensions` (probably on our end) and discovered that it didn't exist. That was a little weird, so I thought I'd bring it up here.

## How can we address it?

I think you can probably just delete this bullet point.

## Why is it important?

I know it's a minor thing, but it was still a jarring thing to run into while following the documentation.

## Who needs this?

All RN devs who might look into the `useWindowDimensions()` hook.

lunaleaps:
Thanks for reporting! I had no idea about this hooks library, would you like to submit a PR to remove it? We can probably remove it from all versions of the docs we have active.

PR #3983, opened 2024-01-09
remove community packages references from useWindowDimensions

tarunrajput:
Fixes: #3977
Removes community alternatives for the useWindowDimensions hook, as they have either been removed or are no longer maintained.


lunaleaps:
Thanks @tarunrajput for your help cleaning up the React Native documentation!

PR #3987, opened 2024-01-12
Added steps to install Android SDK Command-line Tools

luishenriqueturani:
Added steps to install Android SDK Command-line Tools


luishenriqueturani:
Yes, you are correct, it appears that an anomaly has occurred here.
I'm having problems running the build of my project, and running "npx
react-native doctor" it shows me that the Android SDK was not being
recognized, even though it was installed with the ANDROID_HOME environment
variable defined, so I went to the system settings Android Studio and I saw
that this package, Android SDK Command Line Tools, was not installed, I
installed and ran the "npx react-native doctor" command again and it
started to be recognized. However, after restarting my system again (Fedora
39) it no longer recognized the Android SDK, so I still cannot run the
build.
Sorry for taking your time, I'll be more careful and test more next time.

Em sex., 12 de jan. de 2024 às 10:56, Nicola Corti ***@***.***>
escreveu:

PR #3990, opened 2024-01-18
Fix bad link in 2022-07-08-hermes-as-the-default.md

Huxpro:
Oops how did I miss that!


Simek (approved):
Nice catch, thanks for the correction! 🙏

cortinico (approved):

PR #4008, opened 2024-02-08
Change class example to a function

gaearon:
Self-explanatory.


lunaleaps (approved):
Can we fix the lint? lgtm otherwise

PR #4029, opened 2024-03-01
[For Netlify Redirect Test] Add a fake redirect

lunaleaps:
See https://answers.netlify.com/t/placeholder-redirects-dont-support-periods/113543/2

Head's up I couldn't reply to that thread anymore because Netlify reported my account as spam
If you try `https://deploy-preview-4029--react-native.netlify.app/docs/0.72/sourcemaps` this link -- the redirect fails

If anyone else can message on that thread with this PR, that will be helpful, thanks!

Edit: Actually what's even stranger is that it does work.. but just not for 0.72? 
Do we hardcode something for 0.72? I'm very confused

### DIFF
--- a/website/static/_redirects
+++ b/website/static/_redirects
@@ -1,5 +1,8 @@
 # Redirects from what the browser requests to what we serve
 
+# Adding this for testing
+/docs/:version/sourcemaps       https://github.com/reactwg/react-native-new-architecture#guides
+
 # Renamed pages
 /docs/android-setup             /docs/getting-started
 /docs/building-for-apple-tv     /docs/building-for-tv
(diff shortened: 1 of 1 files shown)


lunaleaps:
@knupska hey! haha yes an aggressive spam filter

Yea this is super strange since hard-coding a redirect from `0.72` also fails. It's pretty high-up on the _redirects page too. Anyway, it does seem related to Netlify's placeholders and periods as I had concluded -- thanks for all your help investigating!

lunaleaps:
Figured it out, we needed to actually delete the documentation for the redirects to match. It's not 100% consistent though which is still a mystery but unblocked!

PR #4030, opened 2024-03-01
Migrate inline requires + RAM bundles docs to new "optimizing JS loading" guide

motiz88:
1. As a key followup from https://github.com/expo/expo/issues/27279#issuecomment-1971610698, we want to better explain the current state of the "inline requires" feature in React Native, which lacks full ESM support and is disabled by default in Expo. This PR achieves that, but also...
2. ...rewrites the existing "RAM Bundles and Inline Requires" page more extensively while we're at it. The existing page is pretty outdated, and for many users, has the wrong emphasis. In particular, the new framing allows us to highlight `React.lazy` as a safer and more intentional alternative to applying the inline requires transform wholesale to a codebase.

## [View the new page](https://deploy-preview-4030--react-native.netlify.app/docs/next/optimizing-javascript-loading)

Also note that [/docs/next/ram-bundles-inline-requires](https://deploy-preview-4030--react-native.netlify.app/docs/next/ram-bundles-inline-requires) redirects to the new page.


blakef:
@motiz88 and I spoke offline, the feature has now been [remove](https://github.com/facebook/react-native/commit/58b45e86a6fd78409dc358e27ae736c95fee9f39) from the CLI.  These docs decouple RAM bundles and make it easier to remove from the docs.  I'll follow once this lands.

blakef:
If you're happy to strip RAM bundles out, please do.

motiz88:
Hmm, one reason not to remove all mention of RAM bundles is that it would break existing links from the Metro docs. https://metrobundler.dev/docs/configuration#gettransformoptions

image

motiz88:
Discussed with @blakef further online. Assuming the RAM bundle feature isn't going away in 0.74, it makes sense to land this PR *with* RAM bundle docs and then follow up (across RN and Metro's docs) after the RN docs for 0.74 are cut.

docs/optimizing-javascript-code-loading.md
```diff
@@ -0,0 +1,222 @@
...
+:::
+
+### Tip: Avoid module side effects
+
+Lazy-loading components can change the behavior of your app if your component modules (or their dependencies) have _side effects_, such as modifying global variables or subscribing to events outside of a component. Most modules in React apps should not have any side effects.
+
+```tsx title="SideEffects.tsx"
+import Logger from './utils/Logger';
+
+// Side effect! This must be executed before React can even begin to render
```
rickhanlonii: Could add a 🚩

```suggestion
// 🚩🚩🚩 Side effect! This must be executed before React can even begin to render
```

docs/optimizing-javascript-code-loading.md
```diff
@@ -0,0 +1,222 @@
...
+
+Hermes is the default engine for new React Native apps, and is highly optimized for efficient code loading. In release builds, JavaScript code is fully compiled to bytecode ahead of time. Bytecode is loaded to memory on-demand and does not need to be parsed like plain JavaScript does.
+
+:::info
+Read more about using Hermes in React Native [here](./hermes).
+:::
+
+## Recommended: Lazy-load large components
+
+If a component with a lot of code/dependencies is not likely to be used when initially rendering your app, you can use React's [`lazy`](https://react.dev/reference/react/lazy) API to defer loading its code until it's rendered for the first time. Typically, you should consider lazy-loading "screen"-level components in your app, so that adding new screens to your app does not increase its startup time.
```
rickhanlonii: The quotes make it harder to read:

```suggestion
If a component with a lot of code/dependencies is not likely to be used when initially rendering your app, you can use React's [`lazy`](https://react.dev/reference/react/lazy) API to defer loading its code until it's rendered for the first time. Typically, you should consider lazy-loading screen-level components in your app, so that adding new screens to your app does not increase its startup time.
```

docs/optimizing-javascript-code-loading.md
```diff
@@ -0,0 +1,222 @@
...
+// the SplashScreen component, and can unexpectedly break code elsewhere in
+// your app if you later decide to lazy-load SplashScreen.
+global.logger = new Logger();
+
+export function SplashScreen() {
+  // ...
+}
+```
+
+## Advanced: Call `require` inline
```
rickhanlonii: Should we callout and link to the Pitfalls section below?

docs/optimizing-javascript-code-loading.md
```diff
@@ -0,0 +1,222 @@
...
+        <Text>Load</Text>
+      </TouchableOpacity>
+      {this.state.needsExpensive ? <VeryExpensive /> : null}
+    </View>
+  );
+}
+```
+
+:::info
+Some React Native frameworks disable this behavior. In particular, in Expo projects, `require` calls are not inlined by default. You can enable this optimization by editing your project's Metro config and setting `inlineRequires: true` in [`getTransformOptions`](https://metrobundler.dev/docs/configuration#gettransformoptions).
```
rickhanlonii: Could we add a code block here with the full example? When I was first looking into this, it wasn't clear to me how to actually structure the JSON to add this option, something like:

```
// metro.config.js
{
  ...
  transformer: {
    getTransformOptions: {}
  }
}
```

... I actually can't even write the example without looking up a metro config file from somewhere

rickhanlonii (approved):
LGTM! Left some nits to consider but we can merge as-is!

docs/optimizing-javascript-code-loading.md
motiz88: I'm treating those as pitfalls of *automatic* inline requires. With manual inlining (which is already called out as "advanced"), you are in control, so if your code doesn't work, you can dig into why; you aren't dealing with surprising breakages "at a distance" like you would be with automatic inlining.

docs/optimizing-javascript-loading.md
```diff
@@ -0,0 +1,222 @@
...
+
+:::info
+Some React Native frameworks disable this behavior. In particular, in Expo projects, `require` calls are not inlined by default. You can enable this optimization by editing your project's Metro config and setting `inlineRequires: true` in [`getTransformOptions`](https://metrobundler.dev/docs/configuration#gettransformoptions).
+:::
+
+### Pitfalls of inline `require`s
+
+Inlining `require` calls changes the order in which modules are evaluated, and can even cause some modules to _never_ be evaluated. This is usually safe to do automatically, because JavaScript modules are often written to be side-effect-free.
+
+If one of your modules does have side effects - for example, if it initializes some logging mechanism, or patches a global API used by the rest of your code - then you might see unexpected behavior or even crashes. In those cases, you may want to exclude certain modules from this optimization, or disable it entirely.
```
TheSavior, line 124: Doesn't the transform also skip inlining modules that aren't stored in a variable?

```
import {useCallback, useState} from 'react';
import {TouchableOpacity, View, Text} from 'react-native';
require('ModuleWithSideEffects')
```

Won't that require never be inlined?

PR #4034, opened 2024-03-03
no source map file will be generated

YaoHuiJi:
react-native-xcode.sh cannot access the variable SOURCEMAP_FILE


huntie:
Stale: Closing

PR #4039, opened 2024-03-05
Add a new `Platform.isVision` property (React Native 0.74.x)

retyui:
My college @okwasniewski recently introduce a new Platform prop `isVision` https://github.com/facebook/react-native/commit/258d8e51b451b221e557dad4647cbd210fe37392

### DIFF
--- a/docs/platform.md
+++ b/docs/platform.md
@@ -112,6 +112,20 @@ Returns a boolean which defines if device is a TV.
 
 ---
 
+### `isVision`
+
+```tsx
+static isVision: boolean;
+```
+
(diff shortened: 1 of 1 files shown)


docs/platform.md
```diff
@@ -112,6 +112,20 @@ Returns a boolean which defines if device is a TV.
 
 ---
 
+### `isVision`
+
+```tsx
+static isVision: boolean;
+```
+
+Returns a boolean which defines if device is an Apple Vision.
```
okwasniewski, line 121: I would also add that this is true only in non-compatibility mode. If we are running on Apple Vision Pro (Designed for iPad) visionOS will return false and isPad will be true

lunaleaps (approved):

PR #4046, opened 2024-03-08
Add info block about moved docs for New Architecture

lunaleaps:
CleanShot 2024-03-08 at 13 28 47@2x


Simek (approved):
Great clarification idea! 👌

PR #4048, opened 2024-03-10
Cats should not drink dairy milk

what-the-functor:


Rexogamer:
taking a look, this appears to be true - [a lot of cats are lactose intolerant](https://www.pdsa.org.uk/what-we-do/blog/vet-qa-can-cats-drink-milk), and food also fits better imo since the variable is `isHungry`

cortinico (approved):

PR #4057, opened 2024-03-21
chore: changed href to the new react.dev site

svendjanis:
I changed the href link to point to https://react.dev/ instead of https://reactjs.org/. I know reactjs.org redirects to react.dev, but I feel like it could be updated on the site as well


Simek (approved):
Thanks for the fix, and welcome to the community!

PR #4069, opened 2024-03-28
Update optimizing-javascript-loading.md to remove class syntax

lewiscasewell:
Remove this.state which seems to be left over from a class component code snippet, in favour of functional component


cortinico (commented):
Thank you 🙏 Could you also update the versioned pages? @lewiscasewell

cortinico (approved):

PR #4105, opened 2024-05-06
docs: Update frame buffer size for VisionCamera example

mrousavy:
Just a tiny detail, it doesn't really matter much but I thought I wanted to clarify this a bit.

There's two formats in video processing; YUV and RGB.
RGB is always BGRA (4 bytes), and YUV is a bi-planar format with a Y plane of 1 byte per pixel, and UV plane of half the size of the image.

For 4k buffers, let's calculate the size of one frame:

- YUV = Y(3840 × 2160) + UV(3840 × 2160 / 2) = 12 MB
- RGB = (3840 × 2160 × 4) = 33 MB

While VisionCamera implements optimizations to trim the buffers and uses YUV or even compressed YUV whenever possible, almost 90% of the times people need to use RGB because the ML models just work in RGB.

So the exact number would be 33.177.600 bytes for a Frame, which is 1.990.656.000 bytes per second (or ~2 GB per second) of data flowing through the Frame Processor.

Thank to JSI it does not matter how big the data is, because we only pass references without making any copies or serialization - this is the part that should be highlighted here.


cortinico (approved):

Simek:
The New Architecture pages has been moved out of versioned docs, which make it impossible to resolve the conflict from the GitHub.

Closing in favor of:
* #4732

PR #4113, opened 2024-05-12
Replace RCTEventEmitter with RCTModernEventEmitter in Android Native UI Components guide

IgorKhramtsov:
I was following this guide and noticed that `RCTEventEmitter` is marked as deprecated in the latest (0.74) ReactNative release. Looks like there is a `RCTModernEventEmitter` that should be used instead.

> This [RCTModernEventEmitter] is a transitional replacement for RCTEventEmitter that works with Fabric and non-Fabric renderers. RCTEventEmitter works with Fabric as well, but there are negative perf implications and it should be avoided.

### DIFF
--- a/docs/native-components-android.md
+++ b/docs/native-components-android.md
@@ -217,7 +217,7 @@ class MyCustomView(context: Context) : View(context) {
     }
     val reactContext = context as ReactContext
     reactContext
-        .getJSModule(RCTEventEmitter::class.java)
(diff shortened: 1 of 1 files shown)


IgorKhramtsov:
Actually, the methods are different in `RCTModernEventEmitter` which also should be reflected in the docs and since it's not clear how to use new one I'm closing this PR.

PR #4116, opened 2024-05-15
Add hover text for tags in ScrollView

apicgg:
@cortinico  I know it's not much but it adds the hover text on platform tags for ScrollView page.

see #3047

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -88,7 +88,7 @@ A React Component that will be used to render sticky headers, should be used tog
 
 ---
 
-### `alwaysBounceHorizontal` <div class="label ios">iOS</div>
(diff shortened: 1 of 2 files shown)


Simek:
Thanks for working on this improvement! 👍 

We probably should make the platform tag a component, which accepts `platform` prop and based on that render the correct content, so we don't need to copy over the `alt` text to every platform mark use-case across smany doc pages. But this also meant that we need to replace all manual HTML header additions with new component.

Please let me know if you are interested in trying to introduce this new component and address the issue in more comprehensive way. Happy to help you with this, but if you are not interested it's also fine, I can make a follow up PR in the future.

apicgg:
@Simek Making it a platform tag component sounds good. I am more than happy to help introducing this new component. Thanks.

cortinico (approved):

apicgg:
This is being fixed by #4132

PR #4119, opened 2024-05-17
Fix hero styles content outside viewport (width)

LucasReinaldo:
Patch to fix hero image going outside viewport.


LucasReinaldo:
> Hey @LucasReinaldo, thanks for putting the fix out so quickly! 🙏
> 
> Unfortunately, I had to ship #4127 instead, due to missing CLA check on your PR, hope you can understand that.

That's all good, my bad for not getting back to it, just subscribed and happy to help anytime.

PR #4132, opened 2024-06-01
Add css for hover text on platform tags

### LINKED ISSUE #3047: iOS and Android tags are used, but not explained

apicgg:
Adds css for hover text which explains the platform tags (Android or iOS). This is a generic CSS rule which applies the style in every pages instead of manually adding `title` prop on each page.

closes #3047 

see #4116 

c/c @Simek


Simek (commented):
Hello @apicgg, the currently proposed solution unfortunately has its flaws - it is hard to read, and overwrites the already existing content in `:::after` used for tag "cut-out" visualization. 

Screenshot 2024-06-02 082931

Those issues are the reason why I still would lead more toward new component or a 3rd party library, which can dynamically hooked up to DOM elements with certain classes on client side.

apicgg:
Ah, I see now. I will try with your suggested approach as we were discussing earlier. 

Thanks for highlighting this.

I will convert this PR to draft now and will work on this.

Simek:
Thank you for the willingness of putting a bit more time into solving this! 🙏 

If you have more questions about this feel free reach me out on Discord or via X DMs.

apicgg:
@Simek I it okay for you to assign the issue #3047 to me?

Simek:
> I it okay for you to assign the issue https://github.com/facebook/react-native-website/issues/3047 to me?

@apicgg Yes, I think that enough time has passed since the last activity of the previous person who volunteer, but you need to comment on the issue first. Otherwise I cannot do this, as per:
* https://docs.github.com/en/issues/tracking-your-work-with-issues/assigning-issues-and-pull-requests-to-other-github-users#about-issue-and-pull-request-assignees

PR #4148, opened 2024-06-17
Mention React Native visionOS in OOT platforms page

szymonrybczak:
In this Pull Request I added React Native visionOS to `out-of-tree-platforms.md` page  that we created and maintain at @callstack.


Simek (approved):
Thanks for the addition! 👍

PR #4177, opened 2024-08-13
Upgrade dependencies. Remove Axios as dependency

realsoelynn:
This is to resolve the issue of Axios security vulnerabilities https://github.com/advisories/GHSA-8hc4-vh64-cxmj


TheSavior:
Have you validated that we have that docusaurus commit?

TheSavior (approved):
That's a strange lint error to come from this. Once that's fixed, this seems fine

PR #4178, opened 2024-08-15
PlatformTag Component

### LINKED ISSUE #3047: iOS and Android tags are used, but not explained

apicgg:
Adds a standalone component which takes `platform` as a prop to show the platform related tag and hover text which explains the platform tags (Android or iOS).

closes #3047

### DIFF
--- a/website/src/theme/PlatformTag/index.js
+++ b/website/src/theme/PlatformTag/index.js
@@ -1,0 +1,14 @@
+import React from 'react';
+
+const PlatformTag = ({platform}) => {
+  const platformLabel = platform === 'ios' ? 'iOS' : 'Android';
+  return (
+    <div
+      className={`label ${platform}`}
+      title={`This section is related to ${platformLabel} platform`}>
+      {platformLabel}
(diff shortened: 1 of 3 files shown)


Simek (approved):
Thanks for working on it, LGTM! 👍

PR #4187, opened 2024-08-19
update Docusaurus, fix swizzled components, correct authors data, UI tweaks

Simek:
# Why

We did not update Docusaurus lately, ans skipped few releases. Let's switch to the latest release, so we can utilize new blog authors functionalities.

# How

The following changes have been made:
* bump Docusaurus to 3.5 release
* fix already swizzled components
* swizzle out blog authors components to tweak UI and address `React is not defined` issue with social icons
* update, extract and move around data used for authors of blog posts to leverage the new feature
* move `PlatformTag` out of swizzle root
* few other small UI tweaks for the blog site section

CC: @slorber 

# Test plan

The changes have been reviewed by running website app locally. I was not able to spot any regressions.

# Preview

Screenshot 2024-08-19 at 13 04 10

Screenshot 2024-08-19 at 13 55 52


cortinico (approved):
I've skimmed through the diff and looks good to me 👍 

@slorber anything we need to be aware of over here?

website/src/theme/Blog/Components/Author/Socials/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
slorber, line 1: what did you change in this file? Looks similar to what we had before 🤔

website/src/theme/Blog/Components/Author/index.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
slorber, line 1: what did you change in this file?

slorber (commented):
Looks good to me

Maybe you'll want to add `page: true` to authors so that each author has an individual page?

(note: this means creating more pages, and might impact build time)

website/blog/2021-03-12-version-0.64.md
```diff
@@ -1,10 +1,6 @@
 ---
 title: Announcing React Native 0.64 with Hermes on iOS
-author: Mike Grabowski
-authorTitle: CTO and Co-Founder at Callstack
-authorURL: 'https://twitter.com/grabbou'
-authorImageURL: 'https://avatars0.githubusercontent.com/u/2464966?s=460&v=4'
-authorTwitter: grabbou
+
```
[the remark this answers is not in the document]
Simek: That was a mistake, fixed!

website/src/theme/Blog/Components/Author/Socials/index.js
Simek: Used swizzle command to extract `Blog/Components/Author`, this nested field was extracted in the process, reverted it.

website/src/theme/Blog/Components/Author/index.js
Simek: Only one `className`, but I can also overwrite it via CSS.

Simek: There as also few changes in module CSS file. Just removed the code file locally, but it looks like the style file is not used, when JS part is gone.

slorber: Yes we need the actual component to import the CSS.

What could help reduce the swizzle code a bit is to wrap the component instead of ejecting it, and then doing `<WrapperComponent className={styles.myCustomClassName}/>`. That's a trick to know about, but it is not always the most convenient in all situations.

PR #4194, opened 2024-08-19
Delete `progressviewios.md`

yungsters:
The recommendation since 0.72 has been to use [`@react-native-community/progress-view`](https://github.com/react-native-progress-view/progress-view), so this page is no longer relevant.

As such, this PR deletes them so that we no longer have to maintain them.

I've also updated the 0.72 blog post to link to the docs at 0.72 instead of next (so that the blog post links don't break when this PR is published).

PR #4195, opened 2024-08-19
Delete `slider.md`

yungsters:
The recommendation since 0.72 has been to use [`@react-native-community/slider`](https://github.com/callstack/react-native-slider/tree/main/package), so this page is no longer relevant.

As such, this PR deletes them so that we no longer have to maintain them.


yungsters:
This will pass after https://github.com/facebook/react-native-website/pull/4194.

PR #4200, opened 2024-08-29
Remove ref in render antipattern in animated docs

poteto:
Reading or writing to a ref in render is a [rule of React violation](https://react.dev/reference/react/useRef). This PR updates the docs for animated to use the `useAnimatedValue` hook instead, which does use a ref under the hood but isolates the rule of React violation to just that hook.

This allows users that follow code examples for animated to have their
components and hooks be compilable by React Compiler.

This hook was added in
[0.71](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#:~:text=Introduce%20useAnimatedValue%20hook%20to%20make%20it%20easier%20working%20with%20Animated.Values%20in%20function%20components.%20(e22217fe8b%20by%20%40fabriziocucci))
so I'm also updating versioned docs from 0.71 onwards.

### DIFF
--- a/docs/animated.md
+++ b/docs/animated.md
@@ -13,14 +13,21 @@ The core workflow for creating an animation is to create an `Animated.Value`, ho
 
 The following example contains a `View` which will fade in and fade out based on the animated value `fadeAnim`
 
-```SnackPlayer name=Animated%20Example
-import React, {useRef} from 'react';
-import {Animated, Text, View, StyleSheet, Button} from 'react-native';
+```SnackPlayer name=Animated%20Example&supportedPlatforms=ios,android
+import React from 'react';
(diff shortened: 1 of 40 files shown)


yungsters (approved):
Thanks!

poteto:
> Isn't useAnimatedValue also reading ref.current in render? https://github.com/facebook/react-native/blob/8b2ccea243d8b8156d6f6e1e3e08b6064f82ed84/packages/react-native/Libraries/Animated/useAnimatedValue.js#L24

Yeah that's true. But updating the docs means that we can remove a Rule of React violation from user code, so if React Compiler is being used these components/hooks can still be compiled, while the offending useAnimatedValue will not (if we compile it).

PR #4202, opened 2024-08-30
docs: Update running-on-device.md for iOS to include mention about USB-C cables

petetnt:
Times change and Lightning cables are no longer needed for connecting your iOS device! This PR adds a mention about USB-C cables, hooray!


Simek (approved):
Thanks for the update! 👍

PR #4219, opened 2024-09-17
Add GitHub prettier check

johnf:
Prettier wasn't being run in CI

### DIFF
--- a/website/.prettierignore
+++ b/website/.prettierignore
@@ -1,0 +1,1 @@
+src/theme
--- a/website/package.json
+++ b/website/package.json
@@ -30,9 +30,10 @@
     "lint:versioned": "eslint versioned_docs/**",
     "lint:markdown": "remark ../docs --quiet -r .remarkrc.mjs",
     "lint:markdown:versioned": "remark ./versioned_docs --quiet -r .remarkrc.mjs",
+    "lint:format": "prettier --check {{core,src}/**/*.js,*.js} ../docs/*.md {versioned_docs/**/*.md,blog/*.md} src/**/*.{scss,css}",
     "language:lint": "cd ../ && alex && case-police 'docs/*.md' -p brands,general,products,softwares -d ./website/react-native-dict.json",
     "language:lint:versioned": "cd ../ && alex . && case-police '**/*.md' -p brands,general,products,softwares -d ./website/react-native-dict.json",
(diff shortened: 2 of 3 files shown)


johnf:
@cipolleschi Turns out prettier wasn't being run in CI

cipolleschi:
Could you rebase on top of main? Your other changes have been merged, so we will isolate the changes on prettier here!

johnf:
> Could you rebase on top of main?

Done

Simek:
Prettier is run via ESLint and the preferable way would be to extend the setup to include `*.css/*.scss` files:
* https://github.com/facebook/react-native-website/blob/main/.eslintrc.json

The `theme` files are not re-formatted on purpose, since they are coming from the [Docusaurus Swizzle](https://docusaurus.io/docs/swizzling) mechanism, and unnecessary alterations makes an upgrade process more difficult.

johnf:
@Simek Are you sure that eslint supports CSS or even SCSS files?
My research leads me to believe it doesn't.

For some background, I'm implementing this because when I submitted #4218, I followed the contributor instructions to run prettier and I noticed some issues.

@cipolleschi asked me to put them in a different PR, and we discovered that GitHub was not enforcing linting on CSS files.

I'll modify this PR now to exclude the autogenerated files from prettier.

I'm also happy to follow the eslint path, but I think it takes us down a stylelint path. But maybe that's going a bit far, with styling decisions made by an external contributor.

How would you like me to proceed?

cipolleschi:
@Simek is the real expert here, so I'll defer the decision to them!

Simek:
Okay, so I think I know that it's going on here, there is a lot of confusion due to the setup we have.

The concept of linting is to let people know about the code issues. The reason why any of Prettier checks are not run directly on the CI is conceptual - we don't want to automatically edit the files from PR branch on the CI, and that's what Prettier does (it just reformat code for you), while ESLint only warns you about the problem, unless you specifically want to use to to apply changes.

That's why I have referred to updating ESLint configuration instead. It should be possible via [`eslint-plugin-css`](https://github.com/ota-meshi/eslint-plugin-css), but due to SASS usage it might be tricky.

So the way forward I recommend is just to reformat SCSS files locally via command and commit them to the repo, without setting up the `prettier`direct usage on the CI.

johnf:
@Simek It is possible to warn with prettier.
I've implemented this PR using prettier --check, this will exit with a status of 1 and list the files that have issues.
So if someone has forgotten to run prettier locally then the tests won't pass

PR #4231, opened 2024-09-26
update examples to use cross-platform SafeAreaView

Simek:
# Why & How

Update most of the `unversioned` examples to use cross-platform `SafeAreaView`, perform multiple tweaks and fixes to the existing code.

Not every evample really needed the safe area container, but  at the end for most of them, there is a value, so I have updated almost all examples to relay on `SafeAreaView`, unless there is concrete reason to not wrap the code.

# Preview

### DIFF
--- a/docs/accessibilityinfo.md
+++ b/docs/accessibilityinfo.md
@@ -9,7 +9,8 @@ Sometimes it's useful to know whether or not the device has a screen reader that
 
 ```SnackPlayer name=AccessibilityInfo%20Example&supportedPlatforms=android,ios
 import React, {useState, useEffect} from 'react';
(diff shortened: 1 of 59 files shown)


alanleedev:
@Simek @cortinico It seems the updated example code is not reflected in Deploy Preview link above. Is there some additional step that is missing?

Simek:
I have update only unversioned files, so to see the changes you need to select "Next" from the versions dropdown:
* https://deploy-preview-4231--react-native.netlify.app/docs/next/getting-started

alanleedev (approved):
Thanks for working on this and help migrating away from SafeAreaView.

PR #4235, opened 2024-09-27
Update docs related to position to account for static

joevilches:
`position: static` has existed for some time now, but no documentation on it, so I updated all of that for `position` that I could find. Besides the docs themselves, the only other thing I updated was the `package.json` for the linting examples. That was linting against RN 0.73, so I changed it to be the same as the expo version that is running on the website snippets which is 0.74.1

Screenshot 2024-09-27 at 13 00 55

Screenshot 2024-09-27 at 12 57 24

Screenshot 2024-09-27 at 12 57 17

https://github.com/user-attachments/assets/9a59a610-4777-46d5-b0db-41eb14ab0268

https://github.com/user-attachments/assets/a494afe3-8b12-400e-bcac-8397299df00c


NickGerleman:
> That was linting against RN 0.73, so I changed it to be the same as the expo version that is running on the website snippets which is 0.74.1

Any chance we could update this to the 0.76 RC? That would be needed to add the docs for boxShadow and filter.

scripts/lint-examples/package.json
```diff
@@ -20,7 +20,7 @@
     "glob-promise": "^4.2.2",
     "prettier": "2.8.8",
     "react": "18.2.0",
-    "react-native": "^0.73.0",
+    "react-native": "^0.74.1",
```
NickGerleman, line 23: Please update the other dependencies here as well. See https://react-native-community.github.io/upgrade-helper/?from=0.73.0&to=0.76.0-rc.2#RnDiffApp-package.json for the versions which have changed.

You will also need to run `yarn install` to update the lockfile and check that in.

docs/flexbox.md
```diff
@@ -2582,18 +2582,20 @@ export default WidthHeightBasics;
...
 
-- `absolute` When positioned absolutely, an element doesn't take part in the normal layout flow. It is instead laid out independent of its siblings. The position is determined based on the `top`, `right`, `bottom`, and `left` values.
+- `absolute` When positioned absolutely, an element doesn't take part in the normal layout flow. It is instead laid out independent of its siblings. The position is determined based on the `top`, `right`, `bottom`, and `left` values. These values will posistion the element relative to its containing block.
+
+- `static` When positioned statically, an element is positioned according to the normal flow of layout, and will ignore the `top`, `right`, `bottom`, and `left` values. This `position` will also cause the element to not form a containing block, unless some other supersceding style prop is present (e.g. `transform`). This allows `absolute` elements to be positioned to something that is not their parent. Note that `static` is only available on the new architecture.
```
NickGerleman, line 2593: Could we add the caveat this is only supported in new arch?

NickGerleman: This was also introduced a while enough back that we could add to the previous versioned copies of documents, but there has historically been push back on including new arch APIs before 0.76 where we enable by default.

NickGerleman (changes requested):
See above

NickGerleman:
We have some other new arch layout props that were previously being excluded from documentation, that we could also add 😅. I had a todo here for myself. https://github.com/facebook/react-native/commit/3b44182c1d5f880f3253354d16707a81d1ea66d5

scripts/lint-examples/package.json
NickGerleman: I did... but someone added a post-commit hook that broke it... https://github.com/facebook/react-native-website/blob/888a1f590a1a44389fe6d46b96b1893a86e9659c/.github/workflows/pre-merge.yml#L19

docs/flexbox.md
cipolleschi, line 2593:
```suggestion
- `static` When positioned statically, an element is positioned according to the normal flow of layout, and will ignore the `top`, `right`, `bottom`, and `left` values. This `position` will also cause the element to not form a containing block, unless some other superceding style prop is present (e.g. `transform`). This allows `absolute` elements to be positioned to something that is not their parent. Note that **`static` is only available on the New Architecture**.
```
When we call out the New Architecture, we always use capital letters :wink:

docs/flexbox.md
```diff
@@ -2861,6 +2865,24 @@ export default PositionLayout;
...
 
+## The Containing Block
+
+The containing block of an element is an ancestor element which controls its position and size.
+The way containing blocks work in React Native is very similar to [how they work on the web](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block), with some simplifications due to the lack of some web features.
+
+The `top`, `right`, `bottom`, and `left` values of an absolutely positioned element will be
+relative to its containing block.
+
+Percentage lengths like `width: '50%'` or `padding: '10%'` applied to absolutely positioned elements will be calculated relative to the size of its containing block. So for example, if the containing block is 100 points wide, then `width: 50%` on an absolutely positioned element will cause it to be
```
cipolleschi, line 2876:
```suggestion
Percentage lengths (e.g.: `width: '50%'` or `padding: '10%'`) applied to absolutely positioned elements will be calculated relatively to the size of its containing block. For example, if the containing block is 100 points wide, then `width: 50%` on an absolutely positioned element will cause it to be 50 points wide.
```

docs/flexbox.md
```diff
@@ -2861,6 +2865,24 @@ export default PositionLayout;
...
+50 points wide.
```
cipolleschi, line 2877:
```suggestion

```

Why this was on a new line? 🤔

docs/layout-props.md
```diff
@@ -905,17 +905,19 @@ Setting `paddingVertical` is like setting both of `paddingTop` and `paddingBotto
 
 ### `position`
 
-`position` in React Native is similar to regular CSS, but everything is set to `relative` by default, so `absolute` positioning is always relative to the parent.
+`position` in React Native is similar to [regular CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/position), but everything is set to `relative` by default.
 
-If you want to position a child using specific numbers of logical pixels relative to its parent, set the child to have `absolute` position.
+If you want to position a child using specific numbers of logical pixels relative to itself, set the child to have `relative` position.
```
cipolleschi, line 910: If `relative` is the default, can we just remove this sentence? 🤔 It feels a bit confusing to say that `relative` is the default and then to ask our users to set the position to `relative` to achieve a specific result result.

cipolleschi (changes requested):
Thanks for working on the docs and for taking care of these new props.
I left some suggestions but it's a great job!

cipolleschi:
@joevilches There is no need to take screenshots and videos when editing the docs.
Our automation will [deploy your diff to a staging website](https://github.com/facebook/react-native-website/pull/4235#issuecomment-2380050132), so reviewers can explore your changes themselves! ;) 
But thanks for doing that!

joevilches:
> > That was linting against RN 0.73, so I changed it to be the same as the expo version that is running on the website snippets which is 0.74.1
> 
> Any chance we could update this to the 0.76 RC? That would be needed to add the docs for boxShadow and filter.

Yeah I can. I think the expo playgrounds we use will be on 74, but its also old arch anyway so the point is moot in reference to boxShadow and filter

docs/flexbox.md
joevilches: oops, just a typo

docs/layout-props.md
joevilches: I feel we should keep the gist of the sentence there so people know what relative does. Honestly, rereading this, I think the phrasing is just weird. The "if you want... then do this". I think I should just change this to describe what each one does without any fluff.

docs/layout-props.md
```diff
@@ -907,13 +907,15 @@ Setting `paddingVertical` is like setting both of `paddingTop` and `paddingBotto
 
 `position` in React Native is similar to [regular CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/position), but everything is set to `relative` by default.
 
-If you want to position a child using specific numbers of logical pixels relative to itself, set the child to have `relative` position.
+`relative` will position an element according to the normal flow of the layout. Insets (`top`, `bottom`, `left`, `right`)
+will offset relative to this layout.
```
cipolleschi, line 911: These are on two lines, but they should not

cipolleschi (commented):
last nit!

joevilches:
@cipolleschi done!

PR #4243, opened 2024-10-04
chore: move native component and module docs to legacy

blakef:
From 0.76 the New Architecture is enabled by default. Native Component
and Native Modules are effectively legacy API, and we're giving more
prominence to Turbo Native Modules and Fabrice Native Components. The
first step is moving these under a legacy header in the sidebar.


blakef:
Fixed the redirects.

website/sidebars.json
```diff
@@ -116,7 +104,33 @@
         ]
       }
     ],
-    "New Architecture": ["the-new-architecture/landing-page"]
+    "New Architecture": ["the-new-architecture/landing-page"],
+    "Legacy": [
```
cipolleschi, line 108: "Legacy Architecture" or "Old Architecture" to stay consistent with the terminology we used in OSS

website/static/_redirects
```diff
@@ -56,6 +56,12 @@
 /docs/:version/new-architecture-troubleshooting                     https://github.com/reactwg/react-native-new-architecture#guides
 /docs/:version/new-architecture-appendix                            https://github.com/reactwg/react-native-new-architecture#guides
 
+# New Architecture is enabled by default, these have shifted to a legacy folder
+/docs/next/direct-manipulation      /docs/next/legacy/direct-manipulation
+/docs/next/local-library-setup      /docs/next/legacy/local-library-setup
+/docs/next/native-modules-*         /docs/next/legacy/native-modules-:splat
+/docs/next/native-components-*      /docs/next/legacy/native-components-:splat
```
cipolleschi, line 63: what's the difference between `-*` and `-:splat`?

docs/more-resources.md
```diff
@@ -40,4 +40,7 @@ React Native has a community of thousands of developers like you making content,
 
 Can’t find what you’re looking for in the Core Components? Visit [React Native Directory](https://reactnative.directory) to find what the community has been creating.
 
-Interested in making your own Native Component or Module? Making modules for your own use case and sharing them with others on NPM and GitHub helps grow the React Native ecosystem and community! Read the guides to making your own Native Modules ([Android](native-modules-android.md), [iOS](native-modules-ios.md)) and Native Components ([Android](native-components-android.md), [iOS](native-components-ios.md)).
+:::caution
```
cipolleschi, line 43: shouldn't we refer to the new guides to TM and Fabric components once we have them, rather than pointing the users to the old legacy docs?

docs/animations.md
```diff
@@ -671,7 +671,7 @@ This example uses a preset value, you can customize the animations as you need,
 
 ### `setNativeProps`
 
-As mentioned [in the Direct Manipulation section](direct-manipulation), `setNativeProps` allows us to modify properties of native-backed components (components that are actually backed by native views, unlike composite components) directly, without having to `setState` and re-render the component hierarchy.
+As mentioned [in the Direct Manipulation section](legacy/direct-manipulation), `setNativeProps` allows us to modify properties of native-backed components (components that are actually backed by native views, unlike composite components) directly, without having to `setState` and re-render the component hierarchy.
```
cipolleschi, line 674: It's likely that the direct manipulation section won't change much, tbh. but let's keep it this way for now.

cipolleschi (changes requested):
Good job. I left a couple of suggestions but they should be quick.

docs/more-resources.md
blakef: I've left this `::caution` more as a way to signal to us that this needs additional work.  My preferences is to move these docs under `legacy/` now then to update as we introduce the new docs.

website/sidebars.json
blakef: I like that.

cipolleschi (approved):

PR #4255, opened 2024-10-08
Remove iOS tag from `snapToAlignment` of `ScrollView` prop

fabriziocucci:
`snapToAlignment` is actually supported on Android.


cortinico (approved):
@fabriziocucci could you backport this down to the versioned docs till 0.72 ?

PR #4262, opened 2024-10-09
feat: respect users' preferred color schema

jewhyena:
These are the changes to the [Docusaurus theme config](https://docusaurus.io/docs/api/themes/configuration#color-mode---dark-mode) so that it prefers the user's system color scheme over the default light one.

I left the light mode as the default one for cases when system preferences are unavailable to be consistent with the official [react.dev](https://react.dev/) documentation ([see this line of code](https://github.com/reactjs/ru.react.dev/blob/main/src/styles/index.css#L384)).

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -212,6 +212,11 @@ module.exports = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     ({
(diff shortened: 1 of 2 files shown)


Simek (approved):
Thanks for this improvement! Tested the changes on the preview, everything works as described, LGTM! 👍

PR #4267, opened 2024-10-10
Fix MDX double `<p>` tags

slorber:
Docusaurus v3.6 (https://github.com/facebook/react-native-website/pull/4268) will have a new HTML minifier. 

It has the ability to report some markup errors, such as `<p>` inside `<p>` which is not valid HTML.

We probably missed that during the MDX v2+ upgrade, but the usage of `<p>` + line breaks lead to duplicate paragraphs. The solution is to either use fragments or remove the line breaks.

See MDX playground:

CleanShot 2024-10-10 at 12 39 56

cc @Simek this shouldn't have any meaningful impact of the site but fixes the markup and will avoid printing another upcoming build warning.

### DIFF
--- a/docs/appregistry.md
+++ b/docs/appregistry.md
@@ -5,9 +5,7 @@ title: AppRegistry
 
 <div className="banner-native-code-required">
   <h3>Project with Native Code Required</h3>
-  <p>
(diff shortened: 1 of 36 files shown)


Simek (approved):
Tested the changes on the preview, everything looks like expected, thanks for addressing that Seb! 👍

PR #4272, opened 2024-10-11
docs: update `position` documentation, types (add `static`, explain)

sorenfrederiksen:
# Docs ✍️

While reading the `position` style prop documentation in Layout Props, I noticed the "Type" specified (`enum('absolute', 'relative')`) did not match that in the React Native implementation (`"absolute" | "relative" | "static" | undefined`).

Screenshot 2024-10-11 at 07 51 32

In particular:

- The `static` position prop was missing.
- As a result, the documentation was out of date and incorrect.

To fix this, I've:

- Fixed the `enum` type specified.
- Updated the prop's explanation to reflect the current implementation, basing this on more or less entirely on [the Yoga Layout Position documentation](https://www.yogalayout.dev/docs/styling/position) and its subdocuments.

## Before 🐛

- `static` position missing.
- Documentation incomplete and, in a couple instances, inaccurate.

## After 🦋

- `static` position added.
- Documentation updated to reflect current Yoga Layout and React Native implementations.

------

Please take and edit this contribution as you see fit. Hope it's of some use 🙏

### DIFF
--- a/docs/layout-props.md
+++ b/docs/layout-props.md
@@ -887,9 +887,9 @@ When direction is `ltr`, `paddingStart` is equivalent to `paddingLeft`. When dir
 
 `paddingTop` works like `padding-top` in CSS. See https://developer.mozilla.org/en-US/docs/Web/CSS/padding-top for more details.
 
-| Type            | Required |
-| --------------- | -------- |
-| number, ,string | No       |
+| Type           | Required |
+| -------------- | -------- |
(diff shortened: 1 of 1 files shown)


sorenfrederiksen:
Oh - just noticed this has already been updated by someone else: https://github.com/facebook/react-native-website/pull/4235

Lovely! Closing 🫡

PR #4273, opened 2024-10-11
Add borderRadius Percentage description

Faisalmajeed123:
#4264

PR #4276, opened 2024-10-12
Add borderRadius Percentage description #4273

Faisalmajeed123:
@Simek reverted the yarn.lock file

#4273

### DIFF
--- a/docs/view-style-props.md
+++ b/docs/view-style-props.md
@@ -195,7 +195,7 @@ On iOS 13+, it is possible to change the corner curve of borders.
 
 ### `borderRadius`
 
-If the rounded border is not visible, try applying `overflow: 'hidden'` as well.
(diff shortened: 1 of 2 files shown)


docs/view-style-props.md
```diff
@@ -195,7 +195,7 @@ On iOS 13+, it is possible to change the corner curve of borders.
 
 ### `borderRadius`
 
-If the rounded border is not visible, try applying `overflow: 'hidden'` as well.
+You can set a borderRadius using absolute numbers or a percentage. If the rounded border is not visible, try applying `overflow: 'hidden'` as well.
```
Simek, line 198: Thanks for reverting lock changes! 👍 

```suggestion
You can set `borderRadius` using absolute numbers or a percentage value string. If the rounded border is not visible, try applying `overflow: 'hidden'` as well.
```

Nit, clarification for percentage value.

PR #4277, opened 2024-10-12
docs: measureLayout() with native node handler is obsolete when using new architecture

LunatiqueCoder:
```
 (NOBRIDGE) LOG  Bridgeless mode is enabled
 (NOBRIDGE) ERROR  Warning: ref.measureLayout must be called with a ref to a native component.
```

`measureLayout()` with `relativeToNativeNode` handler (instead of reference) seems to be **OBSOLETE** when using the new architecture.


javache:
Can you update this in the latest docs instead of the 0.75 version?

LunatiqueCoder:
Added for 0.76. Let me know if I should remove the `version-0.75` change as well.

Abbondanzo:
Keeping the `version-0.75` is great. Could you make the same changes to versioned docs for 0.76 too?

LunatiqueCoder:
@Abbondanzo Hmm, so weird. Pressing the "Edit this page"on the docs seems to be leading me to the wrong place on 0.76.

Updated, hope everything's ok now.

Abbondanzo:
> @Abbondanzo Hmm, so weird. Pressing the "Edit this page"on the docs seems to be leading me to the wrong place on 0.76.
> 
> Updated, hope everything's ok now.

Good callout, the "Edit this page" button will only point to the latest version of the docs and not the version you're looking at, but we can update that. Thanks for making these last changes so quickly!

PR #4288, opened 2024-10-21
Add Fabric Native Component Guide

blakef:
Added the docs for Fabric Native Components, updated to be compatible with 0.76.


docs/fabric-native-components.md
```diff
@@ -0,0 +1,217 @@
...
+
+export default (codegenNativeComponent<NativeProps>(
+  'RTNCenteredText',
+): HostComponent<NativeProps>);
+```
+
+</TabItem>
+</Tabs>
+
+Add an `index.js` to wrap your
```
cipolleschi, line 102:
```suggestion

```

docs/fabric-native-components.md
```diff
@@ -0,0 +1,217 @@
...
+Use this specification for our Text Component:
+
+<Tabs groupId="language" queryString defaultValue={constants.defaultJavaScriptSpecLanguage} values={constants.javaScriptSpecLanguages}>
+<TabItem value="typescript">
+
+The specification file must be named `<MODULE_NAME>NativeComponent.{ts,tsx}` to work with Codegen:
+
+```typescript title="RTNCenteredText/js/RTNCenteredTextNativeComponent.ts"
+import type {ViewProps} from 'react-native';
+import type {HostComponent} from 'react-native/Libraries/Renderer/shims/ReactNativeTypes';
```
cipolleschi, line 60: this is not used in the spec, no?

docs/fabric-native-components.md
```diff
@@ -0,0 +1,217 @@
...
+
+</TabItem>
+<TabItem value="flow">
+The specification file must be named `<MODULE_NAME>NativeComponent.{js,jsx}` to work with Codegen:
+
+```flow title="RTNCenteredText/js/RTNCenteredTextNativeComponent.js":
+// @flow strict-local
+
+import type {ViewProps} from 'react-native/Libraries/Components/View/ViewPropTypes';
+import type {HostComponent} from 'react-native';
```
cipolleschi, line 83: this is not used in the specs, no?

docs/fabric-native-components.md
```diff
@@ -0,0 +1,217 @@
...
+</TabItem>
+</Tabs>
+
+Add an `index.js` to wrap your
+
+As with Turbo Native Modules, you're able to have multiple specification files in the `js/` directory. For more information about the types you can use, and the platform types these map to see the [appendix](appendix.md#codegen-typings).
+
+## 2. Configure the Component for Codegen and Auto-Linking
+
+Typically Fabric Nactive Components are distributed using npm, it's also how our Codegen is configured. Add a [package.json](https://docs.npmjs.com/cli/v10/configuring-npm/package-json) to `RTNCenteredText/package.json`:
```
cipolleschi, line 108: For consistency, we should use the same approach we used for TM. We should create them inside the app and then move them outside in a library.

docs/fabric-native-components-android.md
```diff
@@ -0,0 +1,439 @@
+---
+id: fabric-native-components-android
+title: 'Fabric Native Modules: Android'
+---
+
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
+
+React Native uses Gradle to manage dependencies, auto-linking and codegen. To use this, we're going to add the `RTNCenteredText` Fabric Native Component to our `Demo` application.
+
+### 1. Link RTNCenteredText as an npm dependency:
```
cipolleschi, line 10: this step should be removed. For consistency, we should show how to create a Fabric Component as part of the application

docs/fabric-native-components-android.md
```diff
@@ -0,0 +1,439 @@
...
+```bash
+cd RTNCenteredText
+yarn link
+cd ../Demo
+yarn link rtn-centered-text
+```
+</TabItem>
+</Tabs>
+
+We also need to help [Metro](https://metrobundler.dev/docs/configuration/#watchfolders) understand where to find packages that are symlinked through the `node_modules`:
```
cipolleschi, line 31: If the fabric component is part of the application, we can remove this step.

docs/fabric-native-components-android.md
```diff
@@ -0,0 +1,439 @@
...
+                    └── CenteredTextPackage.{kt,java}
+```
+
+Create your directories:
+
+```bash
+mkdir -p android/src/main/java/com/rtncenteredtext
+```
+
+#### Add build.gradle
```
cipolleschi, line 88: if the component is part of the application, this can be removed

docs/fabric-native-components-android.md
```diff
@@ -0,0 +1,439 @@
...
+
+<Tabs groupId="android-language" queryString defaultValue={constants.defaultAndroidLanguage} values={constants.androidLanguages}>
+<TabItem value="java">
```
cipolleschi, line 91: I think we can have solo one of the two gradle files. Either in groovy or in kotlin. I believe that the default is kotlin, but we can remove this step easily.

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,258 @@
+---
+id: fabric-native-components-ios
+title: 'Fabric Native Components: iOS'
+---
+
+import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
+
+React Native uses CocoaPods to manage iOS dependencies, auto-linking and codegen. You will need to add a `.podspec`.
+
+## Create your Pod
```
cipolleschi, line 10: we should not ask the users to create a podspec for a component in app.

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,258 @@
...
+end
+```
+
+You'll notice that this is using the information we entered in the `package.json`, to keep that as the single source of truth.
+
+## Using RTNCenteredText to your Application
+
+You can [manually run](the-new-architecture/codegen-cli) the Codegen, however it's simpler to use the application you're going to demo the component in to do this for you.
+
+### 1. Link RTNCenteredText as an npm dependency:
```
cipolleschi, line 40: this step can be avoided if the component is written in app.

cipolleschi (changes requested):
Thanks for putting this together. This is a good starting pint, but we should not suggest the users to write the components as separate libraries. Instead, we should do the same thing we did for turbomodules and let them create them in app.

Also, there is an empty `_fabric-native-component-js.md` file that I can't understand what it is... Can we remove it?

blakef:
Sure I'll redraft this. Thanks for taking the time to go through all of it @cipolleschi

docs/fabric-native-components-android.md
```diff
@@ -0,0 +1,391 @@
...
+      @Override
+      public Map<String, ReactModuleInfo> get() {
+        Map<String, ReactModuleInfo> map = new HashMap<>();
+        map.put(CenteredTextManager.REACT_CLASS, new ReactModuleInfo(
+                CenteredTextManager.REACT_CLASS,
+                CenteredTextManager.REACT_CLASS,
+                false,
+                false,
+                false,
+                BuildConfig.IS_NEW_ARCHITECTURE_ENABLED
```
cipolleschi, line 342: perhaps we should add a comment with what these parameters are: why do we have two REACT_CLASS? what are the three `false`s?

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,184 @@
...
+[Codegen] Searching for podspec in the project dependencies.
+[Codegen] Supported Apple platforms: ios for RTNCenteredTextSpecs
+[Codegen] Generating Native Code for FBReactNativeSpec - ios
+// highlight-next-line
+[Codegen] Generated artifacts: /Users/me/Demo/ios/build/generated/ios
+```
+
+:::warning
+You should be careful about committing this generated code to your repository. The generated code is specific to each version of React Native. Use npm [peerDependencies](https://nodejs.org/en/blog/npm/peer-dependencies) to restrict compatibility with version of React Native.
+:::
```
cipolleschi, line 36: this can probably be removed: the gitignore already ignores the `ios/build` folder where the code is generated by Codegen.

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,184 @@
...
+### 2. Write your Native Platform Code
+
+At the end of this, you'll have the following iOS code:
+
+```
+ios
+├── RTNCenteredText.h
+├── RTNCenteredText.mm
+└── RTNCenteredTextManager.mm
+```
```
cipolleschi, line 47: we should mention that these files needs to be created in Xcode, as we did in the previous guide for TM, otherwise they are not part of the project and they will not be picked up.

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,184 @@
...
+// This macro lets Fabric retrieve and instantiate your module
+// highlight-next-line
+RCT_EXPORT_MODULE(RTNCenteredText)
+
+// This macro exposes the text property of the component. Note it exposes the name and type.
+// highlight-next-line
+RCT_EXPORT_VIEW_PROPERTY(text, NSString)
+
+@end
+```
```
cipolleschi, line 72: This is not required in a pure New Architecture component. However, in this case, we has to modify the AppDelegate to manually register the component.

docs/fabric-native-components-ios.md
```diff
@@ -0,0 +1,184 @@
...
+  [super updateProps:props oldProps:oldProps];
+}
+
+@end
+
+Class<RCTComponentViewProtocol> RTNCenteredTextCls(void)
+{
+  return RTNCenteredText.class;
+}
+```
```
cipolleschi, line 184: we should explain what's going on here.

docs/fabric-native-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+<Tabs groupId="language" queryString defaultValue={constants.defaultJavaScriptSpecLanguage} values={constants.javaScriptSpecLanguages}>
+<TabItem value="typescript">
+
+The specification file must be named `<MODULE_NAME>NativeComponent.{ts,tsx}` to work with Codegen:
+
+```typescript title="Demo/js/RTNCenteredTextNativeComponent.ts"
+import type {ViewProps} from 'react-native';
+import type {HostComponent} from 'react-native/Libraries/Renderer/shims/ReactNativeTypes';
+import type {Int32} from 'react-native/libraries/Types/CodegenTypes';
```
cipolleschi, line 56: this is not used in the specs

docs/fabric-native-components.md
```diff
@@ -0,0 +1,188 @@
...
+
+import type {ViewProps} from 'react-native/Libraries/Components/View/ViewPropTypes';
+import type {HostComponent} from 'react-native';
+import type {Int32} from 'react-native/libraries/Types/CodegenTypes';
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+
+type NativeProps = $ReadOnly<{|
+  ...ViewProps,
+  text: ?string,
+  margin?: Int32,
```
cipolleschi, line 84: the flow specs does not have the margin anymore. I think we should remove them from here as well and we should also remove the import of the Int32

cipolleschi (changes requested):

cipolleschi:
IIRC, we also agreed with expo to use a webview as an Example, like they do here: https://docs.expo.dev/modules/native-view-tutorial/

I think that it is a matter of updating the examples, basically all the content is there, with the TM structure and the the text already written. We should only harmonize them and update the examples.

I can work on this today while @blakef is traveling.

cipolleschi (approved):

PR #4185, opened 2024-08-19
Update 2024-08-12-release-0.75.md

KingAmo:
fix  https://github.com/facebook/react-native/issues/46089

PR #4299, opened 2024-10-24
Update 2024-10-23-release-0.76-new-architecture.md

KingAmo:
fix 
```
npm error arg Argument starts with non-ascii dash, this is probably invalid: –version
```

just the same typo as  https://github.com/facebook/react-native-website/pull/4185


Simek (approved):
Nice catch, thanks for the correction! 🙏

PR #4333, opened 2024-11-01
Rename `logo-{og => share}.png` to Invalidate Cache

yungsters:
Meta's image caching is not being invalidated after https://github.com/facebook/react-native-website/pull/4332, so let's try this.


yungsters:
This can be reverted, probably.

Alternatively, we could delete `logo-og.png` and update all the older versions of the docs.

PR #4352, opened 2024-11-12
Add missing listitem role do accessibiltiy doc

6TELOIV:
The accessibility doc seemed to be missing info about the `listitem` role, which is an important complementary role to the `list` role.

I added an entry for it with a simple description of the purpose.

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -363,6 +363,7 @@ Assign this property to a custom function which will be called when someone perf
 - **img** Used when the element should be treated as an image. Can be combined with a button or link, for example.
 - **link** Used when the element should be treated as a link.
 - **list** Used to identify a list of items.
+- **listitem** Used to itentify an item in a list.
 - **menu** Used when the component is a menu of choices.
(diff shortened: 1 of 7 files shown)


6TELOIV:
Preview link seems to not have the change; not sure if I missed a step!

6TELOIV:
Found my change; visible at https://deploy-preview-4352--react-native.netlify.app/docs/next/accessibility#role (didn't realize it'd be under /next. It should apply to the current version as well, as it's an existing and supported feature, and is just missing from the docs)

Simek:
Hello @6TELOIV, thanks for the update! 👍 

The versioned doc files are located in the directory linked below. Would be great if you can backport that change.

Also, would be nice to provide the context (PR/commit) when this role has been added, so we would know how far this change needs to be backported.

6TELOIV:
@Simek From what I can tell, [commit a50e6fb](https://github.com/facebook/react-native/commit/a50e6fb341eacd955071e4a7c6d734e7d3c94cd7) added the listitem role. That would be v0.71.0. I'll update the pull request.

PR #4353, opened 2024-11-14
added color code.md

bhuvabhavik:
added color code for better visibility.
- The `javascript` after the first backticks tells the site to use JavaScript syntax highlighting.


bhuvabhavik:
added color code for better visibility.

PR #4355, opened 2024-11-17
Fix the build error in the example code

foyoodo:


cipolleschi (approved):
Thanks for spotting this.

PR #4361, opened 2024-11-19
Fix code example in "Tracking dynamic values"

Egor-Kozlov:
In the docs this code uses useNativeDriver in the a wrong position


Simek (approved):
Hello @Egor-Kozlov, nice catch, thank you for the fix and backport! 👍

Simek:
@Egor-Kozlov Can you fix the lint issues?

PR #4362, opened 2024-11-20
Update Android profiling instructions

gaearon:
I couldn't get any of this to work last time I tried. I assumed the entire page is completely broken and outdated. Turns out, all of it is still relevant, it's just recommending tools that haven't worked for many years. This changes the recommendation to tools that actually work.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -5,15 +5,13 @@ title: Profiling
 
 Profiling is the process of analyzing an app's performance, resource usage, and behavior to identify potential bottlenecks or inefficiencies. It's worth making use of profiling tools to ensure your app works smoothly across different devices and conditions.
 
-For iOS, Instruments is an invaluable tool, and on Android you should learn to use [`systrace`](profiling.md#profiling-android-ui-performance-with-systrace).
-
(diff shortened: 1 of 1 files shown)


docs/profiling.md
```diff
@@ -5,53 +5,33 @@ title: Profiling
 
 Profiling is the process of analyzing an app's performance, resource usage, and behavior to identify potential bottlenecks or inefficiencies. It's worth making use of profiling tools to ensure your app works smoothly across different devices and conditions.
 
-For iOS, Instruments is an invaluable tool, and on Android you should learn to use [`systrace`](profiling.md#profiling-android-ui-performance-with-systrace).
-
```
javache, line 9: Can you keep the iOS reference?

javache:
Thanks for this! We're looking to migrate our in-depth tracing markers to be based on perfetto (we've switched to this workflow at Meta) and will look to make this part of the open-source tooling too.

gaearon:
Addressed + added info about method profiling

blakef:
There's no reason that build should be failing, just tried it locally and it built fine...

cipolleschi:
Linter is failing with:
Screenshot 2024-11-22 at 10 58 03

gaearon:
lmao "it's easy to get confused" is not the meaning of "easy" the linter was designed to prevent :D

PR #4364, opened 2024-11-21
collapsable is not Android only prop in new architecture

sammy-SC:
Before:
Screenshot 2024-11-21 at 14 45 05

After:
Screenshot 2024-11-21 at 14 45 00


cipolleschi (approved):
Thanks @sammy-SC for fixing this!

PR #4365, opened 2024-11-21
add collapsableChildren to docs and add default value for collapsable

sammy-SC:
[collapsableChildren](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/View/ViewPropTypes.js#L582) is not documented on the website. Let's add it.

In this PR, I also add default value column to `collapsable`.

Screenshot 2024-11-21 at 14 55 27


cipolleschi (approved):
Amazing, thank you!

PR #4371, opened 2024-11-23
Adjust the "Edit this page" button to always point to versioned docs

Abbondanzo:
Per [a recent comment here](https://github.com/facebook/react-native-website/pull/4277#issuecomment-2495582155), it was pointed out that the "Edit this page" button in the footer of all our pages do not actually point at the versioned docs but at the latest copy of the docs. 

This PR changes the URL to match the respective doc location, not the most recent copy of docs. So, if I'm looking at docs for 0.76, the edit button will link me to the exact 0.76 copies in the repository. 

I am hesitant to merge it because it might be more desirable to point people at latest by default. It will also result in broken URLs when the older versions are archived and no longer in the repository.


yungsters:
Hmm, yeah… I'm not sure that this is ideal.

I think there's certainly a problem of helping contributors understand documentation versioning, but I think most people would reasonably expect that their changes will make it into at least next and all future versions of the docs.

They probably also want it to be in the currently viewed version (most commonly, the latest versioned docs).

What if we had two buttons? One that leads to the current version, and one the leads to master / next?

Abbondanzo:
>What if we had two buttons? One that leads to the current version, and one the leads to master / next?

We have a custom footer component swizzled in there and I wonder if we can return an array of URLs from the `editUrl` function. At worst, we have piles of context we can tap into to build both URLs. I'll give this a try tomorrow if it sounds reasonable. The naming portion is hard, what do you think?

[Edit this page] [Edit latest page]

[Edit this version] [Edit latest version]

[Edit version XYZ] [Edit latest version]

yungsters:
Hmm… if we can determine the current version being viewed (ideally without parsing the current URL, but including that as an option), maybe we could do something like…

- If viewing next (master)…
  - Edit page for next release
  - (Just one button.)
- If viewing latest version (latest versioned)…
  - Edit page for next release
  - Edit page for current release
- If viewing an older version…
  - Edit page for next release
  - Edit page for 0.XY release

Maybe something like that? It's more work, so of course don't feel obligated to do all of this. We can still make incremental improvements. 😁

slorber:
I guess I should add support for edit url arrays in Docusaurus 😄 

Already wanted to add support for multiple edit variants (GitHub, stackblitz, codesandbox), will think about it. 

I thought about presenting the edit options in some kind of modal, possibly explaining the tradeoffs of each option. In your case, it probably doesn't make sense though, but you could still disable that modal through swizzle.

But your use-case also makes sense and probably should be the default for versioned docs sites? 🤔

slorber:
That's true. The edit url is a plugin-specific feature (not Docusaurus core) and each plugin can have different opinions (docs has versioning for example). Supporting arrays in docs/blog might be a good first step, but I'll need to design the feature a bit more for docs.

hichemfantar:
it seems that the cause in my pr was upgrading to eslint-plugin-mdx v3

hichemfantar:
actually nevermind it seems the latest commit in this pr was ran after my pr was merged so the cause is 100% https://github.com/facebook/react-native-website/pull/4384

like i said previously, i think the rule started working in my pr but i'm investigating

Simek:
Just a small nit, otherwise looks great! 👍 

Can we change the button label in non-versioned docs to be just "Edit this page"?

Screenshot 2025-02-01 at 13 48 34

PR #4383, opened 2024-12-03
Enhance .gitignore files for better exclusion of unnecessary files

hichemfantar:
Update .gitignore to include additional logs, editor files, and generated artifacts, improving project cleanliness and reducing clutter in version control.


.gitignore
```diff
@@ -1,13 +1,38 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
```
Simek, line 8: We should only ignore the files from package manager supported in repo, adding other manager files can obsuscate an user issue.

.gitignore
```diff
@@ -1,13 +1,38 @@
...
+!.vscode/extensions.json
 .idea
-.nvmrc
-.docusaurus
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
```
Simek, line 22: Those entries are totally unrelated to the project.

.gitignore
```diff
@@ -1,13 +1,38 @@
...
-sync-api-docs/generatedComponentApiDocs.js
-sync-api-docs/extracted.json
+# https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
Simek, line 33: This part is for Yarn Berry, it should be added in the PR migrating to it.

website/.gitignore
```diff
@@ -0,0 +1,20 @@
...
+# Misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
Simek, line 20: We should stick to one global gitignore at monorepo root, since we don't support sparse checkout anyway.

Simek (commented):
In general, I would prefer us to stay with manually managed `.gitignore`, and only add entries in there if any issue arose. 

Would be nice if you can chop down the changeset down to only necessary changes that would be great.

.gitignore
hichemfantar: included these from the vite template but sure i'll get rid of em

hichemfantar: on second thought, those are actually node config files for visual studio so they should be ignored

hichemfantar: for reference https://github.com/vitejs/vite/blob/main/packages/create-vite/template-react-ts/_gitignore

.gitignore
```diff
@@ -1,13 +1,31 @@
...
+yarn-error.log*
+
 node_modules
-.DS_Store
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+!.vscode/settings.json
```
Simek, line 13: Last nit, there is no need to store any VSCode related files since we do not recommend any specific IDE, you can use whatever you want, so let's ignore all editor files:
```suggestion
```

PR #4384, opened 2024-12-04
Update Prettier to v3 and eslint mdx to support mdx 3

hichemfantar:
Upgrade Prettier and related ESLint plugins to their latest versions for improved formatting and linting capabilities.


Abbondanzo (approved):

hichemfantar:
@Abbondanzo it seems the dead external url rule is causing very long execution times, should we keep or get rid of the rule?

i think it started actually properly working after the mdx plugin was upgraded

Abbondanzo:
> @Abbondanzo it seems the dead external url rule is causing very long execution times, should we keep or get rid of the rule?
> 
> i think it started actually properly working after the mdx plugin was upgraded

Let's disable the rule for now, we can choose to opt in smaller subsets of folders (like current docs only) if we want and make full passes selectively

hichemfantar:
alright then let's discuss this at https://github.com/facebook/react-native-website/pull/4392

PR #4399, opened 2024-12-13
Fix default value of removeClippedSubviews on Android

javache:
Align docs with https://reactnative.dev/docs/flatlist#removeclippedsubviews


Simek:
Hey @javache, thanks for the update! 🙏 

After looking at code on `main` I have spotted that this has been correct for few last versions, so I have backported this to all version currently in main docs, and also added the wording changed you have made.

Closing in favour of:
* #4846

PR #4411, opened 2025-01-02
Update border radius styles for consistency improvements

hichemfantar:
Enhance the theme styles to be more consistent with the rest of the website.

before
image

after
image

before
image

after
image

before
image

after
image

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -39,6 +39,7 @@
   --ifm-color-warning: #ffe564;
   --ifm-alert-color: var(--ifm-font-color-base);
   --ifm-pagination-nav-color-hover: var(--ifm-color-emphasis-300);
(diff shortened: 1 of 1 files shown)


cortinico (approved):

Abbondanzo:
My bad, saw the review request after pressing merge

Simek:
@hichemfantar I personally think that sindle-sided rounded border does not look great with radius, that why GitHub also does not use borders directly for admonitions.

* https://github.com/orgs/community/discussions/16925

I would change the styling to include full border around box, like info box, or even better, align the callout/quote boxes with what we have on react.dev:
* https://react.dev/reference/rules/components-and-hooks-must-be-pure

hichemfantar:
is this close to what you had in mind?

before
image

after
image

PR #4413, opened 2025-01-02
Make table border thicker and easier to see

hichemfantar:
Use the default width and eliminate the redundant table border width setting in the custom theme to streamline the styling.

before
image

after
image

before
image

after
image

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -40,7 +40,6 @@
   --ifm-alert-border-radius: 0;
   --ifm-alert-color: var(--ifm-font-color-base);
   --ifm-pagination-nav-color-hover: var(--ifm-color-emphasis-300);
-  --ifm-table-border-width: 0.005rem;
   --ifm-navbar-sidebar-width: 100%;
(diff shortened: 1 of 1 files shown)


Abbondanzo (approved):

Simek:
@hichemfantar Can you elaborate way this change has been needed, and what exact it improves/fixes?

This is a CSS trick for HDPI displays, and should not affect regular pixel density appearance.

hichemfantar:
I wasn’t aware that’s why you set this.
I do notice a difference in border clarity with this change but feel free to change this back or perhaps make the change upstream

Abbondanzo:
This improves contrast on the border between table cells and is a welcome change in my opinion. On mobile devices, the border isn't visible at all. Is the intention of the original CSS to remove borders?

|Before|After|
|-|-|
|image|image|

Simek:
Lack of borders on mobile device was definitely not intended, maybe there was other unrelated to RN website changes in Docusaurus or Infima, which caused that appearance. 

The situations like this show why we need visual diff CI runs when upgrading Docusaurus, and throw testing. I wonder if Argos did not catch that, or we just did not run it for mobile viewport, will follow up on that with Sebastien.

PR #4418, opened 2025-01-03
Add codeblock language as title remark plugin

hichemfantar:
Introduce a new remark plugin that sets the title metadata for code blocks based on their language, along with updates to dependencies and README documentation for clarity and consistency.

A custom [remark](https://github.com/remarkjs/remark) plugin that automatically sets the `title` metadata for code blocks in Markdown files based on the block's language.

## Features

- **Adds `title` Metadata**: For code blocks with a specified language (e.g., `js`, `python`), the plugin sets the `title` metadata to match the language.
- **Preserves Existing Metadata**: If the code block already contains metadata (`meta`) but lacks a `title`, the plugin appends the `title` without removing the existing metadata.
- **Ignores Code Blocks with Existing `title`**: If the `meta` field already includes a `title` property, the plugin does nothing to avoid overwriting it.

read more at https://github.com/facebook/docusaurus/issues/10808

before
image

after
image

before
image

after
image

### DIFF
--- a/plugins/remark-codeblock-language-as-title/README.md
+++ b/plugins/remark-codeblock-language-as-title/README.md
@@ -1,0 +1,9 @@
+# languageAsTitleRemarkPlugin
+
+A custom [remark](https://github.com/remarkjs/remark) plugin that automatically sets the `title` metadata for code blocks in Markdown files based on the block's language.
+
+## Features
+
+- **Adds `title` Metadata**: For code blocks with a specified language (e.g., `js`, `python`), the plugin sets the `title` metadata to match the language.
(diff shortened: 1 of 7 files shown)


Abbondanzo (approved):

Simek:
@hichemfantar I think that it's a nice feature, but we should not add language as a code block title by default, the title should be fully controlled by author, and only added when specified.

This approach usually create a noise, and does not add that much information/context for the users, also stretching the page in length. You can see that we don't use that pattern on [react.dev](https://react.dev/) nor [docs.expo.dev](https://docs.expo.dev/).

hichemfantar:
I personally find this very helpful but I could make it more presentable by mapping the language to a display name as shown here https://github.com/facebook/docusaurus/issues/10808#issuecomment-2569163639

PR #4420, opened 2025-01-04
Change default language from 'jsx' to 'tsx' in Docusaurus config

hichemfantar:
Update the default language setting in the Docusaurus configuration to take into account snippets with TypeScript.


Simek (approved):
Looks good! 👍 Confirmed on the preview that there are no regression in highlight due to the change.

PR #4421, opened 2025-01-04
Style updates for content banner

hichemfantar:
Add a border radius to the content banner and fix its width for improved layout consistency.


Simek (approved):
The idea behind the left offset is to align content not the wrapper/decoration, which is also reasonable. Don't have strong opinion on that one, but in general would be nice to have a bit more context, and the details on the changes you decide to make.

hichemfantar:
In this case, I just wanted it to look like the other banners in the theme

PR #4422, opened 2025-01-04
Remove unused scrollbar styles from themes

hichemfantar:
Eliminate unnecessary scrollbar styles from both shared and custom themes to streamline the codebase.

### DIFF
--- a/website/src/css/_shared.scss
+++ b/website/src/css/_shared.scss
@@ -50,30 +50,3 @@
     color: var(--ifm-link-color);
   }
 }
-
-%scrollbar-style {
-  &::-webkit-scrollbar {
-    width: 7px;
-    height: 7px;
-  }
-
-  &::-webkit-scrollbar-thumb {
-    background: #888;
-    border-radius: 10px;
-  }
-
-  &::-webkit-scrollbar-track {
-    background: #f1f1f1;
-    border-radius: 10px;
-  }
-}
-
-%scrollbar-style-dark {
-  &::-webkit-scrollbar-track {
-    background: #141414;
-  }
-
-  &::-webkit-scrollbar-thumb {
-    background: var(--ifm-color-emphasis-200);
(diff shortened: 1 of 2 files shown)


Simek:
I think that those scrollbars are used and don't see anything wrong with them. In contrast, I think that they offer more control and flexibility for the developers, allowing to unify the overall appearance of website, no matter which platform or browser user uses..

Can you explain few points from you take?

> streamline the codebase
What exactly you have in mind by that? Does the scrollbars hinder or negatively affect the code base?

> they seem to only work partially on codeblocks
Have you tried to fix the spotted issue? Can you elaborate a bit more when they work and when not, and what is an expected behavior in your eyes?

> it's always better to rely on native scrollbars as they behave more closely to user expectations
Can you explain a bit more that statement? I would say that the functionally the scrollbars are the same, no matter how they styled. What are the resources or stats you are basing your observations about users expectations on?

hichemfantar:
I checked the places where the custom scrollbar is being applied in the css (main scrollbar, main sidenav, blog sidebar, table of contents, codeblock) and they all use the native scrollbar. I believe these custom styles are from the v1 to v2 migration and are now broken and are being overwritten by docusaurus.

> > it's always better to rely on native scrollbars as they behave more closely to user expectations
Well this is from personal experience and other devs i worked with agreed but here's a collection of forums where the matter was heavily discussed.
https://www.reddit.com/r/webdev/comments/1bhhu7k/scrollbar_customization/
https://news.ycombinator.com/item?id=36302002
https://www.reddit.com/r/web_design/comments/akmnq5/why_do_developers_feel_the_need_to_reinvent_and/
https://www.reddit.com/r/webdev/comments/onz3a6/whats_up_with_styling_scrollbars/

hichemfantar:
@Simek I added a screenshot for visual proof in the original post

PR #4424, opened 2025-01-05
Add lazy loading to custom component images for performance improvements

hichemfantar:
Implement lazy loading for images across various components to enhance performance and reduce initial load times.

### DIFF
--- a/website/src/pages/Home/Community/index.tsx
+++ b/website/src/pages/Home/Community/index.tsx
@@ -81,10 +81,10 @@ function Community() {
                 <li key={i} className="item">
                   {app.infoLink ? (
                     <a href={app.infoLink}>
-                      <img src={imgSource} alt={app.name} />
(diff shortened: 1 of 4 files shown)


Simek (commented):
Adding `loading="lazy"` is only recommended for images placed not in the initial user viewport, otherwise lazy loading could lead to worse feeling for the users, see:
* https://www.w3schools.com/tags/att_img_loading.asp

For most altered components we cannot guarantee they will be below the fold.

Also, the Firefox has a know issue where the order of attributes matter, and `loading` need to be placed before `src` attribute, and the prop it itself have some requirements which need to be met which can also slow down the total time website takes to load:
* https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading#usage_notes

For now, I'm leaning into leaving the code as-is, but if you can provide a performance analysis which will show that there is a meaningful benefit thanks to the changes, it would be a great way to win me over in this case!

PR #4431, opened 2025-01-08
Better contrast for component grid component header

hichemfantar:
Change the background color of the component grid to enhance the contrast.

before
image

image

after
image

image

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1707,8 +1707,8 @@ article .component-grid {
       font-weight: 600;
       margin: 0;
       padding: 0 10px;
-      background-color: #357da1;
-      color: white;
+      background-color: var(--home-button-primary);
+      color: var(--home-button-primary-text);
       line-height: 36px;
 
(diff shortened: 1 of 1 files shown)


website/src/css/customTheme.scss
```diff
@@ -1707,7 +1707,7 @@ article .component-grid {
       font-weight: 600;
       margin: 0;
       padding: 0 10px;
-      background-color: var(--ifm-color-primary);
+      background-color: #357da1;
```
Simek, line 1710: Can we use one of existing variables instead raw hex? 

It might be nice to align the appearance (also text color) with primary buttons on the home page:



In general, we should stick to two or three blue shades which are derived from base brand color.

hichemfantar: what do you think?

image

image

Simek: Looks nice, thanks!

Simek (approved):
Thanks for the updates! 🚀

PR #4441, opened 2025-01-14
Better table accessibility and responsiveness and fix overflow bug

hichemfantar:
Improve table readability by refining styles and enhance the colors header and border for better contrast and visibility.
It also fixes table overflow scroll on smaller devices

research:
https://www.a11y-101.com/design/capitalisation

https://accessibility.huit.harvard.edu/design-readability

before
image

after
image

before
image

after
image

before (mobile overflow)
image

after (mobile overflow scrollable)

https://github.com/user-attachments/assets/dc061824-0c03-4f45-b3ff-1a265465c4d3


hichemfantar:
@Abbondanzo @Simek @cortinico what do you think of this change?

hichemfantar:
cc @Simek

website/src/css/customTheme.scss
```diff
@@ -305,26 +304,14 @@ hr {
...
-
-    thead tr {
-      border: 0;
-    }
-
     tr th {
-      text-transform: uppercase;
-      padding: 6px 10px;
-      font-size: 0.7rem;
+      font-size: 0.9rem;
```
Simek, line 308: Can retain the original font size and apply dimmed color (satisfying WCAG AA check) in comparison to the regular text?

Screenshot 2025-03-15 at 17 49 10



Removing the uppercase decoration, altering border color and fixing overflow are great changes overall! 👍

hichemfantar: thanks!
i increased the font size to compensate for the removal of uppercase

look at how small it looks with the old size
image

I don't know about the dimmed color, i find the default docusaurus header colors contrasty enough as headers
image

wdyt

Simek: We can increase the font size a bit, uppercase variants were more readable in combination with smaller sizes. Let's also make sure that headers are not using the same text color as default text, as show on screenshot above it is possible to achieve the compliant contrast without using default text color.

Simek: I most cases on the website table headers include less important information than cells, and on APIs and Components pages the table headers are very repetitive, and do not add much to the context, hence the approach of deprioritizing them visually.

hichemfantar: changes applied

hichemfantar:
> Looks like tables are not stretched correctly after the changes.

i tried setting `width: 100%; display: table;` but that breaks responsiveness

i usually wrap the table with a div then set `width: 100%;` on the table and that usually gives me responsive full width tables. html tables are a pain

how do you wanna do this?

hichemfantar:
my usual fix to set `display:table;` (which is usually the default but docusaurus overrides it) on the table then wrap the table in a div with overflow: auto;

i believe for this to work, we need to add a wrapper on the docusaurus table renderer

Simek:
Hey @hichemfantar, I have tried to add review comment to you changeset, but I found out that it would be too confusing due to change to the approach we had to make, so instead I have created a PR from the ground up. I have tried to address most of the points and feedback you have bring up, but if I miss sone, feel free to report an issue, or create a follow up PR. And in general, thanks for bringing those things up! 🙏

PR #4444, opened 2025-01-15
Add filter documentation

joevilches:
tsia! added documentation for filter which was release on 0.76.


docs/dropshadowvalue.md
```diff
@@ -0,0 +1,55 @@
...
+  standardDeviation: '15px',
+  color: 'blue',
+}
+```
+
+## Keys and values
+
+### `offsetX`
+
+The offset on the x-axis. This can be positive or negative.
```
cipolleschi, line 23: Positive means that the dropshadow is skewed toward right, while negative means that it is skewed toward left?
Perhaps is a good idea to specify it, wdyt?

cipolleschi: How does it works with RTL languages? Is it inverted or not?

docs/dropshadowvalue.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Type             | Optional |
+| ---------------- | -------- |
+| number \| string | No       |
+
+### `offsetY`
+
+The offset on the y-axis. This can be positive or negative.
```
cipolleschi, line 31: Positive is up and negative is down?

docs/view-style-props.md
```diff
@@ -309,6 +309,35 @@ Sets the elevation of a view, using Android's underlying [elevation API](https:/
 
 ---
 
+### `filter`
+
+Adds a graphical filter to the View. This filter is comprised of any number of _filter functions_, which each represent some atomic change to the graphical composition of the View. The complete list of valid filter functions is defined below. `filter` will apply to descendants of the View as well as the View itself. `filter` implies `overflow: hidden`, so descendants will be clipped to fit the bounds of the View.
```
cipolleschi, line 314:
```suggestion
Adds a graphical filter to the View. This filter is comprised of any number of _filter functions_, which each represent some atomic change to the graphical composition of the `View`. The complete list of valid filter functions is defined below. `filter` will apply to descendants of the `View` as well as the `View` itself. `filter` implies `overflow: hidden`, so descendants will be clipped to fit the bounds of the `View`.
```

docs/view-style-props.md
```diff
@@ -309,6 +309,35 @@ Sets the elevation of a view, using Android's underlying [elevation API](https:/
...
+
+The following filter functions work across all platforms:
+
+- `brightness`: Changes the brightness of the View. Takes a non-negative number or percentage.
+- `opacity`: Changes the opacity, or alpha, of the View. Takes a non-negative number or percentage.
```
cipolleschi, line 319: Do we have plans to implement the other filters for iOS as well? 
If yes, we should add a box to mention that. If not, we should have a box explaining why we don't have them in iOS and we are not planning to implement them.

docs/view-style-props.md
```diff
@@ -309,6 +309,35 @@ Sets the elevation of a view, using Android's underlying [elevation API](https:/
...
+
+<div class="label basic android">Android</div>
+
+The following filter functions work on Android only:
+
+- `blur`: Blurs the View with a [Guassian blur](https://en.wikipedia.org/wiki/Gaussian_blur), where the specified length represents the radius used in the blurring algorithm. Any non-negative DIP value is valid (no percents). The larger the value, the blurrier the result.
```
cipolleschi, line 325:
```suggestion
- `blur`: Blurs the `View` with a [Guassian blur](https://en.wikipedia.org/wiki/Gaussian_blur), where the specified length represents the radius used in the blurring algorithm. Any non-negative DIP value is valid (no percents). The larger the value, the blurrier the result.
```

cipolleschi: All mention of `View` should be backticked in the page as it is a proper RN component.

docs/dropshadowvalue.md
joevilches: ooo yeah good point should add clarity there

docs/view-style-props.md
joevilches: I would be hesitant on saying that yet. I can explain why we could not get it to work in iOS. Basically we just looked at some SwiftUI code and thought we could get it working.

joevilches:
@cipolleschi responded to comments!

PR #4480, opened 2025-02-06
Add documentation for outline on View Style Props page

jorge-cab:
Added documentation for outline props which were released on 0.77

### DIFF
--- a/docs/view-style-props.md
+++ b/docs/view-style-props.md
@@ -379,6 +379,8 @@ The following filter functions work on Android only:
 `outlineColor` is only available on the [New Architecture](/architecture/landing-page)
 :::
 
+`outline-color` sets the color of an element's outline. See https://developer.mozilla.org/en-US/docs/Web/CSS/outline-color for more details.
+
 | Type               |
 | ------------------ |
 | [color](colors.md) |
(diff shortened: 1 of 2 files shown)


joevilches:
@jorge-cab make sure you add the same documentation for previous versions. `outline` came out in 0.77 iirc? If so we need to edit this page too: https://github.com/facebook/react-native-website/blob/main/website/versioned_docs/version-0.77/view-style-props.md?plain=1

docs/view-style-props.md
```diff
@@ -371,6 +371,64 @@ The following filter functions work on Android only:
...
+
+---
+
+### `outlineWidth`
+
+:::note
+`outlineWidth` is only available on the [New Architecture](/architecture/landing-page)
+:::
+
+`outlineWidth` works like `outline-width` in CSS. The outline is a line drawn around an element, outside the border. `outlineWidth` does not cause changes in the layout. See https://developer.mozilla.org/en-US/docs/Web/CSS/outline-width for more details.
```
joevilches, line 424: * if you are linking the web documentation I think its implied it works like web, so I would omit the first sentence
* "does not affect layout" again
* If you omit the first line I would do something like "The width of the outline, which is a line drawn around..."

joevilches (changes requested):

joevilches:
@jorge-cab can you make the changes to the original docs too, not just 0.77? We need some script to sync these lol

joevilches:
yay thank you!

PR #4489, opened 2025-02-08
Update to Yarn Berry

hichemfantar:
Update Yarn configuration to use node-modules linker and refine package dependencies. Adjust scripts for lock file management and improve .gitignore for Yarn-related files.


website/package.json
```diff
@@ -69,8 +69,11 @@
     "@react-native-website/update-redirects": "0.0.0",
     "@types/google.analytics": "^0.0.46",
     "alex": "^10.0.0",
+    "case-police": "^0.5.14",
+    "eslint": "^8",
```
Simek, line 73:
```suggestion
    "eslint": "^8.57.1",
```

Revert more granular range.

scripts/lint-examples/package.json
```diff
@@ -11,12 +11,12 @@
...
     "@babel/preset-env": "^7.25.4",
     "@babel/runtime": "^7.25.6",
-    "@react-native/babel-preset": "^0.76.0-rc.2",
     "@react-native-community/slider": "^4.5.3",
+    "@react-native/babel-preset": "^0.76.0-rc.2",
     "@react-native/eslint-config": "^0.76.0-rc.2",
     "@react-native/typescript-config": "^0.76.0-rc.2",
     "@types/react": "^18.2.79",
-    "eslint": "^8.57.1",
+    "eslint": "^8",
```
Simek, line 19:
```suggestion
    "eslint": "^8.57.1",
```

Revert more granular range.

package.json
```diff
@@ -22,22 +24,15 @@
     }
   },
   "dependencies": {
-    "case-police": "^0.5.14",
-    "eslint": "^8.20.0",
+    "eslint": "^8",
```
Simek, line 27:
```suggestion
    "eslint": "^8.57.1",
```

Revert more granular range.

README.md
```diff
@@ -17,14 +17,15 @@ If you are looking for the source code of the [React Native Archive website](htt
 ### Prerequisites
 
 1. [Git](https://git-scm.com/downloads).
-1. [Node](https://nodejs.org/en/download/) _(version 12 or greater)_.
-1. [Yarn](https://yarnpkg.com/lang/en/docs/install/) _(version 1.5 or greater)_.
+1. [Node](https://nodejs.org/en/download/) _(`engines` field in `package.json`)_.
+1. [Yarn](https://yarnpkg.com/corepack/) _(`packageManager` field in `package.json`)_.
```
Simek, line 21: This change is not helpful, let's revert to stating required versions directly in README file, but make sure they reflect the changes.

package.json
```diff
@@ -1,15 +1,17 @@
...
   "workspaces": [
     "website",
     "plugins/*",
     "scripts/*",
     "sync-api-docs"
   ],
   "scripts": {
-    "postinstall": "yarn update-lock",
-    "update-lock": "yarn-deduplicate",
+    "update-lock": "yarn dedupe",
```
Simek, line 14:
```suggestion
    "postinstall": "yarn update-lock",
    "update-lock": "yarn dedupe",
```

We still want to run dedupe on each install automatically, the fact that `dedupe` cannot be a part of `posinstall` is known, but besided proposed CI solution we need one for local checkout, see:
* https://github.com/yarnpkg/berry/issues/4976

.github/workflows/pre-merge.yml
```diff
@@ -16,10 +16,12 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "20"
-          cache: "yarn"
+
+      - name: Enable Corepack
+        run: corepack enable
```
Simek, line 21: Enabling Corpack is not yet that common and adds additional requirements for potential contributors. 

It's fine to enable it now, but if we encounter some reports about problems setuping ti, we might revert back to using `yarnPath` and bin uploaded in the repo.

.github/workflows/pre-merge.yml
```diff
@@ -16,10 +16,12 @@ jobs:
...
 
       - name: Install dependencies
-        run: yarn install --frozen-lockfile --no-progress --non-interactive
+        run: yarn install --immutable
```
Simek, line 25: We need to add `yarn dedupe` lock check manually now, after removal of `postinstall` script. For that we should use [`check` flag](https://yarnpkg.com/cli/dedupe#details) which errors out when lock change is encountered:

```sh
yarn dedupe --check
```

Simek (changes requested):

package.json
hichemfantar: i removed this because i had a memory leak issue where yarn install will spawn endless node processes. will revert and see if it happens again

.github/workflows/pre-merge.yml
hichemfantar: my only issue with committing yarn binaries is the security risk and i've seen plenty of devs recommend not committing the binary to vc

README.md
hichemfantar: are you sure about this because this way we have one source of truth and it can never get out of sync.

Simek: Yes, it's the part of the maintenance job to keep those versions inline. 

We should not expect users to jump across the files to find out which exact version of needed tools to run app are required, all information should be contained in a single place to make the process as streamline, as possible, especially for OSS newcomers.

package.json
Simek: If the `posinstall` approach fails after the upgrade we can try using an another lifecycle stage which makes sense, or write very simple yarn plugin which will run the command in `afterAllInstalled` hook, like suggested in the referenced above issue.

.github/workflows/pre-merge.yml
Simek: There is nothing wrong with Corepack by itself, or with using it, I'm just saying that we might consider reverting it at some point of time. 

The problem is that often tools developer expectation does not really much match the current ecosystem state and often not put that much attention to tech-savy individuals. If people are having troubles to setup the repo with corepack we will need to switch back. 

And there might be a legitimate reasons why a lot of repositories still use the binary approach. When you look at adoption of Corepack, you can see that it's even worse rate than Classic vs Berry, which does not look great too, and pnpm seems to be the main drive behind Corepack adoption spread.

package.json
hichemfantar: it seems calling yarn dedupe directly works, but invoking it from another script causes the infinite loop.

hichemfantar: feel free to reproduce the bug on your machine

hichemfantar: nevermind it works locally but breaks in ci, sigh

hichemfantar: i created a [plugin](https://github.com/facebook/react-native-website/pull/4489/commits/ed83f09caa4d04b4c9c87ad2823683c8460c94d6) and it caused a freeze as well, i think it's because yarn dedupe runs install internally so i's stuck in an infinite loop

We should just get rid of it locally and make it run in ci only. if ci fails, then just run dedupe locally and push

package.json
```diff
@@ -22,22 +24,19 @@
...
-    "update-notifier": "^6.0.2",
-    "@sideway/formula": ">=3.0.1"
+    "pretty-quick": "^4.0.0"
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "packageManager": "yarn@4.6.0",
+  "devDependencies": {
+    "@yarnpkg/core": "^4.2.0",
+    "@yarnpkg/fslib": "^3.1.1"
+  }
```
Simek, line 41:
```suggestion
```

Testing leftovers.

README.md
```diff
@@ -17,14 +17,16 @@ If you are looking for the source code of the [React Native Archive website](htt
 ### Prerequisites
 
 1. [Git](https://git-scm.com/downloads).
-1. [Node](https://nodejs.org/en/download/) _(version 12 or greater)_.
-1. [Yarn](https://yarnpkg.com/lang/en/docs/install/) _(version 1.5 or greater)_.
+1. [Node](https://nodejs.org/en/download/) _(>=v20)_.
+1. [Yarn](https://yarnpkg.com/corepack/) _(v4)_.
```
Simek, line 21:
```suggestion
1. [Node](https://nodejs.org/en/download/) _(version 20 or greater)_..
1. [Yarn](https://yarnpkg.com/corepack/) _(version 4 or greater)_.
```

There is no need to abbreviate the message, let's retain the existing form, just replace with correct values.

README.md
```diff
@@ -17,14 +17,16 @@ If you are looking for the source code of the [React Native Archive website](htt
...
 1. A fork of the repo _(for any contributions)_.
 1. A clone of the `react-native-website` repo.
 
 ### Installation
 
 1. `cd react-native-website` to go into the project root.
+1. Run `npm install -g corepack@latest` to make sure [Corepack](https://github.com/nodejs/corepack?tab=readme-ov-file#-corepack) is installed and up to date.
+1. Run `corepack enable` to enable Corepack.
```
Simek, line 29:
```suggestion
1. Run `corepack enable` to enable Corepack.
  > If the command above fails, run `npm install -g corepack@latest` to install the latest version of [Corepack](https://yarnpkg.com/corepack#installation).
```

According to Yarn docs this step should not be necessary on in most use cases. Let's make it as an optional hint, when enabling Corepack fails.

package.json
Simek: It's fine approach short term, but it would make the development process a bit more cumbersome, since we validate the deduplication state on the CI. Would be nice to mention the required now dedupe step in contributing section.

But in long term, we should explore other options how to ensure that the lock is deduplicated automatically, when running locally.

huntie:
Any reason we didn't go for the committed `.yarn/releases/` setup here?

e.g.

- A `yarnPath` in `.yarnrc.yml`
- Committed binary under `.yarn/releases/` (this seems to be the desire based on `.gitignore` changes, but was not committed)

AFAIK this is the default setup when running `yarn set version berry`.

```
# .yarnrc.yml
yarnPath: .yarn/releases/yarn-4.6.0.cjs
```

This would remove the added `corepack enable` requirement, and transparently enable devs to keep running `yarn` after they pull this version.

hichemfantar:
there are many reasons but the biggest one is the security risk associated with committing binaries.
Here's the ai answer:

Using **Corepack** instead of committing the Yarn binary has a few advantages:  

### 1. **Smaller Repository Size**  
   - Avoids committing binary files (`.yarn/releases/yarn-*.cjs`), keeping the repo clean and reducing its size.  

### 2. **Automatic Version Management**  
   - Corepack ensures that developers always use the correct Yarn version specified in `package.json#packageManager`, preventing mismatches.  
   - No need to manually update the committed Yarn binary when upgrading versions.  

### 3. **Less Repository Maintenance**  
   - No need to track, commit, or clean up old Yarn binaries when updating.  
   - Avoids issues with developers forgetting to commit the new Yarn binary after upgrading.  

### 4. **More Flexibility for Developers**  
   - Developers can globally manage their package managers with Corepack rather than relying on a project-committed binary.  
   - Works well across different environments without requiring binary execution permissions.  

### 5. **Better Integration with Node.js**  
   - Corepack is built into recent Node.js versions (16.10+), so it’s a native and recommended approach for managing package managers like Yarn, PNPM, and npm.  

### 6. **Improved Security**  
   - Avoids committing and executing potentially outdated or compromised binaries from the repo.  
   - Corepack ensures that the correct version of Yarn is fetched from a trusted source, reducing the risk of running a tampered binary.  

### When Not to Use Corepack?  
- If you need a **fully self-contained** repo where every dependency (including the package manager) is controlled.  
- If some developers use older Node.js versions that don’t have Corepack enabled by default.  

Overall, using Corepack simplifies versioning and keeps the repo cleaner while ensuring consistency across environments.

huntie:
@hichemfantar Interesting — from that list, I'd say it's the security point that's valid/significant 👍🏻. In which case, we should update the `.gitignore` to be correct.

I guess, leave feedback against the PR I just opened 😅

hichemfantar:
the .gitignore is based on the example provided by [yarn](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored)

PR #4497, opened 2025-02-11
Enable Yarn without Corepack, commit Yarn release binary

huntie:
## Summary

Follows https://github.com/facebook/react-native-website/pull/4489.

**Motivation**: Simplify/reduce surprises by removing the `corepack enable` requirement from this repo. Setup is zero config based on the previous requirement of having a global `yarn` binary, which self-manages based on `.yarnrc.yml`.

✅ Preserves `update-lock`/`dedupe`, and other improvements from the original PR 🙂.

Changes:

- Configures and commits `.yarn/releases/yarn-4.6.0.cjs` as the active Yarn binary — bypassing Corepack requirement, and aligning with the already set `.gitignore` config.
- Adds recommended `.gitattributes` config.
- This Yarn version switch was rebuilt based on 181cd6b5 — `yarn.lock` dependencies are lowered to their previously locked versions as of this commit (defensive).

## Test plan

```
yarn
yarn start
```

image


Simek (approved):
Thanks for working on this, tested the changes locally, LGTM! 👍

hichemfantar:
My only issue with this is if someone commits a compromised binary then everyone is compromised

hichemfantar:
i'd be okay with this if there was a way to verify the integrity of the binary

huntie:
@hichemfantar just replied on the original PR about the Corepack justification. It's a developer experience / security tradeoff; let's review.

hichemfantar:
Committing binaries without a verification step makes for the perfect supply chain attack

huntie:
@hichemfantar Yeah I'm with you (can't ignore an objectively present security delta). Seems like there's theoretically the same issue with binaries in `.yarn/plugins/` too 😓

Don't have the bandwidth to code search harder for what Yarn does against this, so happy to abandon.

Ideally, we do follow up and align:
- Restore originally locked dep versions.
- Ignore only Yarn artifacts relevant to the Corepack/`nodeLinker` setup.

hichemfantar:
@huntie wdym by `Restore originally locked dep versions`

huntie:
I still reckon the developer experience argument is strong, keeping the contributor audience of this repo in mind.

- We can mitigate the binary issue with a CI step — e.g. forcibly pull the expected version and check for any `.yarn/releases/` delta.
- Also won't prioritise this today 😅

huntie:
Yeah, I'm maybe not arguing against Corepack as a pattern — more that we want(ed) to align on developer readiness for this change in this repo.

That said, with the new learnings above, I'm happy to go with the one-time `corepack enable` here. On one of our other repos, this would definitely have been higher impact — but the website is a good test enough bed 👍🏻

PR #4504, opened 2025-02-13
update Husky version

hichemfantar:
upgrade Husky to version 9 for improved functionality.


hichemfantar:
cc @cortinico

Simek (commented):
The hooks are not working for me locally on commit, when using this branch as base. Re-attempted the same test on `main`, and looks like commit code is re-formatted correctly.

@hichemfantar How did you verified the changes made during the upgrade?

hichemfantar:
It’s working for me.
Let me try again

hichemfantar:
my bad i must've accidentally removed the prepare script before pushing
everything works now

Simek:
Something is still not working fully correctly for me locally. Assume changes like below has been applied to file:
Screenshot 2025-02-19 at 19 52 47

When commiting from `main`, the diff is reduced to:
Screenshot 2025-02-19 at 19 52 01

While trying the exact same change and procedure on `update-husky` branch lead to committing the following changes:
Screenshot 2025-02-19 at 19 53 45

hichemfantar:
have you tried rerunning yarn install?

Simek:
Yup, and made sure that prepare` was called and `.husky` directory has been created. After digging a bit more it looks like Husky have some compatibility issues with JetBrains IDEs: 
* https://github.com/typicode/husky/issues/1560

And the root cause is caused by lack of execution flags set on the script generated script:
* https://intellij-support.jetbrains.com/hc/en-us/community/posts/4403339945490-Webstorm-Run-git-hooks-option-disappeared

Altering the execution flag manually, after hook shell script is generated, yields the correct results (and committed files are cleaned up correctly). So, you need to update the flags on the script file and commit the changes, so OS know it can be executed.

hichemfantar:
great catch! i updated the permission via `chmod +x .husky/pre-commit`

PR #4505, opened 2025-02-13
Update AVD Manager image size in documentation

hichemfantar:
Adjust the image size for the Android Studio AVD Manager in the documentation to improve clarity and presentation.

image


cortinico:
Ideally getting a higher resolution asset would also be needed if we increase it in side

hichemfantar:
I'm happy to do it if you could provide a high res icon because I can't find it.
in the meantime, i believe this to be a good enough fix for now since the icon is actually legible now

cortinico (changes requested):
It's one of the AVD manager listed here:
https://android.googlesource.com/assets/android-studio-ux-assets/+/refs/heads/studio-main/icons/shell/toolbar/

Could you pick it and import it here?

hichemfantar:
used the 2x one but too bad it's not svg

image

cortinico (approved):

cortinico:
Here the SVG actually :) https://github.com/JetBrains/android/blob/645918268343dcc972348fcafcf1168b68ebacdf/artwork/resources/studio/icons/shell/menu/avd-manager.svg?plain=1

Let's use this one instead so it's not pixelated

hichemfantar:
arigato for the svg

Omar200455:
> Adjust the image size for the Android Studio AVD Manager in the documentation to improve clarity and presentation.
> 
> image
>

Omar200455 (commented):
71c8beed5408a8823eab26b434be587032d0f958

PR #4520, opened 2025-02-19
Add danger text to feature flag section of the Pointer Events blog post

vincentriemer:
As we get closer to enabling Pointer Events by default we want to ensure that people understand that utilizing these feature flags will cause breakages when we phase them out.

### DIFF
--- a/website/blog/2022-12-13-pointer-events-in-react-native.md
+++ b/website/blog/2022-12-13-pointer-events-in-react-native.md
@@ -106,6 +106,12 @@ Our Pointer Events implementation is still experimental but we’re interested i
 
 ### Enable Feature Flags
 
+:::danger
+
+Enabling these feature flags may break your setup soon as we are working to phase them out in order for us to roll Pointer Events out more broadly.
+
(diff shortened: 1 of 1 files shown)


website/blog/2022-12-13-pointer-events-in-react-native.md
```diff
@@ -106,6 +106,12 @@ Our Pointer Events implementation is still experimental but we’re interested i
 
 ### Enable Feature Flags
 
+:::danger
+
+Overriding the native feature flags below (such as RCTConstants and ReactFeatureFlags) is technically reaching into React Native's internals so doing so may break your setup soon as we are working to phase them out in order for us to roll Pointer Events out more broadly.
```
Simek:
```suggestion
Overriding the native feature flags below (such as `RCTConstants` and `ReactFeatureFlags`) is technically reaching into React Native's internals so doing so may break your setup soon as we are working to phase them out in order for us to roll Pointer Events out more broadly.
```

Formatting nits.

PR #4521, opened 2025-02-20
Enable yarn cache in GitHub Actions workflow

hichemfantar:
this was disabled when we migrated to yarn berry


Simek (approved):
Thanks for the updates, LGTM! 👍

PR #4531, opened 2025-03-14
Have column reverse below column instead of on the diagonal

bkowshik:
Having `column-reverse` below `column` felt more natural to try it when compared to on the diagonal.

image


cortinico (approved):
Can you update also the non versioned docs?

cortinico (approved):

PR #4539, opened 2025-03-19
TextInput: update `numberOfLines` section

j-piasecki:
https://github.com/facebook/react-native/pull/49549 Added support for the `numberOfLines` prop on iOS on the new architecture.

This PR also changes the description to say `maximum number of lines` to more accurately reflect that the behavior on the new architecture (TextInput expanding until it reaches the set limit) is the correct one.


Simek (approved):
Thanks for the docs update! 👍

PR #4544, opened 2025-03-20
Update subtle color variables for improved contrast

hichemfantar:
Enhance the contrast of subtle color variables in the custom theme to improve visibility and accessibility.

continues #4541

before
image

after
image


Simek:
As suggested in previous PR we should aim at complying with WCAG AA (WCAG AAA is too strict, and restrains the design option a lot). Also the color difference between regular font color, and `subtle` variant seems to be too low after the updates in this changeset:

## PROD

Screenshot 2025-03-20 at 12 25 53

## Preview

Screenshot 2025-03-20 at 12 25 30

hichemfantar:
alright i dropped the colors to AA

Simek (approved):
Tested the changes on the preview, thanks for the follow up! 🚀

PR #4546, opened 2025-03-25
Update intro-react.md

Jonnius00:
extented explanation why a **state** denoted as a **const** even it is changed over the time


DhruvTheDev1 (commented):
Hi, Nice update to the intro-react.md.

cipolleschi (approved):

PR #4554, opened 2025-04-05
Improve `Platform.select` description

ycmjason:
Hello there!

I have made some changes to the doc around `Platform.select`.

I have also included `windows`, `macos` and `web` and clearly state that those are only applicable when using their corresponding packages.

I described `native` as `the fallback for all platforms except for web` and confirmed that it is the case in [`windows`](https://github.com/microsoft/react-native-windows/blob/ae24df1fcdd5534ab7593d3f9324ad74528dfd5f/vnext/src-win/Libraries/Utilities/Platform.windows.js#L74) and [`macos`](https://github.com/microsoft/react-native-macos/blob/1d0c636e3ce74aea091293dc533d578a7e897b49/packages/react-native/Libraries/Utilities/Platform.macos.js#L89). Just to be safe I have also [checked](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Platform/index.js#L13) that `web` will not use `native` as fallback.

Thank you for your considerations!


website/versioned_docs/version-0.78/platform.md
```diff
@@ -191,10 +191,13 @@ Select method returns the most fitting value for the platform you are currently
...
-- `ios` (any)
-- `native` (any)
-- `default` (any)
+- `android`
+- `ios`
+- `macos` (only applicable when using [react-native-macos](https://github.com/microsoft/react-native-macos))
+- `native` (fallback for all platforms except for `web`)
+- `web` (only applicable when using [react-native-web](https://github.com/necolas/react-native-web))
+- `windows` (only applicable when using [react-native-windows](https://github.com/microsoft/react-native-windows))
+- `default` (fallback for all platforms)
```
cipolleschi, line 200: do you mind sorting them alphabetically, please?

cipolleschi (changes requested):
Thanks @ycmjason for the change. I left a small comment.

website/versioned_docs/version-0.78/platform.md
ycmjason: (the original list also had `default` in the last)

cipolleschi (approved):

PR #4572, opened 2025-04-11
Remove platform distinction on snapToAlignment

Abbondanzo:
The `snapToAlignment` prop has been supported in Android since 0.67 (added https://github.com/facebook/react-native/commit/c6e5640e87e7cb5b514ded2c8d4cbb039bd02c5f for vertical, https://github.com/facebook/react-native/commit/deec1db9fdf2848941326ba5bebc11f3592a301e for horizontal) so this updates all of our active docs to remove the distinction that it's only on iOS

### DIFF
--- a/docs/scrollview.md
+++ b/docs/scrollview.md
@@ -640,7 +640,7 @@ When `true`, shows a vertical scroll indicator.
 
 ---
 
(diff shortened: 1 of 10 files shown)


cipolleschi (approved):
Thanks for the PR! @Abbondanzo

PR #4586, opened 2025-04-19
Cleanup showcase with outdated users

ginnyTheCat:
Inspired by [this videos](https://www.youtube.com/watch?v=E3Yjx0fFeaA) I verified its claims for all the Android apps (i do not have access to a apple device at the moment to verify this on iOS).

- Facebook iOS, uses a different framework according to [a official blogpost](https://engineering.fb.com/2023/02/06/ios/facebook-ios-app-architecture).
- Amazon Appstore, [deprecated](https://www.amazon.com/gp/mas/get/amazonapp)
- Walmart, Pinterest and Tencent QQ, verified by me on Android to not use React Native, iOS claim from video seems believable
- Words with Friends 2, verified to use Unity on Android, unlikely to have a seperate iOS codebase
- Call of Duty Companion App, doesn't exist anymore

I marked some other popular apps as pinned to keep the 5x5 grid on the Home page.

### DIFF
--- a/website/showcase.json
+++ b/website/showcase.json
@@ -3,7 +3,6 @@
     {
       "name": "Facebook",
       "icon": "facebook.webp",
-      "linkAppStore": "https://apps.apple.com/app/facebook/id284882215",
       "linkPlayStore": "https://play.google.com/store/apps/details?id=com.facebook.katana",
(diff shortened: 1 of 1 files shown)


ginnyTheCat:
Closing in favor of #4605

PR #4602, opened 2025-05-08
Update turbo-native-modules.md

alexkev:
Resolving [Issue: #47929](https://github.com/facebook/react-native/issues/47929)


cortinico (approved):

Simek:
We have a note now below bullet points stating the facto that "Native" prefix is needed, so I think that we don't need to update bullet content anymore.

Screenshot 2026-06-11 135226

Issue #4588, opened 2025-04-21
React Native Showcase page showing non-React Native apps


dbarros:
## Description

The [React Native Showcase page](https://reactnative.dev/showcase) has numerous app listed that are have not even been written using React Native.

## What is the problem?

The Showcase page is misleading by including apps which are not written using React Native. This includes big name apps, even some from Meta themselves.

[This video](https://www.youtube.com/watch?v=E3Yjx0fFeaA) shows various examples of this.

## How can we address it?

Remove all apps which no longer use React Native.

## Why is it important?

It's misleading and deceitful.

PR #4605, opened 2025-05-12
Remove outdated entries in React Native showcase 

robinheinze:
Fixes https://github.com/facebook/react-native-website/issues/4588 

This PR removes several apps that no longer use React Native or are no longer available including:
- Skype ([Shutting Down May 2025](https://support.microsoft.com/en-us/skype/skype-is-retiring-in-may-2025-what-you-need-to-know-2a7d2501-427f-485e-8be0-2068a9f90472#:~:text=As%20of%20May%205th%2C%202025,same%20core%20features%20and%20more.))
- Amazon App Store ([deprecated](https://www.amazon.com/gp/mas/appstore/android/faq))
- Pinterest - No longer uses React Native
- Walmart Shopping & Grocery - Walmart uses React Native in other areas of the business, but not in the main grocery app. 
- Words with Friends 2 - uses Unity
- Call of Duty Companion App - No longer available
- Tencent QQ
- Gyroscope (Android)

Thank you to @ginnyTheCat for the [Original PR](https://github.com/facebook/react-native-website/pull/4586) which I based this on
(Note: I verified with Meta that all of the information on the existing Meta apps in the showcase is correct, so I'm leaving in the Facebook iOS app. )

# Visuals

CleanShot 2025-05-12 at 13 44 03@2x

CleanShot 2025-05-12 at 13 44 09@2x

CleanShot 2025-05-12 at 13 45 32@2x

### DIFF
--- a/website/showcase.json
+++ b/website/showcase.json
@@ -3,6 +3,7 @@
     {
       "name": "Facebook",
       "icon": "facebook.webp",
+      "linkAppStore": "https://apps.apple.com/app/facebook/id284882215",
       "linkPlayStore": "https://play.google.com/store/apps/details?id=com.facebook.katana",
       "linkMetaQuest": "https://www.meta.com/experiences/facebook/7495711360547796/",
       "pinned": true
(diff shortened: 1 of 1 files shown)


Abbondanzo:
>I think pinning one more app would make the grid on the front page to be 5x5 again and it seems like that is expected here.

+1 to this. There are no Amazon apps in the pinned app showcase, can't hurt to add some there. Mind pinning the Amazon Shopping app?

robinheinze:
Great catch! Just pushed an update. 

Amazon Shopping is a great choice there 🎉 
CleanShot 2025-05-12 at 17 56 30@2x

Abbondanzo (approved):

jamonholmgren:
Fantastic work! Glad this is updated!

PR #4607, opened 2025-05-14
Docusaurus 3.8

slorber:
This PR upgrades the React Native website to Docusaurus 3.8: 
https://docusaurus.io/blog/releases/3.8

---

## Visual regression tests

I'm running visual regression tests from this PR: https://github.com/slorber/rnw-visual-tests/pull/10

All the reported differences have been handled with little custom CSS changes.

## Benchmarks

This release continues to improve build times significantly.

### Bundling

Default:
- Cold build: 37s
- Warm rebuild: 17s

With the expensive optimization disabled:
- Cold build: 19s
- Warm rebuild: 3s

### SSG

Before: 13s
After: 6s

### Docusaurus Faster Impact

Here's the global updated impact of [Docusaurus Faster](https://github.com/facebook/docusaurus/issues/10556) on total build times, with all the flags on/off for the React Native website (including ~2k pages)

| Setup              | Cold Build             | Warm Rebuild             |
|-------------------|------------------------|------------------------|
| ❌ Docusaurus Slower (webpack) | 120s (baseline)   | 33s (3.6 × faster)     |
| ✅ Docusaurus Faster (Rspack) | 31s (3.8 × faster)     | 17s (7 × faster)     |


website/src/css/customTheme.scss
```diff
@@ -43,7 +43,6 @@
   --ifm-table-border-color: var(--ifm-toc-border-color);
   --ifm-table-cell-padding: 10px;
   --ifm-table-stripe-background: rgba(0, 0, 0, 0.02);
-  --docusaurus-blog-social-icon-size: 16px;
```
slorber, line 46: This was not used, and after introducing cascade layers, it became used and produced a visual diff that my tool caught.

The effective icon size on the RN website in prod is 17px, not 16px.

website/src/css/customTheme.scss
```diff
@@ -223,8 +222,14 @@ html[data-theme="dark"] {
...
-  padding-top: 2rem !important;
+/*
+Layer needed to override utility rules using !important in the Infima layer
+See also https://css-tricks.com/css-cascade-layers/#aa-important-layers
+*/
+@layer docusaurus.infima.importantOverrides {
+  main[class^="docMainContainer"] > .container {
+    padding-top: 2rem !important;
+  }
 }
```
slorber, line 233: Our Infima utility types such as `margin--xl` use `!important`

Once layered, these rules can't be overridden by unlayered CSS, and you have to add your override directly to the Infima layer with `!important` to be able to override it.

All this is explained here: https://css-tricks.com/css-cascade-layers/#aa-important-layers.

This may seem annoying, but I think it makes sense. Utility classes serve to drive the styling from the markup. If you write `margin--xl` in the HTML, it's a bit weird if the final margin is not "xl" but another value due to an unexpected override. You can see this as a way to be explicit and say "hey, I know what I'm doing, and really want to override even if this is dangerous and can lead to unexpected behavior.

The proper way would be to swizzle the component and remove `margin--xl` for example, but I understand how it might feel more practical to just use CSS 🤪 

PS: the name `"importantOverrides` doesn't matter much, what matters is that it's a "sub-layer" of Infima.

slorber (commented):
Ready to review, see my additional inline comments.

I'll release Docusaurus v3.8 tomorrow and then update the PR.

I fixed all the visual diffs that my tooling reported here: https://app.argos-ci.com/slorber/rnw-visual-tests

Normally this should look exactly as before

Simek:
Thanks for putting all of the work into migrating the website, and explaining all of the changes needed! 🙏 

Tested the website locally, and on the PR preview, it looks and performs great!

PR #4612, opened 2025-05-15
Update React's web address

dsvwilson:
Updated the React web address to https://react.dev/ from the old and non-functional reactjs.org.


Simek:
Hey @dsvwilson, thanks for pointing this issue out! 👍 

We have updated most of the legacy website link for all versions available in the docs, so I will close that PR. Feel free to open a new one, when you spot that we have missed anything!

PR #4628, opened 2025-06-03
[docs] Add user guide for the Strict TypeScript API (0.80)

huntie:
As titled, will be launched with the 0.80 docs cut. Intending to link to the breaking changes listed here from our release comms.

Thanks @j-piasecki for helping compile this! 🙏🏻

# [View the rendered preview](https://deploy-preview-4628--react-native.netlify.app/docs/next/strict-typescript-api)

### DIFF
--- a/docs/strict-typescript-api.md
+++ b/docs/strict-typescript-api.md
@@ -1,0 +1,191 @@
+---
+id: strict-typescript-api
+title: Strict TypeScript API (opt in)
+---
+
+The Strict TypeScript API is a preview of our future, stable JavaScript API for React Native.
+
+Specifically, this is a new set of TypeScript types for the `react-native` npm package, available from 0.80 onwards. These provide stronger and more futureproof type accuracy, and will allow us to confidently evolve React Native's API into a stable shape. Opting in to the Strict TypeScript API brings some structural type differences, and is therefore a one-time breaking change.
+
+The new types are:
+
+1. **Generated directly from our source code** — improving coverage and correctness, so you can expect stronger compatibility guarantees.
+2. **Restricted to `react-native`'s index file** — more tightly defining our public API, and meaning we won't break the API when making internal file changes.
+
+When the community is ready, the Strict TypeScript API will become our default API in future — synchronized with deep imports removal.
+
+## Opting in
+
+We're shipping these new types alongside our existing types, meaning you can choose to migrate when ready. We encourage early adopters and newly created apps to opt in via your `tsconfig.json` file.
+
+Opting in is a **breaking change**, since some of our new types have updated names and shapes, although many apps won't be affected. You can learn about each breaking change in the next section.
+
+```json title="tsconfig.json"
+{
+  "extends": "@react-native/typescript-config",
+  "compilerOptions": {
+    ...
+    "customConditions": ["react-native-strict-api"]
+  }
+}
+```
+
+:::note Under the hood
+
+This will instruct TypeScript to resolve `react-native` types from our new [`types_generated/`](https://www.npmjs.com/package/react-native?activeTab=code) dir, instead of the previous [`types/`](https://www.npmjs.com/package/react-native?activeTab=code) dir (manually maintained). No restart of TypeScript or your text editor is required.
+
+:::
+
+The Strict TypeScript API follows our [RFC](https://github.com/react-native-community/discussions-and-proposals/pull/894) to remove deep imports from React Native. Therefore, some APIs are no longer exported at root. This is intentional, in order to reduce the overall surface area of React Native's API.
+
+:::tip API feedback
+
+**Sending feedback**: We will be working with the community to finalize which APIs we export over (at least) the next two React Native releases. Please share your feedback in our [feedback thread](https://github.com/react-native-community/discussions-and-proposals/discussions/893).
+
+See also our announcement blog post (incoming) for more info on our motivation and timelines.
+
+:::
+
+## Migration guide
+
+### Codegen types should now be imported from the `react-native` package
+
+Types used for codegen, like `Int32`, `Double`, `WithDefault` etc. are now available under a single `CodegenTypes` namespace. Similarly, `codegenNativeComponent` and `codegenNativeCommands` are now available to import from the react-native package instead of using the deep import.
+
+Namespaced `CodegenTypes` as well as `codegenNativeCommands` and `codegenNativeComponent` are also available from `react-native` package when the Strict API is not enabled to make the adoption easier for third-party libraries.
+
+**Before**
+
+```ts title=""
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+import type {
+  Int32,
+  WithDefault,
+} from 'react-native/Libraries/Types/CodegenTypes';
+
+interface NativeProps extends ViewProps {
+  enabled?: WithDefault<boolean, true>;
+  size?: Int32;
+}
+
+export default codegenNativeComponent<NativeProps>(
+  'RNCustomComponent',
+);
+```
+
+**After**
+
+```ts title=""
+import {CodegenTypes, codegenNativeComponent} from 'react-native';
+
+interface NativeProps extends ViewProps {
+  enabled?: CodegenTypes.WithDefault<boolean, true>;
+  size?: CodegenTypes.Int32;
+}
+
+export default codegenNativeComponent<NativeProps>(
+  'RNCustomComponent',
+);
+```
+
+### Removal of `*Static` types
+
+**Before**
+
+```tsx title=""
+import {Linking, LinkingStatic} from 'react-native';
+
+function foo(linking: LinkingStatic) {}
+foo(Linking);
+```
+
+**After**
+
+```tsx title=""
+import {Linking} from 'react-native';
+
+function foo(linking: Linking) {}
+foo(Linking);
+```
+
+The following APIs were previously named as `*Static` plus a variable declaration of said type. In most cases there was an alias so that value and the type were exported under the same identifier, but some were missing.
+
+(For example there was an `AlertStatic` type, `Alert` variable of type `AlertStatic` and type `Alert` which was an alias for `AlertStatic`. But in the case of `PixelRatio` there was a `PixelRatioStatic` type and a `PixelRatio` variable of that type without additional type aliases.)
+
+**Affected APIs**
+
+- `AlertStatic`
+- `ActionSheetIOSStatic`
+- `ToastAndroidStatic`
+- `InteractionManagerStatic` (In this case there was no relevant `InteractionManager` type alias)
+- `UIManagerStatic`
+- `PlatformStatic`
+- `SectionListStatic`
+- `PixelRatioStatic` (In this case there was no relevant `PixelRatio` type alias)
+- `AppStateStatic`
+- `AccessibilityInfoStatic`
+- `ImageResizeModeStatic`
+- `BackHandlerStatic`
+- `DevMenuStatic` (In this case there was no relevant `DevMenu` type alias)
+- `ClipboardStatic`
+- `PermissionsAndroidStatic`
+- `ShareStatic`
+- `DeviceEventEmitterStatic`
+- `LayoutAnimationStatic`
+- `KeyboardStatic` (In this case there was no relevant `Keyboard` type alias)
+- `DevSettingsStatic` (In this case there was no relevant `DevSettings` type alias)
+- `I18nManagerStatic`
+- `EasingStatic`
+- `PanResponderStatic`
+- `NativeModulesStatic` (In this case there was no relevant `NativeModules` type alias)
+- `LogBoxStatic`
+- `PushNotificationIOSStatic`
+- `SettingsStatic`
+- `VibrationStatic`
+
+### Some core components are now function components instead of class components
+
+- `View`
+- `Image`
+- `TextInput`
+- `Modal`
+- `Text`
+- `TouchableWithoutFeedback`
+- `Switch`
+- `ActivityIndicator`
+- `ProgressBarAndroid`
+- `InputAccessoryView`
+- `Button`
+- `SafeAreaView`
+
+Due to this change, accessing ref types of these views requires using `React.ComponentRef<typeof View>` pattern which works as expected for both class and function components, e.g.:
+
+```ts title=""
+const ref = useRef<React.ComponentRef<typeof View>>(null);
+```
+
+## Other breaking changes
+
+### Changes to Animated types
+
+Animated nodes were previously generic types based on their interpolation output. Now, they are non-generic types with a generic `interpolate` method.
+
+`Animated.LegacyRef` is no longer available.
+
+### Unified types for optional props
+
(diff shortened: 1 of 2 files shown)


Simek:
Hey @huntie, just merged #4630 in which should fix the LLMS generation throwing when PR contains new page (one which do not exist on PROD). If you rebase and force-push the PR no, CI should be green. Sorry for the inconvenience!

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+:::tip New in 0.80!
+
+The **Strict TypeScript API** is a new, opt-in, mode for using React Native with TypeScript, providing stronger type guarantees aligned with our direction for a stable React Native API. See the **announcement post (incoming)**.
+
+:::
+
+The Strict TypeScript API is a preview of our future, stable JavaScript API for React Native.
+
+Specifically, this is a new set of TypeScript types for the `react-native` npm package, available from 0.80 onwards. These enable stronger and more futureproof type accuracy, and will be a one-time breaking change.
```
cortinico: IMHO this is not entirely true.

What if we remove a lean core a package next month? That will also be a breaking change.

I think this should be reworded so that it's not a "one-time" breaking change but more on the sense of "fewer" breaking changes

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+}
+```
+
+:::note Under the hood
+
+This will instruct TypeScript to resolve `react-native` types from our new [`types_generated/`](https://www.npmjs.com/package/react-native?activeTab=code) dir, instead of the previous [`types/`](https://www.npmjs.com/package/react-native?activeTab=code) dir (manually maintained). No restart of TypeScript or your text editor is required.
+
+:::
+
+:::tip API feedback
```
cortinico: nit: Two admonitions back to back might not read well

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+The Strict TypeScript API follows our [RFC](https://github.com/react-native-community/discussions-and-proposals/pull/894) to remove deep imports from React Native. Therefore, some APIs are no longer exported at root. This is intentional, in order to reduce the overall surface area of React Native's API.
+
+**Sending feedback**: We will be working with the community to finalize which APIs we export over (at least) the next two React Native releases. Please share your feedback in our [feedback thread](https://github.com/react-native-community/discussions-and-proposals/discussions/893)!
+
+:::
+
+## Breaking changes
```
cortinico: This feels weird to me. As this page is inside the `docs/` section, it should stand the passing of time. This is like the **landing page** of "Strict TypeScript API"

Instead having a "Breaking changes" section here makes it really specific to the 0.79 -> 0.80 migration only. So that in 3/6 months, that's not relevant anymore

cortinico (commented):
Amazing stuff! I've left some comments

docs/strict-typescript-api.md
huntie: Ah yeah, I'm referring specifically to the initial swap-over being a breaking change for existing apps. This is constant.

If I move this around in the sentence it should work!

docs/strict-typescript-api.md
huntie: Thinking is:
- We need a place to include detailed notes for each breaking change.
- Indeed, this docs page will only live until we switch to the new types by default — then will be collapsed.

docs/strict-typescript-api.md
huntie: Should be okay given they have contrasting styling. Each are additional notes around the opt-in.

docs/strict-typescript-api.md
huntie: Specifically, I envision us deleting this page for 0.82 and backlinking to a `/0.81` (at that point) version via an admonition in the main "Using TypeScript" guide.

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
+---
+id: strict-typescript-api
+title: Strict TypeScript API (opt in)
+---
+
+:::tip New in 0.80!
+
+The **Strict TypeScript API** is a new, opt-in, mode for using React Native with TypeScript, providing stronger type guarantees aligned with our direction for a stable React Native API. See the **announcement post (incoming)**.
+
+:::
```
cipolleschi: This is probably not needed. the docs are versioned, so the page won'r be there for versions < 80.

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+The new types are:
+
+1. **Generated directly from our source code** — improving coverage and correctness, so you can expect stronger compatibility guarantees.
+2. **Restricted to `react-native`'s index file** — more tightly defining our public API, and meaning we won't break the API when making internal file changes.
+
+When the community is ready, the Strict TypeScript API will become our default API in future — synchronized with deep imports removal.
+
+## Opting in
+
+We're shipping these alongside our existing types, meaning you can choose to migrate when ready. We encourage early adopters and newly created apps to opt in via your `tsconfig.json` file.
```
cipolleschi: nit:
```suggestion
We're shipping these types alongside our existing ones, meaning you can choose to migrate when ready. We encourage early adopters and newly created apps to opt in via your `tsconfig.json` file.
```

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+**Breaking**: Some of our new types have updated names and shapes, although many apps won't be affected. You can learn about each breaking change in the next section.
```
cipolleschi: if we have a section, perhaps it is not needed to call it out that this is a breaking change.

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+{
+  "extends": "@react-native/typescript-config",
+  "compilerOptions": {
+    ...
+    "customConditions": ["react-native-strict-api"]
+  }
+}
+```
+
+:::note Under the hood
```
cipolleschi, line 33: I would probably drop this note and leave the content of the note as actual content on the page.
It is the docs duty to explain what's going on, so it's fine to have it clearly stated.

docs/strict-typescript-api.md
cipolleschi: I agree wit @cortinico. I think we are abusing admonition a little bit. 
This is relevant part of the content, there is no need to put it into an admonition. Just leave it as plain text.

We do want to get feedback on those APIs.

docs/strict-typescript-api.md
cipolleschi: what if we call this section **How to migrate** instead?
In this way we can even keep it here for longer. Consider that sometimes there are users that bumps 5-6 versions of React Native, so people won't necessarily migrate from 79 to 80.

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+:::
+
+## Breaking changes
+
+### Codegen types should now be imported from the `react-native` package
+
+Types used for codegen, like `Int32`, `Double`, `WithDefault` etc. are now available under a single `CodegenTypes` namespace. Similarly, `codegenNativeComponent` and `codegenNativeCommands` are now available to import from the react-native package instead of using the deep import.
+
+Namespaced `CodegenTypes` as well as `codegenNativeCommands` and `codegenNativeComponent` are also available from react-native package when Strict API is not enabled to make the adoption easier for third-party libraries.
```
cipolleschi:
```suggestion
Namespaced `CodegenTypes` as well as `codegenNativeCommands` and `codegenNativeComponent` are also available from `react-native` package when Strict API is not enabled to make the adoption easier for third-party libraries.
```

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+```tsx title=""
+import {Linking} from 'react-native';
+
+function foo(linking: Linking) {}
+foo(Linking);
+```
+
+The following APIs were previously named as `*Static` plus a variable declaration of said type. In most cases there was an alias so that value and the type were exported under the same identifier, but some were missing.
+
+(For example there was an `AlertStatic` type, `Alert` variable of type `AlertStatic` and type `Alert` which was an alias for `AlertStatic`. But in the case of `PixelRatio` there was a `PixelRatioStatic` type and a `PixelRatio` variable of that type without additional type aliases.)
```
cipolleschi, line 113: for codegen, we have a description and then an example. Here we have an example and then the description. 
Wouldn't be better to follow the same structure of 
- Explanation
- Example
instead? 🤔

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Unified types for optional props
+
+In the new types, every optional prop will be typed as `type | undefined`.
+
+### Removal of some deprecated types
+
+All of the types listed in `DeprecatedPropertiesAlias.d.ts` will not be accessible using the generated types.
+
+Not exactly deprecated, but in TS there was `TextInputScrollEvent` type for TextInput's `onScroll`, while Flow uses generic `ScrollEvent`.
```
cipolleschi: This is not clear. What should the user use now? `ScrollEvent` also for TS?

cipolleschi (changes requested):
Great job. I left some comments/suggestions that can be addressed!

docs/strict-typescript-api.md
huntie: (ignore nit, this reads fine)

docs/strict-typescript-api.md
huntie: Sure — so what about a highlighted green admonition? :D I think it emphasises towards this goal.

docs/strict-typescript-api.md
huntie: I wanted at least one backlink to the announcement post so people can get all the context. Now integrated further down 👍🏻

docs/strict-typescript-api.md
huntie: I like the use of `:::note` to signify things that are optional to read, which this is 🙏🏻

docs/strict-typescript-api.md
huntie: Precedes a long bullet list — but yup will rearrange 👍🏻

huntie: Precedes a long bullet list — but yup will rearrange 👍🏻

docs/strict-typescript-api.md
huntie: I moved some content back to the body. I still think the green admonition provides useful emphasis to the feedback CTA.

image

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,191 @@
+---
+id: strict-typescript-api
+title: Strict TypeScript API (opt in)
+---
+
+The Strict TypeScript API is a preview of our future, stable JavaScript API for React Native.
+
+Specifically, this is a new set of TypeScript types for the `react-native` npm package, available from 0.80 onwards. These provide stronger and more futureproof type accuracy, and will allow us to confidently evolve React Native's API into a stable shape. Opting in to the Strict TypeScript API brings some structural type differences, and is therefore a one-time breaking change.
```
cortinico, line 8:
```suggestion
Specifically, this is a new set of TypeScript types for the `react-native` npm package, available from 0.80 onwards. These provide stronger and more futureproof type accuracy, and will allow us to confidently evolve React Native's API into a stable shape. Opting-in to the Strict TypeScript API brings some structural type differences, and is therefore a one-time breaking change.
```

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+The new types are:
+
+1. **Generated directly from our source code** — improving coverage and correctness, so you can expect stronger compatibility guarantees.
+2. **Restricted to `react-native`'s index file** — more tightly defining our public API, and meaning we won't break the API when making internal file changes.
+
+When the community is ready, the Strict TypeScript API will become our default API in future — synchronized with deep imports removal.
+
+## Opting in
```
cortinico, line 17:
```suggestion
## Opting-in
```

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+This will instruct TypeScript to resolve `react-native` types from our new [`types_generated/`](https://www.npmjs.com/package/react-native?activeTab=code) dir, instead of the previous [`types/`](https://www.npmjs.com/package/react-native?activeTab=code) dir (manually maintained). No restart of TypeScript or your text editor is required.
+
+:::
+
+The Strict TypeScript API follows our [RFC](https://github.com/react-native-community/discussions-and-proposals/pull/894) to remove deep imports from React Native. Therefore, some APIs are no longer exported at root. This is intentional, in order to reduce the overall surface area of React Native's API.
+
+:::tip API feedback
+
+**Sending feedback**: We will be working with the community to finalize which APIs we export over (at least) the next two React Native releases. Please share your feedback in our [feedback thread](https://github.com/react-native-community/discussions-and-proposals/discussions/893)!
```
cortinico:
```suggestion
**Sending feedback**: We will be working with the community to finalize which APIs we export over (at least) the next two React Native releases. Please share your feedback in our [feedback thread](https://github.com/react-native-community/discussions-and-proposals/discussions/893)
```

This won't fit with the tone IMHO

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+See also our announcement blog post (incoming) for more info on our motivation and timelines.
```
cortinico, line 45: TODO missing link here

docs/strict-typescript-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Migration guide
+
+### Codegen types should now be imported from the `react-native` package
+
+Types used for codegen, like `Int32`, `Double`, `WithDefault` etc. are now available under a single `CodegenTypes` namespace. Similarly, `codegenNativeComponent` and `codegenNativeCommands` are now available to import from the react-native package instead of using the deep import.
+
+Namespaced `CodegenTypes` as well as `codegenNativeCommands` and `codegenNativeComponent` are also available from `react-native` package when the Strict API is not enabled to make the adoption easier for third-party libraries.
+
+**Before**
```
cortinico, line 57: shoudl those be H4?

docs/strict-typescript-api.md
jvenene: Opting in is a verb here and should not be hyphenated. Hyphenation would only be correct if it were to be used as a compound modifier (e.g. "An opting-in approach to the Strict API allows developers to ...").

docs/strict-typescript-api.md
jvenene: Opting in is also a verb here and should not be hyphenated.

docs/strict-typescript-api.md
huntie: Yeah, race condition and will make a note! 😅

docs/strict-typescript-api.md
huntie: *Maybe*, but these will never be jumpable since they aren't unique headings.

PR #4626, opened 2025-05-28
Add I18nManager docs

scarlac:
Added to sidebar as well.
Related to https://github.com/facebook/react-native/pull/51661 and https://github.com/facebook/react-native/pull/51648

PR #4630, opened 2025-06-03
[scripts] LLMS: skip non-existing yet URLs instead of failing

Simek:
# Why

We spotted that LLMS generation script fails  in PR preview deployments when contributor adds a new page, which is not yet in the PROD.

Refs: 
* https://github.com/facebook/react-native-website/pull/4626#issuecomment-2918132359

# How

Instead of failing whole build, skip non-existing URLs and warn about them in logs. All path defined in sidebar should exist, so if there is sth random added the build will still fail, but in the cases when the problem is just with fresh pages missing on PROD this should help, and unblock contributions.

# Tests plan

1. Add random, non existing section, make sure that generation scripts works
2. Remove the bogus sidebar entry before merging.

## Preview

Screenshot 2025-06-03 at 17 49 13

### DIFF
--- a/scripts/generate-llms-txt.js
+++ b/scripts/generate-llms-txt.js
@@ -153,18 +153,10 @@ async function processUrls(urls) {
     }
   }
 
-  const result = {
+  return {
     totalUrls: urls.length,
     unavailableUrls: unavailableUrls,
   };
-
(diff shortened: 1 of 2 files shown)


Abbondanzo (approved):
Thanks for making this change 😄

PR #4639, opened 2025-06-10
Update intro-react-native-components.md to fix test components not being visible on iOS snack player

Jack-Jackk:
Changed the ScrollView component to include a white background for visibility on the iOS snack player, as well as 20% padding to keep component's from being blocked by the island.


Simek:
Hey @Jack-Jackk, thanks for the contribution! 👍 

We have address that on the Snack embed code side, and all examples will load by default in light mode, which should address this issue globally, so closing this one as done.

PR #4640, opened 2025-06-10
Update introduction.md to fix visibility on iOS snack player.

Jack-Jackk:
Adding set background color to the style so that the code is visible on iOS snack player.

### DIFF
--- a/website/versioned_docs/version-0.79/introduction.md
+++ b/website/versioned_docs/version-0.79/introduction.md
@@ -38,6 +38,7 @@ const YourApp = () => {
         flex: 1,
         justifyContent: 'center',
         alignItems: 'center',
+        backgroundColor: "white",
       }}>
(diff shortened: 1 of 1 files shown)


DhruvTheDev1 (commented):
I have tested the demo app on my physical device and it currently works without explicitly adding `backgroundColor: "white",`. However adding this is a good safety measure and can help avoid potential visibility issues.

Simek:
Hey @Jack-Jackk, thanks for spotting that issue! 👍 

I think that we should approach this a bit differently, by decoupling Snack component theme parameter, from Appetize device appearance preference in the Snack code, and then use the new param on some of the examples (simple ones) to set preview device theme to "light", no matter of website theme selected by the user.

Simek:
We have address that on the Snack embed code side (see the referenced PR), and all examples will load by default in light mode, which should address this issue globally, closing this one as done.

Issue #4654, opened 2025-06-17
Animation Website useAnimatedValue


ramonRain:
## Description

The Animations documentation shows code that is not correct and is not working in the selected version 

The animation example falls into an error when trying to show the example code.

The website uses a constant using the hook "useAnimatedValue", as a function, when react-native throws error because is not a function. "useAnimatedValue" doesnt work in the project I am working either (instead Animated.Value is being used, as corrected mentioned in the documentaiton text).

![Image](https://github.com/user-attachments/assets/ef73e9f0-b7b6-4a36-a880-8514d86bdab8)

## Documentation version

0.71, 0.72, 0.73, 0.74, 0.75, 0.76, 0.77, 0.78, 0.79 and next;

DhruvTheDev1:
@cortinico I have opened a PR for this issue

PR #4674, opened 2025-07-11
Fix: update FadeInView example to use Animated.Value

DhruvTheDev1:
Addressed #4654 
- The animation example for TypeScript uses `const fadeAnim = useAnimatedValue(0);` which is not a real function in the React Native API.
- `TypeError: (0 , _reactNative.useAnimatedValue) is not a function`
- Replaced `const fadeAnim = useAnimatedValue(0);` with `const fadeAnim = new Animated.Value(0);`
- Removed useAnimatedValue import
- However the original code does work when scanning the QR code for Expo - perhaps due to some internal support? However the code changes still works and fixes the error.

### DIFF
--- a/docs/animations.md
+++ b/docs/animations.md
@@ -21,11 +21,11 @@ For example, a container view that fades in when it is mounted may look like thi
 <TabItem value="javascript">
 
 ```SnackPlayer ext=js&supportedPlatforms=ios,android
-import React, {useRef, useEffect} from 'react';
-import {Animated, Text, View} from 'react-native';
+import React, {useEffect} from 'react';
+import {Animated, Text, View, useAnimatedValue} from 'react-native';
 
 const FadeInView = props => {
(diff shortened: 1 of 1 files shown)


DhruvTheDev1:
@cortinico Hi! Is the code change okay as it is `const fadeAnim = new Animated.Value(0);` or shall I change it to  `const fadeAnim = useRef(new Animated.Value(0)).current` and add the useRef import?
Based on this - I will update the versioned docs too.

DhruvTheDev1:
@cortinico Hi! I've updated the code to `const fadeAnim = useRef(new Animated.Value(0)).current` and added the useRef import. Let me know if it looks good.

Simek:
Hey @DhruvTheDev1, thanks for working on the fix, and sorry for missing this earlier!

The animation examples should have been fixed now across all pages, and in versioned docs. Please let me know if there are still any issues, or just open a new report/PR, happy to review and ship it! For now, I'm closing this as done.

PR #4684, opened 2025-07-29
Update linking.md (remove override keyword that make build fail in last RN version)

LaGregance:
Remove "override" keyword as AppDelegate inherit directly from UIApplicationDelegate that is a protocol, using override cause error "Method does not override any method from its superclass".


ignaciosantise:
hey @LaGregance 👋 do you know who should we tag to get this merged?

Simek (approved):
Hey @LaGregance, thanks for the correction, and sorry for the late merge.

Additional ref:
* https://github.com/facebook/react-native/blob/main/private/helloworld/ios/HelloWorld/AppDelegate.swift#L20-L23

PR #4690, opened 2025-08-06
Add experimental documentation for `VirtualView`

yungsters:
Adds initial documentation for the experimental `VirtualView` core component.

https://deploy-preview-4690--react-native.netlify.app/docs/next/virtualview

This is not being added to the side bar, which means it will only be accessible via search or a deep link.


docs/virtualview.md
```diff
@@ -0,0 +1,169 @@
...
+
+:::note
+In future developments, a hidden `VirtualView` may instead render its children in an [`<Activity mode="hidden">`](https://react.dev/reference/react/Activity) to preserve state for as long as possible while balancing memory overhead.
+:::
+
+### Blocking the Main Thread
+
+This is the first time in React Native’s feature set where rendering a React component can block the main thread. This is a new capability enabled by the [New Architecture](/architecture/landing-page)!
+
+Blocking the main thread can provide a better user experience by preventing flashes of blank frames that sometimes occurwhen using components like [`FlatList`](flatlist). It can also enable better performance by using main thread priority, which is also typically run on higher performance cores.
```
lunaleaps:
```suggestion
Blocking the main thread can provide a better user experience by preventing flashes of blank frames that sometimes occur when using components like [`FlatList`](flatlist). It can also enable better performance by using main thread priority, which is also typically run on higher performance cores.
```

docs/virtualview.md
```diff
@@ -0,0 +1,169 @@
...
+| mode          | [VirtualViewMode](#virtualviewmode) | New mode of the `VirtualView`.                                                                    |
+| target        | element                             | `VirtualView` emitting this event.                                                                |
+| targetRect    | [Rect](rect)                        | Layout of `target` relative to the nearest ancestor `ScrollView`.                                 |
+| thresholdRect | [Rect](rect)                        | Layout of the threshold that triggered this event, relative to the nearest ancestor `ScrollView`. |
+
+:::note
+For example, if a `VirtualView` enters the visible region of a [`ScrollView`](scrollview)...
+
+- `mode` would be [`VirtualViewMode.Visible`](#virtualviewmode)
+- `thresholdRect` would describe the current scroll position of the nearest ancestor [`ScrollView`](scrollview)
```
lunaleaps: > describe the current scroll position 

I'm wondering if we should word it as the visible viewport of the nearest ancestor ScrollView to be clearer what we mean by scroll position

lunaleaps (approved):
👏

yungsters:
I plan to refine the visual treatment for experimental components, but I will do so in a follow-up PR.

PR #4709, opened 2025-08-09
Home: tweaks and fixes

Simek:
# Why

Spotted that if user OS has the preferred theme set to dark, and website loads with "Auto" theme, the assets on home page use incorrect theme.

Zrzut ekranu 2025-08-09 133122

# How

Summary of changes:

* Fix the home page assets theme on initial load with "Auto" theme via utilizing buil-in Docusaurus ThemedImage component, see: https://docusaurus.io/docs/markdown-features/assets#themed-images

* Improve mobile styles to prevent overflow on narrow mobile devices

* Match some of the custom breakpoints to match already used values on the website

* Add Bluesky profile link in the Watch section

# Preview

Zrzut ekranu 2025-08-09 132713


cortinico (commented):
Can you fix the failures so we can check the preview?

cortinico (approved):

PR #4716, opened 2025-08-11
add Vitali to authors.yml

vzaidman:


vzaidman:
merged in https://github.com/facebook/react-native-website/pull/4701 instead

PR #4721, opened 2025-08-12
fix `class` attr warning for SnackPlayer and on multiple pages

Simek:
# Why

Address `class` attr warning spotted in the dev console caused by SnackPlayer and platform tags on multiple pages.

Screenshot 2025-08-12 at 08 21 09

# How

Replace `class` attr with `className` in SnackPlayer component and for all platform tags used in docs content files.


cortinico (approved):
That's a really long PR. I've skimmed through it and it looks good at first. I suppose this was automatically done and no other modifications are included

Simek:
I can confirm that there are no other changes than `class` => `className` replace. I have to use regex replace on all MD files in `/docs` and `/website/versioned_docs` to address that issue, due to how many instance there were to correct. And after applying the fixes I have run `format:markdown`  script.

PR #4733, opened 2025-08-15
upgrade React to the latest 19.x release

Simek:
# Why

Follow up to:
* https://github.com/facebook/react-native-website/pull/4710#discussion_r2277071096

# How

Upgrade React to the latest 19.x release, run docs app locally, make sure everything works and renders correctly.

I had to also bump `lint-examples` package dependencies to address new error which popped out while linting.


Simek:
To ease the upgrading process in the future and make sure that dependencies across workspace matches I have also added [`@manypkg/cli` package](https://github.com/Thinkmill/manypkg) which runs checks on the packages versions within the monorepo, and prints outs warnings/recommendations if any mismatch has been found.

The latest commit also contains fixes for issues found by new CLI, the resolution on `@typescript-eslint/typescript-estree` has been added to prevent waring when linting TypeScript code and examples. To address this more properly we would need to switch to ESLint v9, rewrite configs to flat version, and update all plugins, which should result in using the latest version of `@typescript-eslint/typescript-estree` automatically.

---

Q: Should we hooked up manypkg check to the CI?

PR #4737, opened 2025-08-18
migrate to ESLint flat config, convert to ESM

Simek:
# Why

Refactor and migrate to flat config ESLint v9 setup. Convert monorepo to ESM to leverage properly minimal required Node version and simplify linting.

# How

Upgrade ESLint to v9, convert setup to flat config, remove redundant stand-alone Prettier checks.

In the process, I have also:
* reorganized the repository structure, so all scripts are placed in `scripts` directory, and all monorepo packages resides in `packages` directory
* fixed `lint-examples` packages scripts, and formatted `./docs` files using it
* fixed `sync-api-docs` packages, it runs correctly, but looks like no files in core repo contain `@generate-docs` annotation anymore
* fixed and re-enabled `remark-lint-no-broken-external-links check` plugin for unversioned docs via new script which should be run locally
* fixed and hooked up `image-check` script
* fixed Snack remark plugin tests

#### Future TODOs

* fix LLMS generation (it's not working properly on PROD, see https://reactnative.dev/llms.txt)
* figure out SCSS linting (`@eslint/css` cannot parse SCSS files yet), for now we only use stand-alone Prettier to format the files

# Test plan

All lint checks and tests across monorepo are working and do not yield any errors. Website apps runs and builds as expected.


clicktodev:
can you make sure ide detection still works? because the eslint extension in vscode sometimes stops working when migrating to eslint v9

here's an [example](https://github.com/shuding/nextra/pull/4388)

Simek:
Thanks for bringing this up! 👍  Usually, in-IDEs detection problems reflects the issues in the lint config definition, or monorepo/workspace setup on users machine. 

For this PR, I have tested the following editors (also with MD/MDX content, not only for JS/CJS/MJS/TS as seen on screenshots below):
* JetBrains IDEA
Screenshot 2025-09-09 at 16 25 19


* VS Code
Screenshot 2025-09-09 at 16 26 43


* Zed
Screenshot 2025-09-09 at 16 27 51

clicktodev:
awesome, it seems as long as the eslint config is exported from the root of the project then it's correctly detected by IDEs, so in the case of nextra i needed to reexport from root it because it's defined in a package

PR #4746, opened 2025-08-21
Update tutorial in`turbo-native-modules.md` to use React Native 0.81.0

kaseken:
## Current issue

The current sample code for `NativeLocalStoragePackage.kt` cannot be built using React Native 0.76.0, due to a previous change (https://github.com/facebook/react-native-website/pull/4522), while the tutorial uses 0.76.0.

## Changes

- Updates `turbo-native-modules.md` to use React Native 0.81.0 instead of 0.76.0.
- Fixes sample code in `turbo-native-modules-android.md`.

## Tests

- I have checked locally that the new sample code works using React Native 0.81.0.

### DIFF
--- a/docs/turbo-native-modules.md
+++ b/docs/turbo-native-modules.md
@@ -20,7 +20,7 @@ The basic steps are:
 Lets work through each of these steps by building an example Turbo Native Module. The rest of this guide assume that you have created your application running the command:
 
 ```shell
-npx @react-native-community/cli@latest init TurboModuleExample --version 0.76.0
+npx @react-native-community/cli@latest init TurboModuleExample --version 0.81.0
 ```
 
 ## Native Persistent Storage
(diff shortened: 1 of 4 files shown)


docs/turbo-native-modules.md
```diff
@@ -20,7 +20,7 @@ The basic steps are:
 Lets work through each of these steps by building an example Turbo Native Module. The rest of this guide assume that you have created your application running the command:
 
 ```shell
-npx @react-native-community/cli@latest init TurboModuleExample --version 0.76.0
+npx @react-native-community/cli@latest init TurboModuleExample --version 0.81.0
```
kaseken, line 23: It might be better to specify versions for `@react-native-community/cli` instead of using `latest`, since each cli has compatible `react-native` versions.
https://github.com/react-native-community/cli?tab=readme-ov-file#compatibility

Issue #4748, opened 2025-08-22
The linking documentation instructs the user to add code that causes errors


dannyhw:
## Description

I believe the `override` here may not be necessary/useful for users trying to add deep/universal links since in the default template it will give you errors 

`method does not override any method from its superclass`

<img width="1689" height="649" alt="Image" src="https://github.com/user-attachments/assets/edbf6fdf-a31a-4a88-82bd-d254c2616fa0" />

https://reactnative.dev/docs/linking?ios-language=swift

as in 
`override func application(` should just be `func application(`

## Documentation version

0.81

Happy to make this change if I'm not mistaken about this

dannyhw:
would be addressed by https://github.com/facebook/react-native-website/pull/4684

PR #4749, opened 2025-08-22
fix: enable linking docs swift codeblock

dannyhw:
#4748

The linking documentation instructs the user to add code that errors in the community cli template

I believe the `override` to not be necessary/useful for users trying to add deep/universal links since in the default template it will give you errors

before this change
image

after this change
image

Not 100% sure but for me it only works without `override` happy to close if I'm mistaken


dannyhw:
just adding a comment to say I signed, It's just loading.

dannyhw:
I actually think this is a duplicate https://github.com/facebook/react-native-website/pull/4684

PR #4785, opened 2025-09-09
Initial docs for globals

rubennorte:
This adds placeholder docs for globals, in a new section under the `APIs` page:

Screenshot 2025-09-09 at 18 51 47


Simek:
We currently use "🚧" emoji to indicate deprecated components/APIs. 

Screenshot 2025-09-09 at 19 03 56

Recently, we started to annotate experimental pages with "⚗️" emoji, maybe this would fit better for the WIP pages?

Screenshot 2025-09-09 at 19 05 14

website/sidebars.ts
```diff
@@ -221,6 +221,44 @@ export default {
       'systrace',
       'transforms',
       'vibration',
+      {
+        type: 'category',
+        label: 'Globals',
+        collapsed: false,
```
Simek:
```suggestion
        collapsed: true,
```

Due to shear amount of the pages, we might want to collapse the section by default, to avoid large scrollable sidebar initially. We might want to collapse also further groups, to match what we do in "Guides".

Alternatively, if we want to retain visibling, we could move the "Globals" section at the end of sidebar.

rubennorte:
> We currently use "🚧" emoji to indicate deprecated components/APIs.

That's an unfortunate use of that emoji. Maybe we should reconsider that and use 🗑️ for deprecated ones, as MDN does:

Screenshot 2025-09-09 at 18 28 19

rubennorte:
@Simek we changed the existing pages for deprecated APIs to use the 🗑️ emoji and moved the 🚧 emoji to the banner instead of in the title in these cases.

Simek:
@rubennorte Yup, that's a good idea, I like it! 👍 

We also use "❌" emoji for the deleted/removed parts from the core, which still had a pages avaible on the website, and with "🗑️" emoji we can still retain this distinction.

PR #4787, opened 2025-09-09
update page status emojis in versioned docs

Simek:
# Why

Follow up to the recent unversioned docs changes:
* https://github.com/facebook/react-native-website/pull/4786
* https://github.com/facebook/react-native-website/pull/4757

# How

Replace "🚧" emojis with "🗑️" and "❌" based on the component/API status for the versioned pages.

### DIFF
--- a/website/versioned_docs/version-0.77/alertios.md
+++ b/website/versioned_docs/version-0.77/alertios.md
@@ -1,6 +1,6 @@
 ---
 id: alertios
-title: '🚧 AlertIOS'
+title: '❌ AlertIOS'
 ---
 
 > **Removed.** Use [`Alert`](alert) instead.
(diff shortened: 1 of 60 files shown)


rubennorte:
That was fast! :)

Is there any precedent for the use of ❌ to indicate a removed API? Could we just use 🗑️ instead and cross the name of the API in the title to indicate it's already been deleted instead?

Simek:
Removed pages are no longer listed in any sidebar, and only serve the purpose for legacy 3rd party content and aims to help users to find alternate solution guiding them to a given package on React Native Directory. 

The idea behind ❌ emoji is to differentiate them from deprecated parts, which still have the code in the core. We recently decided to drop the content fully from those pages, but the changes have not been backported to versioned docs, can work on this in separate PR, if you want, to improve the clarity/purpose of those pages.

An example in unversioned docs:
* https://reactnative.dev/docs/next/clipboard

Also, I'm not sure if Docusaurus supports decorations or MD formatting elements in the page titles, would need to check this out.

rubennorte (approved):

PR #4792, opened 2025-09-10
Initial docs for experimental Web Performance APIs

rubennorte:
Initial docs for the new Web Performance APIs in Canary.


docs/global-EventCounts.md
```diff
@@ -0,0 +1,10 @@
+---
+id: global-EventCounts
+title: EventCounts 🧪
```
Simek, line 3: Tested the addition on PR preview, everything looks good! 👍 

Just wonder if we should place the emoji before the sidebar/page title to match arrangement of already existing pages like "InteractionManager" or "__DEV__"?

I also looked if we can do something about not the greatest sidebar wrapping for few entries:

Screenshot 2025-10-02 201326

It's nothing blocking, but did not found any easy solution for that changing few styles in the DevTools, so it's just a note for me, for later.

docs/global-PerformanceLongTaskTiming.md
```diff
@@ -0,0 +1,14 @@
...
+title: PerformanceLongTaskTiming 🧪
+---
+
+import CanaryAPIWarning from './\_canary-channel-api-warning.mdx';
+
+<CanaryAPIWarning />
+
+The global [`PerformanceLongTaskTiming`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceLongTaskTiming) class, as defined in Web specifications.
+
+:::info Partial support
```
Simek:
```suggestion
:::warning Partial support
```

Wdyt about using other admonition style than `info`? `tip` or `note` also might work, since now we have stacked a lot of content in blue containers:

Screenshot 2025-10-02 201708

docs/global-EventCounts.md
rubennorte: > Just wonder if we should place the emoji before the sidebar/page title to match arrangement of already existing pages like "InteractionManager" or "DEV"?

I followed the same convention we have on react.dev, and it's also used by MDN. I can change the other cases to align with this instead.

docs/global-PerformanceLongTaskTiming.md
rubennorte: I'd keep this as info (again, for alignment with react.dev) and switch the Partial Support one to a warning instead.

Edit: oh, I see that note and tip are rendered different than info now, I'll use a different one (and still use warning for partial support)

docs/_canary-channel-api-warning.mdx
```diff
@@ -0,0 +1,7 @@
+:::tip Canary 🧪
+
+**This API is currently only available in React Native’s Canary and Experimental channels.**
+
```
cipolleschi, line 4: the experimental message has more details than the canary one. I'd suggest that we either provide details for both or none.

docs/global-PerformanceObserver.md
```diff
@@ -0,0 +1,62 @@
...
+    }
+  },
+);
+
+observer.observe({entryTypes: ['mark', 'measure']});
+```
+
+---
+
+# Reference
```
cipolleschi, line 38: why this is the only class with Reference and a more detailed documentation than just a link to the Web specification one?

docs/global-performance.md
```diff
@@ -1,10 +1,91 @@
...
+
+### `eventCounts`
+
+See [documentation in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance/eventCounts).
+
+### `memory`
+
+See [documentation in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory).
+
+### `rnStartupTiming` ⚠️
```
cipolleschi, line 26:
```suggestion
### `rnStartupTiming`
```

There is a warning block below, we can avoid the emoji here, imho.

docs/global-performance.md
```diff
@@ -1,10 +1,91 @@
...
+
+```ts
+get rnStartupTiming(): ReactNativeStartupTiming;
+```
+
+The `ReactNativeStartupTiming` interface provides the following fields:
+
+| Name                                     | Type           | Description                                               |
+| ---------------------------------------- | -------------- | --------------------------------------------------------- |
+| `startTime`                              | number \| void | When the application startup was started.                 |
```
cipolleschi: this might need a bit of more information. In a greenfield react native app, does it takes into account the native startup time?
What does it mean for a brownfield application, where React Native might be initialized only when the user navigate to a specific tab or screen?

docs/global-performance.md
```diff
@@ -1,10 +1,91 @@
...
+| `initializeRuntimeStart`                 | number \| void | When the React Native runtime initialization was started. |
+| `executeJavaScriptBundleEntryPointStart` | number \| void | When the execution of the application bundle was started. |
+| `endTime`                                | number \| void | When the React Native runtime was fully initialized.      |
+
+### `timeOrigin`
+
+:::warning Partial support
+Provides the number of milliseconds from the UNIX epoch until system boot, instead of the number of milliseconds from the UNIX epoch until app startup.
```
cipolleschi, line 49: "until system boot" or "since system boot"? Isn't the system boot the starting point (the 0)?

docs/global-PerformanceObserver.md
rubennorte: Because it has additional methods that are RN-specific.

docs/global-performance.md
rubennorte: "until system boot". Since system boot is just `performance.now()`.

docs/global-performance.md
rubennorte: Actually, I checked and in OSS this is always the time of RN runtime initialization. I fixed that.

docs/global-performance.md
rubennorte: This is aligned with how MDN defines the docs. For non-standard methods, the warning is always shown next to the method name.

PR #4798, opened 2025-09-12
Remove codegenConfig.ios.modulesProvider instructions in turbo-native-modules-ios.md

icedevml:
Hello,

I'm proposing to drop this part of the documentation as it seems to be NO-OP. The codegen will autodiscover the IOS implementation and generate the respective code. Maybe it was some stale snippet from older development versions.


icedevml:
Nevermind, I've confused two versions of the documentation where this changed in the meantime.

PR #4810, opened 2025-09-22
docs: mark `InteractionManager` as deprecated

Simek:
# Why

Fixes:
* #4808

# How

Add deprecation note and and indicator to the `InteractionManager` docs page, and highlight the fact again on the Timers page. The changes have been applied to the docs for 0.80, 0.81 and unversioned ones.

# Preview

Screenshot 2025-09-22 at 10 43 20

PR #4813, opened 2025-09-22
Align `InteractionManager` Deprecation Notice w/ Warning

yungsters:
Adjusts the `InteractionManager` deprecation notice that was introduced by: https://github.com/facebook/react-native-website/pull/4810

### DIFF
--- a/website/versioned_docs/version-0.80/interactionmanager.md
+++ b/website/versioned_docs/version-0.80/interactionmanager.md
@@ -6,7 +6,7 @@ title: 🗑️ InteractionManager
 import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
 
 :::warning Deprecated
-Use the [`setImmediate`](timers) instead.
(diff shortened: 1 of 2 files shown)


docs/interactionmanager.md
```diff
@@ -6,7 +6,7 @@ title: 🗑️ InteractionManager
 import Tabs from '@theme/Tabs'; import TabItem from '@theme/TabItem'; import constants from '@site/core/TabsConstants';
 
 :::warning Deprecated
-Use the [`setImmediate`](timers) instead.
+Avoid long-running work and use [`requestIdleCallback`](global-requestIdleCallback) or [`setImmediate`](timers) instead.
```
rubennorte: I wouldn't suggest `setImmediate` because in the new architecture it's implemented as `queueMicrotask` and it actually blocks the current interaction. We should just say `requestIdleCallback`

Simek:
@yangshun Thanks for the correction! 🙏 Can you also update the versioned docs for 0.80 and 0.81?

Simek (approved):

PR #4815, opened 2025-09-23
optimize assets, update README

Simek:
# How

Optimize static assets using lossless compression, this lead up to saving around ~20 MB out of starting 153 MB.

As a follow up, I would like to check, or create scripts with checks, if some of the assets are no longer used.

### Update README
* add tip about assets optimization
* add Bluesky badge
* update minimal required Node version
* use GitHub decorated callout where possible (unfortunately, they do not work when nested in lists)
* tweak wording in few places

# Preview

https://github.com/facebook/react-native-website/tree/%40simek/optimize-assets-update-readme?tab=readme-ov-file

Screenshot 2025-09-23 at 09 59 33

Screenshot 2025-09-23 at 10 06 30


Simek:
Looks like the cache was an issue which caused the first deploy attempt to fail. I have re-run the job from the Netlify panel, skipping cache, and it has worked. However, looks like the status on the CI job did not update accordingly. Will try force pushing the branch.

PR #4821, opened 2025-09-28
docs: initial docs for console.timeStamp

hoxyq:
Mostly just an outline of the API spec.

Looks a bit weird, since we don't list any other console methods here, but `console.timeStamp` is special, because it is non-standard.

### DIFF
--- a/docs/global-console.md
+++ b/docs/global-console.md
@@ -8,3 +8,48 @@ title: console
 :::
 
 The global `console` object, as defined in Web specifications.
+
+---
+
+## Methods
+
+### `timeStamp()`
+
+```tsx
+console.timeStamp(
+  label: string,
+  start?: string | number,
+  end?: string | number,
+  trackName?: string,
+  trackGroup?: string,
+  color?: DevToolsColor
+): void;
+```
+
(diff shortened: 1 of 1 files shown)


docs/global-console.md
```diff
@@ -8,3 +8,48 @@ title: console
...
+| end        | `string \| number` | No       | <ul><li>If string, the name of a previously recorded timestamp with `console.timeStamp`.</li><li>If number, the [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp). For example, from `performance.now()`.</li><li>If undefined, the current time is used.</li></ul> |
+| trackName  | `string`           | No       | The name of the custom track.                                                                                                                                                                                                                                                                                 |
+| value      | `string`           | No       | The name of the track group.                                                                                                                                                                                                                                                                                  |
```
rubennorte: ~value~ trackGroup

rubennorte:
@hoxyq we don't list any other methods because that page was just a placeholder that we need to fill with the other methods (would be awesome if you could help document them in a separate PR).

I added those recently (in #4785) because I needed a place to document the new performance APIs.

PR #4824, opened 2025-09-30
docs: add a note on Perf panel availability to console.timeStamp

hoxyq:
A short note to clarify and avoid potential confusion.

### DIFF
--- a/docs/global-console.md
+++ b/docs/global-console.md
@@ -28,6 +28,10 @@ console.timeStamp(
 
 The `console.timeStamp` API allows you to add custom timing entries in the Performance panel timeline.
 
+:::note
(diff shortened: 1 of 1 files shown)


docs/global-console.md
```diff
@@ -28,6 +28,10 @@ console.timeStamp(
 
 The `console.timeStamp` API allows you to add custom timing entries in the Performance panel timeline.
```
hoxyq, line 29: "If you are not actively profiling, this is a no-op." - worth clarifying

TBD

Simek:
Hey @hoxyq, is it still true? If so, can we merge it in, and ideally backport to some versioned docs? :)

PR #4825, opened 2025-09-30
Simplify JavaScriptCore opt-out instructions

jvaronenka:
Current instructions were no longer correct - just switching the options would make the application crash. Thus, I removed detailed instructions for switching back to JavaScriptCore from Hermes by linking to the other repository, which has correct instructions.


website/versioned_docs/version-0.81/hermes.md
```diff
@@ -92,27 +92,4 @@ This will compile JavaScript to Hermes Bytecode during build time which will imp
...
-
-```diff
-   use_react_native!(
-     :path => config[:reactNativePath],
-+    :hermes_enabled => false,
-     # An absolute path to your application root.
-     :app_path => "#{Pod::Config.instance.installation_root}/.."
-   )
-```
+React Native also supports using JavaScriptCore as the [JavaScript engine](javascript-environment). Follow the instructions to opt-out of Hermes: https://github.com/react-native-community/javascriptcore
```
Simek: Hey @jvaronenka, thanks for the updates! 👍 

Can you also updated the unversioned (Next) docs file located in the `/docs` directory?

And small suggestion to avoid raw link:
```suggestion
React Native also supports using JavaScriptCore as the [JavaScript engine](javascript-environment). Follow the instructions [from the community repository](https://github.com/react-native-community/javascriptcore) to opt-out of Hermes.
```

Simek (approved):
Thanks for the updates! 👍

PR #4826, opened 2025-09-30
docs: Clearify swapLeftAndRightInRTL functionality

HarwinBorger:
Current docs don’t clearly explain how swapLeftAndRightInRTL interacts with:
	•	marginLeft / marginRight
	•	paddingLeft / paddingRight
	•	Absolute positioning (left / right)
This leads to confusion, because developers may expect left / right to flip automatically, which it does not.

Also it's not clear what the default state of swapLeftAndRightInRTL is. This PR gives explaination on this behaviour.

### DIFF
--- a/website/versioned_docs/version-0.81/i18nmanager.md
+++ b/website/versioned_docs/version-0.81/i18nmanager.md
@@ -117,6 +117,12 @@ static doLeftAndRightSwapInRTL: boolean;
 
 A boolean value indicating whether left and right style properties should be automatically swapped when in RTL mode. When enabled, left becomes right and right becomes left in RTL layouts.
 
+**Important Notes:**
+
+- doLeftAndRightSwapInRTL is `true` by default. This can be set to `false` with `swapLeftAndRightinRTL()`.
+- Does only affect `marginLeft` / `marginRight` / `paddingLeft` / `paddingRight` / `borderLeftWidth` / `borderRightWidth`
+- Does not affect position (relative/absolute) `left` / `right`
+
 ## Methods
 
 ### `allowRTL()`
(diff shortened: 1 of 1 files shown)


HarwinBorger:
CLA is signed, github actions need a re-run

website/versioned_docs/version-0.81/i18nmanager.md
```diff
@@ -162,4 +168,4 @@ Avoid forcing RTL in production apps as it requires a full app restart to take e
 static swapLeftAndRightInRTL: (swapLeftAndRight: boolean) => void;
 ```
 
-Swap left and right style properties in RTL mode. When enabled, left becomes right and right becomes left in RTL layouts. Does not affect the value of `isRTL`.
+Swap left and right style properties in RTL mode. When enabled, left becomes right and right becomes left in RTL layouts. Does not affect the value of `isRTL`.
```
Rexogamer: this change seems to just be an unnecessary space at the end?

HarwinBorger: Thank you, you are correct, I will remove it.

website/versioned_docs/version-0.81/i18nmanager.md
```diff
@@ -117,6 +117,12 @@ static doLeftAndRightSwapInRTL: boolean;
 
 A boolean value indicating whether left and right style properties should be automatically swapped when in RTL mode. When enabled, left becomes right and right becomes left in RTL layouts.
 
+**Important Notes:**
+
+- doLeftAndRightSwapInRTL is `true` by default. This can be set to `false` with `swapLeftAndRightinRTL()`.
```
Simek, line 122: It is quite implicit from the sentence above that swap is enabled by default. I think we can skip that point, or just alter the sentence above.

website/versioned_docs/version-0.81/i18nmanager.md
```diff
@@ -117,6 +117,12 @@ static doLeftAndRightSwapInRTL: boolean;
...
+- Does only affect `marginLeft` / `marginRight` / `paddingLeft` / `paddingRight` / `borderLeftWidth` / `borderRightWidth`
```
Simek, line 123: Can you provide a code reference for this enumeration? Are you sure that's those are the only affected style props?

PR #4828, opened 2025-10-01
visual tweaks for CodeBlock component, `code` and `kbd` elements

Simek:
# Why

I have spotted that inline code blocks does not correctly line with the text in some cases, and `kbd` tags looks off when used inside annotations.

# How

Update some top level styles and vars to avoid overwrites for inline code and code/pre blocks, fix pre background bleeding in light mode for block with a title (header), apply page background color to `kbd`s.

# Preview

Screenshot 2025-10-01 at 13 59 39
Screenshot 2025-10-01 at 13 59 11
Screenshot 2025-10-01 at 14 02 59

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -29,8 +29,8 @@
   --rn-note-background: rgba(255, 229, 100, 0.25);
 
   --docsearch-key-background: none;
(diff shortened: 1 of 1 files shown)


cortinico (approved):
🙏

PR #4832, opened 2025-10-04
docs: fix guide on enabling source map generation for iOS

floydkim:
This PR fixes the guide for generating source maps during iOS builds.

Following the current guide as-is does not generate a source map file.
This diff adds `export` to the environment variable declaration so that child processes can read the `SOURCEMAP_FILE` variable.

If child processes can't read the `SOURCEMAP_FILE`, react-native-xcode.sh script will not recognize it, and `-output-source-map` option will not be added when executing hermesc.
As a result, the source map file is not generated.

https://github.com/user-attachments/assets/449d94a7-1805-4385-882e-2ec542066d34

### DIFF
--- a/docs/debugging-release-builds.md
+++ b/docs/debugging-release-builds.md
@@ -61,7 +61,7 @@ To enable source map generation:
 - Above the other exports, add a `SOURCEMAP_FILE` entry with the desired output path.
 
 ```diff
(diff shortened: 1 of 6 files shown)


Simek:
Thanks for the quick update, let's ship it! 🚀

floydkim:
@Simek Wow. I was going to let you know after the GitHub checks were finished, but you saw it already!
Thank you as always for your hard work.

PR #4833, opened 2025-10-04
cleanup no longer used assets, small reorganization

Simek:
# Why

While working on recent asset optimization, I have spotted that a lot of assets are no longer used within the docs.

# How

Look for usages, nd remove assets with no reference in the current codebase. Move few logos used only in Showces to the dedicated subdirectory.

# Test plan

Make sure that app still builds and there are no errors related to missing assets.


cipolleschi (approved):

cipolleschi:
The deploy preview failed though

Simek:
> The deploy preview failed though

It has failed due to cache issue, and got re-run from Netlify panel, but this does not update GitHub CI workflow status. Will force push the changes and it should address that.

PR #4843, opened 2025-10-09
update macos icon to tahoe

fobos531:
Super minor change, simply updates the macOS icon on the landing page to Tahoe. Used this one: https://x.com/aaronp613/status/1937243831155474872

CleanShot 2025-10-09 at 10 46 20@2x


Simek:
Hey @fobos531, nice attention to the details! 👍 

One thing before we can merge, looks like the icon now has a baked in white background which causes a visual issue in dark mode:

Screenshot 2025-10-09 113128

Looks like it is due to source image change from PNG to JPG.

fobos531:
@Simek Thanks for catching it - let me know if the new one works better.

Simek:
Thanks for the update @fobos531, it looks good now! 🙏 

There are some bit rough edges around the container, but this also can be seen on currently deployed version prior the change, so we can tackle this separately.

One thing which came to my mind: looks like the image base64 string is significantly larger than before. Did you run any loseless optimizations on the PNG asset before converting it, i.e using ImgOptim app or  TinyPng website?

fobos531:
> Thanks for the update @fobos531, it looks good now! 🙏
> 
> There are some bit rough edges around the container, but this also can be seen on currently deployed version prior the change, so we can tackle this separately.
> 
> One thing which came to my mind: looks like the image base64 string is significantly larger than before. Did you run any loseless optimizations on the PNG asset before converting it, i.e using ImgOptim app or TinyPng website?

@Simek Honestly I couldn't remember what the tools were called 😅 Ran it through TinyPng, its now down to 160 KB. I tried running it multiple times but afterwards it starts developing artifacts and the savings are negligible.

PR #4866, opened 2025-11-07
Migrate plugins to TypeScript and update configurations

clicktodev:
Migrate the codebase of plugins to TypeScript for improved type safety and maintainability. Update package configurations to reflect the changes in entry points and dependencies.

remark-snackplayer and remark-lint-no-dead-urls have some types errors related to nodes but i don't the time to solve them rn

### DIFF
--- a/packages/lint-examples/package.json
+++ b/packages/lint-examples/package.json
@@ -9,7 +9,7 @@
     "tsc-examples": "./bin/tsc-examples.ts"
   },
   "scripts": {
-    "lint": "eslint"
+    "lint": "tsc --noEmit && eslint"
   },
   "devDependencies": {
     "@babel/core": "^7.28.4",
--- a/plugins/remark-codeblock-language-as-title/package.json
+++ b/plugins/remark-codeblock-language-as-title/package.json
@@ -14,7 +14,7 @@
     "src/*"
   ],
   "scripts": {
-    "lint": "eslint ."
+    "lint": "tsc --noEmit && eslint ."
   },
   "dependencies": {
     "unist-util-visit": "^5.0.0"
--- a/plugins/remark-lint-no-dead-urls/package.json
+++ b/plugins/remark-lint-no-dead-urls/package.json
@@ -14,7 +14,7 @@
     "src/*"
   ],
   "scripts": {
-    "lint": "eslint .",
+    "lint": "tsc --noEmit && eslint .",
     "test": "yarn node --experimental-vm-modules $(yarn bin jest)"
   },
   "dependencies": {
--- a/plugins/remark-snackplayer/package.json
+++ b/plugins/remark-snackplayer/package.json
@@ -16,7 +16,7 @@
     "src"
   ],
   "scripts": {
-    "lint": "eslint .",
+    "lint": "tsc --noEmit && eslint .",
     "test": "yarn tape tests/index.ts"
   },
   "dependencies": {


docs/network.md
```diff
@@ -143,7 +143,7 @@ const App = () => {
   const getMovies = async () => {
     try {
       const response = await fetch('https://reactnative.dev/movies.json');
-      const json = await response.json();
+      const json = (await response.json()) as any;
```
Simek: We should type the repones to match `movie.json` file share we host, it's here:
* https://github.com/facebook/react-native-website/blob/main/website/static/movies.json

packages/lint-examples/src/lintExamples.ts
```diff
@@ -187,8 +206,23 @@ async function runLinter(command, args) {
...
-async function updateDocuments(mappings) {
-  const mappingsByDocument = {};
+async function updateDocuments(
+  mappings: {
+    documentPath: string;
+    examplePath: string;
+    offset: number;
+    length: number;
+  }[],
+) {
```
Simek: Let's extract this to separate type definition, since it's used more than once.

plugins/remark-lint-no-dead-urls/src/index.ts
```diff
@@ -23,12 +26,18 @@ const HTTP = {
...
+    /^(https?:\/\/)(localhost|127\.0\.0\.1)(:\d+)?/.test(url),
+  isExternal: (url: string) => /(https?:\/\/)/.test(url),
+  isPath: (url: string) => /^\/.*/.test(url),
 };
 
-async function cacheFetch(urlOrPath, method, options) {
+async function cacheFetch(
+  urlOrPath: string,
+  method: Method,
+  // @ts-expect-error TODO: find proper type
```
Simek: Ideally, we would like to avoid shipping lingering TODO comments. If you can address this before the merge that would be nice!

plugins/remark-snackplayer/src/index.ts
```diff
@@ -20,15 +21,16 @@ const parseParams = (paramString = '') => {
...
   return {
     type: 'mdxJsxAttribute',
     name,
     value,
   };
 }
 
-async function toJsxNode(node) {
+async function toJsxNode(node: Node<Data>) {
+  // @ts-expect-error: TODO: find proper type
```
Simek: Same case here, as commented above. Leaving that many ignore comments in seems diminishing for the TS refactor benefits.

scripts/src/generate-llms-txt.ts
```diff
@@ -223,8 +238,28 @@ function mapDocPath(item, prefix) {
   return `${item}.md`;
 }
 
+type UnavailableUrl = {
+  url: string;
+  status: number | string;
+  error: string | null;
+};
+
+// Helper function to extract items array from Items type
```
Simek:
```suggestion
```

This is quite simple and self-explanatory helper, let's skip the comment.

website/modules/snackPlayerInitializer.ts
```diff
@@ -15,12 +15,14 @@ export default (() => {
...
     const theme = document.querySelector('html').dataset.theme;
     document.querySelectorAll('.snack-player').forEach(snack => {
+      // @ts-expect-error: dataset exists
       snack.dataset.snackTheme = theme;
     });
   };
 
   const initSnackPlayers = () => {
     updateSnacksTheme();
+    // @ts-expect-error: ExpoSnack exists
```
Simek: You can use `declare global` and global types file to extend `window` interface, i.e.:
```
declare global {
  interface Window {
    ...
  }
}
```

See:
* https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html

yarn.lock
Simek, line 1: Looks like you have removed and re-added lock file since every possible dependency has been bumped. Let's revert that and make sure that the lock contains only changes related to package addition in this PR.

It's already a chunky PR, but if you want to do some upgrade, let's move those to a separate one. 🙂

plugins/remark-snackplayer/src/index.ts
clicktodev: I'm not sure of the correct types for the AST logic, maybe you can take over this one

yarn.lock
clicktodev: fixed

PR #4870, opened 2025-11-09
Fix Snack player initialization race condition

yash04577:
Snack players were sometimes not initializing on first load or when added dynamically (e.g. in tabs or via MDX). This was due to `ExpoSnack.initialize()` being called before the deferred `embed.js` script finished loading.

---

### Reproduction Steps (Before Fix)
1. Go to any doc with a Snack player, e.g. [FlatList](https://reactnative.dev/docs/flatlist)
2. **Hard refresh the page** (`Ctrl+Shift+R`)
3. Or navigate directly via URL: `https://reactnative.dev/docs/flatlist`
4. **Snack player iframe is blank or stuck on loading**
5. Only after **switching tabs** (TS to JS or JS to TS) does the player initialize

---

<div style="display: flex; gap: 16px; align-items: flex-start;">
  <div style="flex: 1;">

### Old Behavior (Buggy)
<video width="100%" controls>
  <source src="https://github.com/user-attachments/assets/e0adb2e4-b647-476b-a2fe-f2318ba81b21" type="video/webm">
  Your browser does not support the video tag.
</video>
*(Hard refresh → blank Snack → needs tab switch)*

  </div>
  <div style="flex: 1;">

### New Behavior (Fixed)
<video width="100%" controls>
  <source src="https://github.com/user-attachments/assets/77c284f7-e4f7-4440-8620-f02c29db3d6d" type="video/webm">
  Your browser does not support the video tag.
</video>
*(Hard refresh → Snack loads instantly)*

  </div>
</div>

---

### Root Cause
- `ExpoSnack.initialize()` was called **too early** — before the `defer`red `embed.js` script loaded

---

### Fix
- Wait for `embed.js` to load before calling `initialize()`
- Use `script.addEventListener('load', ...)` — **no polling, no timeout**
- Single `MutationObserver` on `document.body` catches **all new `.snack-player` elements**


website/modules/snackPlayerInitializer.js
```diff
@@ -65,8 +65,38 @@ export default (() => {
...
+          if (node.nodeType === 1) {
+            if (
+              node.classList?.contains('snack-player') ||
+              node.querySelector?.('.snack-player')
+            ) {
+              shouldInitialize = true;
+            }
+          }
+        });
+      });
```
Simek: Hey @yash04577, thanks for catching that issue! 👍 

A question, why we need two different class based node checks here? Also, can we use `document.querySelectorAll('.snack-player')` instead, as other observer does?

website/modules/snackPlayerInitializer.js
```diff
@@ -65,8 +65,38 @@ export default (() => {
...
+
+      if (shouldInitialize && window.ExpoSnack) {
+        updateSnacksTheme();
+        window.ExpoSnack.initialize();
+      }
+    });
+
+    observer.observe(document.body, {
```
Simek: We should be able to downscope the observer pool, and just look inside doc content container, marked with `.container` class.

yash04577:
Made the requested changes:  
- Switched to `querySelectorAll('.snack-player')` instead of separate node checks  
- Limited the `MutationObserver` scope to `.container` instead of `document.body`  

@Simek  updated as per your feedback. Thanks for the review! 🙌

Simek:
Thanks for the quick updates @yash04577! 👍 

Hmm, it looks like after the recent change the problem is back on the PR preview deployment. Earlier it was working as described in the main PR comment. Can you look at it?

Screenshot 2025-11-09 205400

yash04577:
Hi @Simek , I’ve pushed an update that fixes the issue. Please take a look when you have a moment.

pooriamo:
I just pulled the latest main but I'm getting an error from this part: 
```ts
observer.observe(document.querySelector('.container'), {
```

Apparently `document.querySelector('.container')` is null 🤔  @yash04577

Simek:
Hey @pooriamo, nice catch, thanks for pointing this out! 🙏 Will merge the fix soon.

PR #4879, opened 2025-11-12
Rename 'Native Component' to 'Fabric Native Component'

GSFZamai:


cipolleschi:
Thanks for the contribution, but we are not going to merge this.
Fabric is a codename and, as sson as we are going to drop the legacy arch, it will lose its meaning as Fabric Native components will be the only native components users can create. It is already like that from 0.82.

GSFZamai:
> Thanks for the contribution, but we are not going to merge this. Fabric is a codename and, as sson as we are going to drop the legacy arch, it will lose its meaning as Fabric Native components will be the only native components users can create. It is already like that from 0.82.

I understand!
Thank you for the feedback

PR #4890, opened 2025-11-18
improve `llms.txt`, generate MD files per page

Simek:
# Why

We would like to improve the content of current `llms.txt` file and offer more AI/MCP friendly format for the content pages.

# How

Add https://github.com/signalwire/docusaurus-plugins/tree/main/packages/docusaurus-plugin-llms-txt plugin to make the `llms.txt` generation process more automatic. Configure and adjust rules to match our expectations. I have also included blog posts previously not listed in LLMS file, but downscoped them to only last three years, to avoid outdated content.

# Preview

* https://deploy-preview-4890--react-native.netlify.app/llms.txt

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -349,7 +349,7 @@ const config: Config = {
           includeBlog: true,
           includePages: true,
           includeVersionedDocs: false,
-          enableLlmsFullTxt: false,
+          enableLlmsFullTxt: true,
           excludeRoutes: [
             '/blog/201*/**',
             '/blog/2020/**',


website/docusaurus.config.ts
```diff
@@ -314,6 +314,56 @@ const config: Config = {
...
+          '/showcase/**',
+          '/contributing/**',
+          '/versions',
+          '/blog/**',
+        ],
+        content: {
+          includeBlog: true,
+          includePages: true,
+          includeVersionedDocs: false,
+          enableLlmsFullTxt: false,
```
cortinico: What happens if you turn this on?

Simek: @cortinico Do you mean `includeVersionedDocs` or `enableLlmsFullTxt` flag? (GH is not precise, unless you use suggestion code block 😓)

* `includeVersionedDocs` - adds per version repeated sections with link to files under that version paths,
* `enableLlmsFullTxt` - generates `llms-full.txt` by concatenating all generated MD files content into one file.

cortinico: > enableLlmsFullTxt - generates llms-full.txt by concatenating all generated MD files content into one file.

Sorry I meant `enableLlmsFullTxt`. Should we just enable it? I see no harm in having it on

PR #4923, opened 2025-12-12
Update release levels example to Swift

bigcupcoffee:
I think this makes more sense as the template has been using Swift for a good while now

### DIFF
--- a/docs/releases/release-levels.md
+++ b/docs/releases/release-levels.md
@@ -3,6 +3,8 @@ id: release-levels
 title: Release Levels
 ---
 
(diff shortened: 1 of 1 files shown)


Simek (commented):
Hey @bigcupcoffee, thanks for pointing this out! 👍 

Instead of removing the Objective-C code, can you use `<Tabs>` component and list examples for both languages? 

You can see an example of that pattern here:
* https://reactnative.dev/docs/linking?syntax=ios&ios-language=objc#enabling-deep-links

PR #4924, opened 2025-12-12
make project name and codegen config consistent

brainbicycle:
The project and package name at the beginning of this [doc](https://reactnative.dev/docs/turbo-native-modules-introduction?platforms=ios&android-language=java#2-configure-codegen-to-run) is different to the one at the end.


yba8543-prog (commented):
Hey

PR #4928, opened 2025-12-16
[docs] Add feature summaries for RNDT Network and Performance panels

huntie:
Docs updates for new React Native DevTools features in 0.83.

Also update existing `.label` badge with a `.primary` variant.

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1901,6 +1901,10 @@ td .color-box {
   }
 
   /* Label variants */
+  &.primary {
+    background: var(--ifm-color-primary);
+  }
+
   &.android {
     background: #a4c936;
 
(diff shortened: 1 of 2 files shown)


website/src/css/customTheme.scss
```diff
@@ -567,6 +567,18 @@ hr {
       }
     }
   }
+
+  .label {
```
Simek: This class name collides with platform labels, and alters their appearance.

* https://deploy-preview-4928--react-native.netlify.app/docs/view-style-props#bordercurve-ios
Image

* https://reactnative.dev/docs/view-style-props#bordercurve-ios
Image

Can re-use the already defined `.label`, by adding a new color class, or rename the new class to something else?

huntie: Huh, okay! :D

It's interesting how these change the TOC styling though — probably not intended and fixable if we scope to `.markdown`

image

PR #4931, opened 2025-12-16
docs: Add matrix transform documentation and example (#1899)

### LINKED ISSUE #1899: transform matrix documentation.

AbhishekGiri04:
## Summary
Fixes #1899

This PR adds comprehensive documentation for the `matrix` transform property, which was previously undocumented despite being supported in React Native since the deprecation of `transformMatrix`.

## Changes
- ✅ Added matrix transform example in the interactive demo
- ✅ Documented the 4x4 matrix structure in column-major order
- ✅ Provided practical code examples
- ✅ Added guidance on when to use matrix vs individual transforms
- ✅ Clarified use cases (UI editors, pre-calculated transforms)

## Context
The `matrix` transform property has been supported but undocumented, causing confusion for developers who need to work with transformation matrices (especially in UI editor applications). This PR addresses that gap.

## Test Plan
- Verified documentation renders correctly
- Confirmed matrix example displays properly in the interactive demo
- Tested that the matrix syntax is accurate

### DIFF
--- a/docs/transforms.md
+++ b/docs/transforms.md
@@ -220,13 +220,25 @@ The matrix is specified in column-major order:
   transform: [
     {
       matrix: [
-        scaleX, skewY, 0, 0,
-        skewX, scaleY, 0, 0,
-        0, 0, 1, 0,
-        translateX, translateY, 0, 1
-      ]
-    }
-  ]
+        scaleX,
+        skewY,
+        0,
+        0,
+        skewX,
+        scaleY,
+        0,
+        0,
+        0,
+        0,
+        1,
+        0,
+        translateX,
+        translateY,
+        0,
+        1,
+      ],
+    },
+  ];
 }
 ```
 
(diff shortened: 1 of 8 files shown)


Simek:
Hey @AbhishekGiri04, thanks for working on that addition! 👍 

Can you backport the changes also to the versioned doc files? They are located in here:

AbhishekGiri04:
@Simek Thanks for the review! I've added the matrix transform documentation to all versioned docs (0.77-0.83) and fixed the lint issues. The changes are now ready for review.

PR #4934, opened 2025-12-20
Enhance visibility of toc active link style with underline

clicktodev:
Improve the visual style of the table of contents links by adding an underline to active links for better user experience.

before

https://github.com/user-attachments/assets/534721df-4b02-453d-8819-5f3fe24df713

after

https://github.com/user-attachments/assets/85f95f8a-bf44-455e-b3a5-476b2c04f99e

### DIFF
--- a/website/src/css/customTheme.scss
+++ b/website/src/css/customTheme.scss
@@ -1493,6 +1493,7 @@ div[class^="tableOfContents"] {
 
     &:hover {
       color: var(--home-button-primary) !important;
+      text-decoration: underline;
 
       code {
         color: var(--home-button-primary) !important;


Simek:
Hey @clicktodev, thanks for proposing that idea! 👍 

Ideally, we want to align React Native and React websites as close as possible. On react.dev we use alternate background color, but the shape overall look of the styling looks better in content container layout is stretched to the full viewport (as seen on react.dev). 

In the past, we have tried to apply this to layout to React Native website too, but were not fully happy with effect, since Docusaurus/Infinima were applying so conditional spacing on multiple nest levels which was a bit hard to fully control and overwrite.

Let's see what's the take of Docusuaurus team on the underline decoration, and if the change lands in Infinima we should consider landing it here too, at least for some time until we figure out better layouting approach.

clicktodev:
looks like infima wants to implement the design from fumadocs but there's no clear plan on when it's gonna be implemented.
i suggest we merge this and then i can work on a PR that implements the fuma design.

PR #4935, opened 2025-12-23
Add documentation for sendAccessibilityEvent

Abbondanzo:
The `sendAccessibilityEvent` API has existed [since v65](https://github.com/facebook/react-native/commit/99b7052248202cee172e0b80e7ee3dfb41316746) but has not been documented in `AccessibilityInfo` directly. The old `setAccessibilityFocus` API is also being deprecated in favor of this API. Let's make sure it's documented properly

### DIFF
--- a/website/versioned_docs/version-0.77/accessibilityinfo.md
+++ b/website/versioned_docs/version-0.77/accessibilityinfo.md
@@ -234,6 +234,10 @@ Query whether reduce motion and prefer cross-fade transitions settings are curre
 
 ### `setAccessibilityFocus()`
 
+:::warning Deprecated
+Prefer using `sendAccessibilityEvent` with eventType `focus` instead.
+:::
+
 ```tsx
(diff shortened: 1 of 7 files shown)


Simek:
Hey @Abbondanzo, this looks great, thanks for the update! 🙏 

Can we backport the additions to the existing (not archived yet) versioned docs - https://github.com/facebook/react-native-website/tree/main/website/versioned_docs? I'm just not sure how far the deprecation warning should go, but you might have better context on that. 🙂

Abbondanzo:
> Hey @Abbondanzo, this looks great, thanks for the update! 🙏
> 
> Can we backport the additions to the existing (not archived yet) versioned docs - https://github.com/facebook/react-native-website/tree/main/website/versioned_docs? I'm just not sure how far the deprecation warning should go, but you might have better context on that. 🙂

I'll add to all of em' since we only go back to 77, should be safe to make that change. Thanks for the callout!

PR #4945, opened 2026-01-03
docs(TouchableWithoutFeedback): fix onLongPress timing to 500ms

Cjgfree123:
Summary
Update TouchableWithoutFeedback docs to state onLongPress fires after 500ms (was 370ms).

Motivation
The 370ms value is incorrect and inconsistent with the current press system defaults.

Changes

docs/touchablewithoutfeedback.md: replace “370 milliseconds” with “500 milliseconds”.

Test Plan

Ran yarn lint / yarn prettier in ./website.

Verified copy for accuracy; no runtime changes.


Cjgfree123:
Signed the CLA. Thanks! 

> Hi @Cjgfree123!
> 
> Thank you for your pull request and welcome to our community.
> 
> # Action Required
> In order to merge **any pull request** (code, docs, etc.), we **require** contributors to sign our **Contributor License Agreement**, and we don't seem to have one on file for you.
> 
> # Process
> In order for us to review and merge your suggested changes, please sign at https://code.facebook.com/cla. **If you are contributing on behalf of someone else (eg your employer)**, the individual CLA may not be sufficient and your employer may need to sign the corporate CLA.
> 
> Once the CLA is signed, our tooling will perform checks and validations. Afterwards, the **pull request will be tagged** with `CLA signed`. The tagging process may take up to 1 hour after signing. Please give it that time before contacting us about it.
> 
> If you have received this in error or have any questions, please contact us at [cla@meta.com](mailto:cla@meta.com?subject=CLA%20for%20facebook%2Freact-native-website%20%234945). Thanks!

Simek:
Hey @Cjgfree123, thanks for the update! 👍 

Can you provide a reference from core repo code to confirm that 500 ms is the correct value?

Cjgfree123:
- https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/Touchable/TouchableWithoutFeedback.js
- https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Pressability/Pressability.js


```jsx
// packages/react-native/Libraries/Components/Touchable/TouchableWithoutFeedback.js

const pressabilityConfig = useMemo(
    () => ({
      ...
      delayLongPress: delayLongPress,
      delayPressIn: delayPressIn,
      onLongPress: onLongPress,
      onPress: onPress,
      onPressIn: onPressIn,
    }),
    [...]
);

const eventHandlers = new Pressability(pressabilityConfig).getEventHandlers()



// packages/react-native/Libraries/Pressability/Pressability.js

const DEFAULT_LONG_PRESS_DELAY_MS = 500;

_createEventHandlers() {
  onResponderGrant() {
    /**
     * Equivalent to: Math.max(0, this._config.delayPressIn ?? 0)
     *
     * If the caller does not pass props.delayPressIn, the computed delayPressIn
     * value here will be 0.
     */
    const delayPressIn = normalizeDelay(this._config.delayPressIn);

    // ...

    /**
     * Equivalent to: Math.max(10, this._config.delayLongPress ?? 500 - delayPressIn)
     *
     * If the caller does not pass props.delayPressIn and props.delayLongPress,
     * this effectively becomes: Math.max(10, 500) = 500ms.
     *
     * Therefore, by default, onLongPress fires after 500ms.
     */
    const delayLongPress = normalizeDelay(
      this._config.delayLongPress,
      10,
      DEFAULT_LONG_PRESS_DELAY_MS - delayPressIn,
    );

    this._longPressDelayTimeout = setTimeout(() => {
      this._handleLongPress(event);
    }, delayLongPress + delayPressIn);
  }
}

function normalizeDelay(
  delay: ?number,
  min: number = 0,
  fallback: number = 0,
): number {
  return Math.max(min, delay ?? fallback);
}
```

Simek:
Thanks for linking the reference @Cjgfree123, however, I think that 370ms delay is correct in there since it's 500ms (DEFAULT_LONG_PRESS_DELAY_MS ) - 130ms (DEFAULT_MIN_PRESS_DURATION):
* https://github.com/facebook/react-native/blob/e53955e508ab84e422d67b50ac2f52f11e3160ee/packages/react-native/Libraries/Pressability/Pressability.js#L778-L782

We probably should explain it better, where the values comes from (how it's calculated) and mention also minimal press duration, instead of changing the value.

Cjgfree123:
Thank you very much for your reply.

However, I still have a question. Both TouchableWithoutFeedback and Pressable are implemented based on the Pressability class, so why does the official documentation state 500ms for Pressable onLongPress, while TouchableWithoutFeedback states 370ms?

Official documentation: https://reactnative.dev/docs/pressable#onlongpress

Simek:
Nice catch, it looks like Pressable prop mention incorrect value. Ideally we should update and align all component pages which mention long press, and explain the calculations and from where values comes from. 

Are you still interested in working on those changes with this PR?

PR #4962, opened 2026-01-12
fix plugins TS setup and tests, update workflow

Simek:
# Why

Tried to run `lint` command inside one of plugins and spotted that the setup is broken and outputs errors.

Screenshot 2026-01-12 225427

Screenshot 2026-01-12 225418

Screenshot 2026-01-12 225402

Refs:
* #4866

# How

Replace unresolved TS config for React Native apps with manually defined one for Node + DOM. Fix incorrect paths in tests imports, migrate `remark-lint-no-dead-urls` to `ts-jest`, so it can parse correctly the plugin file. Add missing `data-snack-device-appearance` attribute in Snack plugin tape output.

I have also added the plugin checks to the CI pre-merge workflow to make sure we catch any issues with plugins earlier in the future.

# Test plan

I have run `yarn lint` and `yarn test` in all plugin files, and made sure that all commands are not erroring out, and tests are passing.

### DIFF
--- a/.github/workflows/pre-merge.yml
+++ b/.github/workflows/pre-merge.yml
@@ -6,7 +6,7 @@ on:
       - main
 
 jobs:
-  lint-workspace:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository


.github/workflows/pre-merge.yml
```diff
@@ -30,8 +30,41 @@ jobs:
...
+        working-directory: plugins/remark-codeblock-language-as-title
+        run: yarn tsc
+
+      - name: Run remark-lint-no-dead-urls plugin checks
+        working-directory: plugins/remark-lint-no-dead-urls
+        run: yarn tsc && yarn test
+
+      - name: Run remark-snackplayer plugin checks
+        working-directory: plugins/remark-snackplayer
+        run: yarn tsc && yarn test
```
cortinico, line 49: nit: ideally we should have a script in the workspace `package.json` that takes care of tihs, rather than having to explicitely call `yarn tsc` and `yarn test` from the GH workflow

Simek: Good idea, will follow up on that!

PR #4963, opened 2026-01-13
run all packages/plugins checks from the workspace

Simek:
# Why

Suggested by previous PR by Nico, ref:
* https://github.com/facebook/react-native-website/pull/4962#discussion_r2685774157

# How

An attempt to simplify checks workflow and introduce one root level script to run them.


cortinico (approved):
nice :)

PR #4969, opened 2026-01-21
turbo-native-modules: Enhance docs

rawatnaresh:
Creating a Turbo Module following the same naming convention as the demo works as expected, but using a different spec name, `codegenConfig.name`, or `codegenConfig.android.javaPackageName` can make it unclear which classes to extend or import—these notes clarify the correct mappings and file locations.


docs/turbo-native-modules-ios.md
```diff
@@ -62,9 +62,15 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 ```
```
sohammeta, line 63: I am trying to view this page in the deployed netlify app
https://deploy-preview-4969--react-native.netlify.app/docs/getting-started

Can you please point me to which section(Guides, Components, API or Architecture) can I find this page? Thanks

PR #4978, opened 2026-02-05
docs: update native module instructions

mafo5:
I recently updated my ReactNative Project to 0.83
In this process I stumbled over the warning that the `ReactPackage.createNativeModules` is replaced by the `BaseReactPackage.getModule` function.

Unfortunately I'm using custom events in my modules and therefor all my modules broke after the migration.
This let me to go step by step through the documentation of the native module sections.

Here I noticed that the code is not alined to the react native community templates and a lot of deprecated code is used. So this is my attempt to fix this documentation to use the current state of code.


website/versioned_docs/version-0.83/turbo-native-modules-android.md
mosoakinyemi, line 1: Awesome! 🔥

PR #4985, opened 2026-02-13
[docs] Remove outdated bridge and null state references from AppState docs

dr-thca:
## Summary

- Removes outdated "bridge" reference from `AppState.currentState` documentation
- Removes mention of `currentState` being `null` at launch, which is no longer the case with the new architecture

## Context

Fixes https://github.com/facebook/react-native-website/issues/4689

Since React Native 0.76, the new architecture (TurboModules) is the default. `AppState.currentState` is now populated synchronously via `getConstants()` during module construction, so it is never `null` when user code accesses it. The old documentation referenced the "bridge" and a transient `null` state that no longer applies.

The TypeScript types already declare `currentState` as non-nullable (`AppStateStatus`), so this change aligns the prose with the types and the runtime behavior.

## Changes

Two sentences updated across 8 files (`docs/appstate.md` + versioned docs for 0.77-0.83):

1. **Basic Usage section**: Removed _"However, `currentState` will be null at launch while `AppState` retrieves it over the bridge."_
2. **Example description**: Removed _", and the null state will happen only momentarily"_


dr-thca:
> Hey @dr-thca, sorry for missing this PR earlier, thanks for the updates, LGTM! 👍

No problem! Thanks for having a look

PR #5033, opened 2026-03-30
mixBlendMode is Android 10+

bigcupcoffee:
`mixBlendMode` uses [android.graphics.BlendMode](https://github.com/bigcupcoffee/react-native/blob/185a6130ddc566ef67d04f83721f41d13651dbd0/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/BlendModeHelper.kt#L10), which [was added in API LEVEL 29+ (Android 10)](https://developer.android.com/reference/android/graphics/BlendMode), react-native kinda [guards it](https://github.com/facebook/react-native/blob/185a6130ddc566ef67d04f83721f41d13651dbd0/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java#L311)

### DIFF
--- a/docs/view-style-props.md
+++ b/docs/view-style-props.md
@@ -392,7 +392,7 @@ The following filter functions work on Android only:
 ### `mixBlendMode`
 
 :::note
-`mixBlendMode` is only available on the [New Architecture](/architecture/landing-page)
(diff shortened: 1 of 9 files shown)


Simek (approved):
Hey @bigcupcoffee, thanks for docs update, and providing context for it, LGTM! 👍

PR #5042, opened 2026-04-08
Do not specify package manager

cortinico:
This breaks internal Meta setup because we run on a different yarn version


Simek (changes requested):
We would need to regenerate lock using classic, update workflows and change some commands (i.e. lock dedupe) to move off Corepack and Yarn Berry.

Simek:
We migrated repository back in the PR mentioned above, closing as done.

PR #5056, opened 2026-04-22
fix: Fix old broken Margelo link

mrousavy:
I know this is an old article, but the website margelo.io is gone since a few years and we are now on margelo.com.


Simek (approved):
Thanks for the correction! 🚀

mrousavy:
wow that was fast lol thanks @Simek!

PR #5057, opened 2026-04-22
feat: Add Vercel hosting support alongside Netlify

mattcarrollcode:
## Summary

- Add `vercel.json` with build config, CORS headers, and all 104 redirect rules converted from Netlify `_redirects` format
- Update `docusaurus.config.ts` to detect both Netlify (`NETLIFY`/`CONTEXT`) and Vercel (`VERCEL`/`VERCEL_ENV`) environments for production and preview build optimizations
- Add Vercel-specific scripts (`build:vercel`, `build:vercel:fast`, `deploy:vercel`, `deploy:vercel:prod`, `dev:vercel`) without modifying existing commands
- Add `.vercel` and `.env*.local` to `.gitignore`

All existing Netlify files and scripts remain untouched — both platforms can coexist.

## Test plan

- [x] `yarn lint` passes
- [x] `yarn ci:lint` passes (website-specific lints, Prettier, example checks)
- [x] `yarn lint:packages` passes
- [x] `yarn lint:plugins` passes
- [x] `yarn check-dependencies` passes
- [x] `yarn build:vercel:fast` builds successfully (Vercel preview mode)
- [x] `yarn build:fast` still works (existing Netlify preview mode)
- [ ] CI pre-merge checks pass
- [ ] Verify redirects work after deploying to Vercel preview

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -1,7 +1,7 @@
 {
-  "buildCommand": "cd website && yarn build:vercel",
-  "devCommand": "cd website && yarn start",
-  "outputDirectory": "website/build",
+  "buildCommand": "yarn build:vercel",
+  "devCommand": "yarn start",
+  "outputDirectory": "build",
   "installCommand": "yarn install",
   "framework": null,
   "git": {
--- a/vercel.json
+++ b/vercel.json
@@ -1,0 +1,543 @@
+{
+  "buildCommand": "cd website && yarn build:vercel",
+  "devCommand": "cd website && yarn start",
+  "outputDirectory": "website/build",
+  "installCommand": "yarn install",
+  "framework": null,
+  "git": {
+    "deepClone": true
+  },
+  "headers": [
+    {
+      "source": "/movies.json",
+      "headers": [
+        {
+          "key": "Access-Control-Allow-Origin",
+          "value": "*"
+        }
+      ]
+    }
+  ],
+  "redirects": [
+    {
+      "source": "/blog/feed",
+      "destination": "/blog/rss.xml",
+      "permanent": true
+    },
+    {
+      "source": "/blog/feed.xml",
+      "destination": "/blog/rss.xml",
+      "permanent": true
+    },
+    {
+      "source": "/docs/android-setup",
+      "destination": "/docs/getting-started",
+      "permanent": true
+    },
+    {
+      "source": "/docs/building-for-apple-tv",
+      "destination": "/docs/building-for-tv",
+      "permanent": true
+    },
+    {
+      "source": "/docs/building-from-source",
+      "destination": "/contributing/how-to-build-from-source",
+      "permanent": true
+    },
+    {
+      "source": "/docs/contributing",
+      "destination": "/contributing/overview",
+      "permanent": true
+    },
+    {
+      "source": "/docs/sourcemaps",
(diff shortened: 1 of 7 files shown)


vercel.json
```diff
@@ -0,0 +1,543 @@
...
+  "devCommand": "yarn start",
+  "outputDirectory": "build",
+  "installCommand": "yarn install",
+  "framework": null,
+  "git": {
+    "deepClone": true
+  },
+  "headers": [
+    {
+      "source": "/movies.json",
```
cortinico, line 12: what's this?

vercel.json
```diff
@@ -0,0 +1,543 @@
...
+      "permanent": true
+    },
+    {
+      "source": "/help",
+      "destination": "/community/overview",
+      "permanent": true
+    },
+    {
+      "source": "/docs/0.85/:path*",
+      "destination": "/docs/:path*",
```
cortinico, line 539: We somehow need to remember about this one when bumping docs.

vercel.json
```diff
@@ -0,0 +1,543 @@
...
+      "source": "/movies.json",
+      "headers": [
+        {
+          "key": "Access-Control-Allow-Origin",
+          "value": "*"
+        }
+      ]
+    }
+  ],
+  "redirects": [
```
cortinico, line 21: Not a huge fan having to maintain this as now we have duplicated redirects (cc @Simek). Perhaps we can have a script generating this instead?

cortinico (changes requested):
I have one concern on how we're going to maintain this @mattcarrollcode 

Other than that: LGTM

vercel.json
Simek: Yeah, having a single source of truth and a script which generates the redirects for both deployments is a good idea.

vercel.json
mattcarrollcode: This is a strange one. Its for the [networking docs page](https://reactnative.dev/docs/network#handling-the-response) that references `https://reactnative.dev/movies.json` to download [this JSON file](https://github.com/facebook/react-native-website/blob/main/website/static/movies.json) as a way to learn React Native networking. Netlify has a [similar redirect in this toml file](https://github.com/facebook/react-native-website/blob/ad88dab6f5ad303cdd98282d273e162b5242e355/netlify.toml#L14)

If its not necessary anymore we can remove it, this PR was intended to be a 1:1 replacement for Netlify so I kept it in here for now.

vercel.json
mattcarrollcode: I'll update the PR to take your suggestion and use a script to create these redirects based off the Netlify config so we don't need to update more than one place

vercel.json
mattcarrollcode: Updated to use a script to create and update the vercel.json redirects from the netlify config and created a GitHub Action to autorun it

mattcarrollcode:
> I have one concern on how we're going to maintain this @mattcarrollcode
> 
> Other than that: LGTM

Updated the PR to use a script to create the Vercel redirects in `vercel.json` from the Netlify config. It runs in a GitHub Action so any changes to Netlify's redirects should be mirrors in Vercel

vercel.json
cortinico: Oh TIL

cortinico (approved):

PR #5066, opened 2026-05-02
[docs] Remove unused PlatformTag component and redundant tooltip on platform tags

apicgg:
## Summary

Cleanup of leftover artifacts from two older PRs (#4097, #4178):

- **Remove `website/core/PlatformTag.tsx`** — the component is not imported or referenced anywhere in the app. Confirmed via repo-wide search; only its own definition and `export default` exist. Safe to delete as dead code.
- **Drop the `title` attribute on inline platform tags in `docs/platform-specific-code.md`** — the "Android" / "iOS" badge labels are self-explanatory, so the hover tooltip ("This section is related to X platform") adds no information.

## Why

These were introduced two years ago but never wired up across the docs. The component has sat unused, and the tooltip duplicates what the badge text already says. Removing both reduces surface area and avoids confusion for future contributors who might wonder which path is canonical.

## Changes

- Delete `website/core/PlatformTag.tsx`
- Update `docs/platform-specific-code.md` to remove `title="..."` from the two inline `<div className="label ...">` tags

## Test plan

- [X] Run the docs site locally and confirm the Android/iOS badges still render correctly on the Platform-Specific Code page
- [X] Confirm no build/type errors from the deleted component

c/c @Simek @cortinico


Simek (approved):
Hey @apicgg, thanks for the updates! 👍 

> Drop the title attribute on inline platform tags in docs/platform-specific-code.md — the "Android" / "iOS" badge labels are self-explanatory, so the hover tooltip ("This section is related to X platform") adds no information.

Ideally, we should have those more verbose descriptions for a11y purpose, but we can use `aria-label` instead. I guess that was the idea with the `PlatformTag.tsx‎` component, but it's fine to remove it for now, since the component does not support all of the platforms anyway.

PR #5070, opened 2026-05-06
[docs] Update Appearance.setColorScheme keys (0.87)

huntie:
Final recent docs update to the `Appearance` API, given https://github.com/facebook/react-native/pull/56696 was accepted.

> [!Note]
> Let's not land this one until the 0.86 version of the docs is cut.
>
> **Sidenote**: It would be great if we could cut a docs version in sync with the RN repo, but not publish it (`0.86-stable` was cut yesterday in facebook/react-native) 😅

### DIFF
--- a/docs/appearance.md
+++ b/docs/appearance.md
@@ -59,13 +59,13 @@ Although the color scheme is available immediately, this may change when not ove
 
 ### App-level overriding
 
-`setColorScheme()` overrides the color scheme at the application level — it does not affect the system setting or other applications. Passing `'unspecified'` removes any override, restoring the system preference.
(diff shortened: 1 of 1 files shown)


Simek:
> Sidenote: It would be great if we could cut a docs version in sync with the RN repo, but not publish it (0.86-stable was cut yesterday in facebook/react-native) 😅

@huntie I think we can, did not try it before, but cutting new files and not adding new version to https://github.com/facebook/react-native-website/blob/main/website/versions.json might just work.

Also `next` docs by design should reflect `main`, and since they are used a source of new version files the changes will be picked up on the next stable release. So in theory we could ship docs stuff for new release before the final cut.

Simek (approved):

huntie:
@Simek Yeah! We'd need a process update here: https://github.com/reactwg/react-native-releases/blob/main/docs/guide-release-candidate.md

PR #5071, opened 2026-05-07
[docs] Cut version 0.86 (unpublished)

huntie:
### Summary

```
yarn version:cut 0.86
```

Cuts docs version 0.86, with two differences:

- Omit 0.86 from `versions.json` (do not publish yet) — this PR is aligning with the branch cut in react-native (done this week), rather than at release time (typically +6w).
- https://github.com/facebook/react-native-website/commit/78e547216f375c818b9871655740a0dc7d752c14 is edited out (eagerly merged, but will only apply to 0.87+).

**👋🏻 Release Crew**: To publish this version on release day, add `"0.86"` to https://github.com/facebook/react-native-website/blob/main/website/versions.json.

### Test Plan

✅ 0.86 is versioned, but not yet enabled


Simek (approved):
Tested the changes locally, also re-added 0.86 to the array to make sure that docs works, everything LGTM! 👍 

> [!note]
> Only thing to keep in mind is that search does not work in 0.86 docs context since Algolia do not have an ability to crawl the pages yet.

slorber:
One idea could be to use an env variable in the Netlify deploy to filter out the 0.86 version automatically, while it's still available locally and in deploy previews

We have a config option that could help filter the versions dynamically: https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-docs#onlyIncludeVersions

PR #5076, opened 2026-05-11
Explain difference between verticalAlign and textAlignVertical

### LINKED ISSUE #5075: TextStyleProps verticalAlign vs textAlignVertical
## Description

I believe we need to add comments to describe difference between these two style props and what should be used `verticalAlign` or  `textAlignVertical`

## What is the problem?

`verticalAlign` or  `textAlignVertical` it unclear what should be used and can it be combined 

example some lib can use one prop but client can use diff prop
```tsx
import {MyText} from 'my-lib';

<MyText /> // apply `verticalAlign: 'middle'` by default

<MyText style={{ textAlignVertical: 'top' }} />  // <- Can I overload by using diff. style prop (will this even work ?) 

```

## How can we address it?

Add description to show the difference  

## Why is it important?

give devs better understanding 

## Who needs this?

Android devs

...

retyui:
Issue: https://github.com/facebook/react-native-website/issues/5075

### DIFF
--- a/docs/text-style-props.md
+++ b/docs/text-style-props.md
@@ -859,6 +859,8 @@ Specifies text alignment. On Android, the value 'justify' is only supported on O
 
 ### `textAlignVertical` <div className="label android">Android</div>
 
+An alias for `verticalAlign` style prop, if you use both properties, `verticalAlign` will take precedence over `textAlignVertical`
+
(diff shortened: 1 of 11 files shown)


Simek (approved):
That's a good idea, thanks for adding a clarification! 👍

PR #5085, opened 2026-05-16
add docusaurus-plugin-copy-page-button

portdeveloper:
adds docusaurus-plugin-copy-page-button to dependencies + plugins array

since the rn docs site is basically the canonical docusaurus showcase, figured this might be worth dogfooding — it puts a "Copy page" button in the sidebar that exports the current page as clean markdown, plus one-click "Open in Claude / ChatGPT / Gemini" actions that prefill the markdown.

typical rn scenario this helps with: someone's on a specific component or API ref page and wants to drop it into an LLM to ask "how do I use this with [X library]" — instead of selecting around the page chrome or digging into the repo for the right .md file.

auto-injects into the TOC sidebar, theme-aware, no extra config.

shipping on Ethereum execution-apis, Sui, Walrus, Seal, SuiNS, Monad, Flare, Kaia, Nillion, Chronicle, Cardano and Arbitrum docs. listed in the Docusaurus community plugins page.

repo: https://github.com/portdeveloper/docusaurus-plugin-copy-page-button
npm: https://www.npmjs.com/package/docusaurus-plugin-copy-page-button (~10k installs/month)
demo: https://portdeveloper.github.io/copy-page-button-showcase/

happy to revert if it's not a fit


Simek:
Hey @portdeveloper, thanks for working on this addition! 👍 

Some feedback:
* button does not show on pages without ToC or ToC hidden, i.e. https://deploy-preview-5085--react-native.netlify.app/docs/handling-text-input or https://deploy-preview-5085--react-native.netlify.app/docs/environment-setup
* button causes layout shift on page load, ideally we should preload it/make static, or set the correct height to the button container
* button and menu uses non-default docs font
* button sometimes does not load on first page render, you had to refresh page to see it
  Screenshot 2026-05-16 162139

If required, feel free to swizzle some components.

portdeveloper:
thanks for the detailed feedback @Simek — really useful. shipped 0.5.2 with fixes for all four:

1. **no-ToC pages** — `injectToFallbackLocation` used to drop the button as `position: fixed` in the top-right viewport corner, which was both ugly and brittle. now inserts inline after `.theme-doc-breadcrumbs` so pages like `/docs/environment-setup` and `/docs/handling-text-input` get the button at the top of the article column instead.

2. **layout shift** — reserved `min-height: 56px` on the sidebar-mode container so the slot is sized before React mounts. only partial — for a fully shift-free first paint we'd need an SSR-rendered placeholder via a swizzled TOC component, which is a bigger change. happy to do that next pass if you want it.

3. **font** — added `font-family: var(--ifm-font-family-base, inherit)` to the button and dropdown items. was inheriting browser default before.

4. **first-render race** — replaced the chained `setTimeout` polling with a `MutationObserver` on `document.body` that fires the moment the TOC or article mounts. periodic check kept as a safety net for `@docusaurus/faster` hydration patterns.

just bumped the PR to `^0.5.2`. the Netlify preview should pick it up on the rebuild — would appreciate another look when you have a sec to confirm the four issues are gone.

(also FYI the changelog entry credits your feedback by name + links back here — feel free to push back if you'd rather not be cited.)

Simek:
Hey @portdeveloper, thanks for the follow up updates. 👍  Unfortunately, looks like two issues are not solved yet.

* The page layout shift still happens, and it's now even more distracting on the pages without a ToC

https://github.com/user-attachments/assets/0962a5d8-7921-4b84-8d3e-07767b23a391

* The button sometimes still does not load on first page visit

https://github.com/user-attachments/assets/deebbd01-ef94-47c3-9c54-de53c19b3be5

portdeveloper:
Updated this PR to use the published public React entrypoint from `docusaurus-plugin-copy-page-button@0.6.1` (`docusaurus-plugin-copy-page-button/react`) and removed the temporary local type declaration.\n\nValidation:\n- `yarn --cwd website build` passes\n- `yarn --cwd website tsc` is currently blocked by the existing TypeScript 6 `baseUrl` deprecation in `tsconfig.json`, unrelated to this change

Simek:
Hey @portdeveloper, thanks for the updates! Verified the changes on the preview deployment and everything seems correct now, thanks! 👍 

I will follow up with small color and layout tweaks, but it looks good to me for the initial shipment.

Also, small feedback for the component - in our case I will use classname to alter the appearance, but it might be nice to expose interaction state styles in the config (i.e. `hoverStyle`, `focusStyle`), might be useful for some, ref:
* https://github.com/portdeveloper/docusaurus-plugin-copy-page-button#custom-styling

Simek:
Actually, just have found one more thing - looks like on first button interaction dropdown is positioned in the top left corner and moves under a button after a short while, can you fix that?


https://github.com/user-attachments/assets/72e3711c-ee5d-4ce3-972e-f2bd348f0669

portdeveloper:
Thanks @Simek ! This PR alone has been awesome feedback for me

Simek (approved):
Hey @portdeveloper, thanks for adding the button, and for all the updates! 👍 

It seems that all issues has been resolved, so we can ship this.

portdeveloper:
Yay best news i have heard today! 
proud of my little button going into the react-native website!

PR #5109, opened 2026-06-03
Update code block styling and syntax theme colors

huntie:
## Summary

Modernise apperance of code blocks by updating font, tweaking sizing, and introducing light/dark variants (aligned with react.dev).

**Changes**

- Remove the custom Source Code Pro web font in favor of the default Docusaurus system monospace font stack (**bonus**: no custom font download).
- Add separate light and dark Prism syntax themes aligned to the react.dev Sandpack color palette, replacing the previous single dark-only custom theme.
- Reduce code block font size and increase line height.
- Update code block title backgrounds to match their code block, with a keyline separator.

Syntax colors source: https://github.com/reactjs/react.dev/blob/6ec61348646040795fdaa9de14a9bec603260f87/src/styles/sandpack.css#L38-L78

**Notes**

The main motivation here is **horizontal space efficiency**, which I felt was poor with Source Code Pro.

Screenshot 2026-06-03 at 14 05 02

## Test plan

**Before**

| Light | Dark |
|--------|--------|
| file-70f1a6e09edb95899722086196de927e | file-bb93ada1f7c92b2f889685e60c6df4d7 | 

**After**

| Light | Dark |
|--------|--------|
| file-0e4c3a2eae7ba1465cf7a675717ebbe0 | file-37271875d08532e756bc950f7f0c0891 |


Simek:
This looks good, great changes overall! 👍 

Just one thing which we might want to address - the light mode code blocks lacks a bit of separation now from the background after the changes. To fix that we can add slight box shadow (like on react.dev) or just use border.

Screenshot 2026-06-03 155143

huntie:
@Simek I reckon this is enough contrast. Not keen on the visual clutter of shadows.

I guess we could try an outline too, matching the admonitions (previewed below).

image

The main thing that matters I'd say is the text contrast (vs contrast with surrounding paragraphs).

**Edit**: This also matches the homepage graphic more closely, actioning.

image

Simek (approved):
Agree, and border adds enough of separation. Thanks for the quick update! 🚀

PR #5113, opened 2026-06-04
docs: include RN dependencies flag for iOS source builds

### LINKED ISSUE #5065: Incomplete description on how to disable precompiled binaries iOS RN 0.84
## Description

https://reactnative.dev/blog/2026/02/11/react-native-0.84
The blog post explaining the default precompiled binaries as true in iOS from RN 0.84 onwards, misses the flag 
RCT_USE_RN_DEP=0 to truly turn the precompiled binaries off.
it only mentions RCT_USE_PREBUILT_RNCORE, which only works for ReactNativeCore, not ReactNativeDependencies.

## Documentation version

Version not valid as the mistake is in the blog post.

sorinc03:
## Summary
- Add `RCT_USE_RN_DEP=0` alongside `RCT_USE_PREBUILT_RNCORE=0` in the React Native 0.84 iOS source-build instructions
- Update the Legacy Architecture opt-in command so both precompiled iOS artifacts are disabled

Fixes #5065

## Test Plan
- `git diff --check`

### DIFF
--- a/website/blog/2026-02-11-react-native-0.84.mdx
+++ b/website/blog/2026-02-11-react-native-0.84.mdx
@@ -61,7 +61,7 @@ Force the installation of the legacy `hermes-compiler` package in your `package.
 
 **iOS**
 
-When installing CocoaPods dependencies, pass the `RCT_HERMES_V1_ENABLED=0` and `RCT_USE_PREBUILT_RNCORE=0` environment variables.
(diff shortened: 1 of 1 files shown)


cipolleschi:
Hi there, thanks for the contributions, but this should not be merged. 
The RNDependencies do not care about new arch/old arch or legacy hermes/hermes v1. 
They are third party libraries built independently from React Native, so users can still benefit from using the precompiled binaries even if they use some different setups than the recommended ones.

You almost never want to build RNDependencies from source.

PR #5161, opened 2026-06-28
fix(copy-page): right-align Copy page button on pages without a TOC

### LINKED ISSUE #5160: "Copy page" button is left-aligned on pages without a table of contents
## Description

On doc pages that have **no right-hand table of contents** (e.g. pages with `hide_table_of_contents: true`), the **"Copy page"** button appears at the top-**left** (above the title) instead of the top-right. On pages that *do* have a TOC, the button correctly sits at the top-right next to the menu.

Compare:

- [Building For TV](https://reactnative.dev/docs/building-for-tv) — `hide_table_of_contents`, button is **mis-aligned to the left**.
- [Out-of-Tree Platforms](https://reactnative.dev/docs/out-of-tree-platforms) — has a TOC, button is correctly on the right.

## Root cause

`website/src/theme/DocItem/Layout/index.tsx` renders the in-article Copy page button inside `.docHeaderContainer` — a flex row with `justify-content: space-between` — next to `<DocVersionBadge />`:

```tsx
<div className={styles.docHeaderContainer}>
  <DocVersionBadge />
  <DocItemCopyPageButton
    className={clsx(
      styles.copyPageArticleAction,
      docTOC.canRender && styles.copyPageArticleActionWithToc
    )}
  />
</div>
```

Two things combine:

1. On the current/latest release, the stock `DocVersionBadge` renders nothing, so `.docHeaderContainer` is left with a **single** flex child. With `justify-content: space-between`, a lone child is pinned to flex-start (the **left**).
2. `.copyPageArticleAction` is referenced via `styles.copyPageArticleAction` but is **not defined** in `website/src/theme/DocItem/Layout/styles.module.css` (only an attribute selector `div[class^="copyPageArticleAction"]` for `margin-bottom` exists). So in the no-TOC case the wrapper has no class to right-align it.

When a TOC is present the article button is `display: none` on desktop (`.copyPageArticleActionWithToc`) and the aside copy button in the right column is shown instead — which is why the bug only appears on pages without a TOC.

## Suggested fix

Define the class so the button is right-aligned regardless of whether the version badge renders:

```css
.copyPageArticleAction {
  margin-left: auto;
}
```

`margin-left: auto` pushes the button to the right edge whether it is the only child (no badge) or sharing the row with the version badge (badge stays left, button right).

## Screenshots
Image

azizbecha:
Fixes #5160.

## Summary

On doc pages without a table of contents (e.g. `hide_table_of_contents: true`), the **"Copy page"** button appeared at the top-**left** (above the title) instead of the top-right.

The in-article button is rendered in `.docHeaderContainer` — a flex row with `justify-content: space-between` — next to `<DocVersionBadge />`. On the current/latest release the stock `DocVersionBadge` renders nothing, so the container is left with a single flex child, which `space-between` pins to flex-start (the left). In addition, `.copyPageArticleAction` was referenced via `styles.copyPageArticleAction` in the component but never defined in `styles.module.css`, so in the no-TOC case the wrapper had no class to right-align it.

## Fix

Define the class with `margin-left: auto`:

```css
.copyPageArticleAction {
  margin-left: auto;
}
```

`margin-left: auto` pushes the button to the right edge whether it is the only child (no version badge) or sharing the row with the badge (badge stays left, button right). When a TOC is present the article button is `display: none` on desktop and the aside button is used instead, so that path is unaffected.

## Test plan

- `prettier --check` passes on the changed file.
- Visual before/after below.

## Screenshots

Before:
Image

After:

Image

### DIFF
--- a/website/src/theme/DocItem/Layout/styles.module.css
+++ b/website/src/theme/DocItem/Layout/styles.module.css
@@ -24,6 +24,13 @@ div[class^="copyPageArticleAction"] {
   margin-bottom: 0;
 }
 
+/* Keep the in-article button right-aligned even when DocVersionBadge renders
+ * nothing (current release), so .docHeaderContainer's space-between doesn't
+ * pin the lone child to the left. */
+.copyPageArticleAction {
+  margin-left: auto;
+}
+
 button[class^="copyPageButton"] {
   border-radius: var(--ifm-global-radius);
   border-color: var(--ifm-color-gray-500);
--- a/website/src/theme/DocItem/Layout/styles.module.css
+++ b/website/src/theme/DocItem/Layout/styles.module.css
@@ -24,9 +24,6 @@ div[class^="copyPageArticleAction"] {
   margin-bottom: 0;
 }
 
-/* Keep the in-article button right-aligned even when DocVersionBadge renders
- * nothing (current release), so .docHeaderContainer's space-between doesn't
- * pin the lone child to the left. */
 .copyPageArticleAction {
   margin-left: auto;
 }


website/src/theme/DocItem/Layout/styles.module.css
```diff
@@ -24,6 +24,13 @@ div[class^="copyPageArticleAction"] {
   margin-bottom: 0;
 }
 
+/* Keep the in-article button right-aligned even when DocVersionBadge renders
+ * nothing (current release), so .docHeaderContainer's space-between doesn't
+ * pin the lone child to the left. */
```
Simek:
```suggestion
```

We don't need the comment which has 2x the content as the fix. 😅 Class name and CSS property are quite self explanatory.

azizbecha: Fixed!

Simek (approved):
Tested the changes on preview deployment on few different pages, looks good! 🙏 

Ref:
* https://deploy-preview-5161--react-native.netlify.app/docs/building-for-tv
* https://deploy-preview-5161--react-native.netlify.app/docs/tutorial

azizbecha:
Thanks @Simek for the review!

PR #5176, opened 2026-07-23
docs: adjust documentation for SceneDelegate adoption

artus9033:
This PR adjusts documentation to reflect changes needed to adopt SceneDelegate.

This PR should be merged after the following PR to React Native is merged: https://github.com/react/react-native/pull/53744

### DIFF
--- a/docs/_integration-with-existing-apps-ios.md
+++ b/docs/_integration-with-existing-apps-ios.md
@@ -248,9 +248,16 @@ React Native initialization is now unbound to any specific part of an iOS app.
 
 React Native can be initialized using a class called `RCTReactNativeFactory`, that takes care of handling the React Native lifecycle for you.
 
-Once the class is initialized, you can either start a React Native view providing a `UIWindow` object, or you can ask for the factory to generate a `UIView` that you can load in any `UIViewController.`
+:::note `RCTAppDelegate` is deprecated
+`RCTAppDelegate` is deprecated and will be removed in a future version of React Native. New apps should bootstrap from an app-owned `SceneDelegate` (see [Bootstrapping with SceneDelegate](#6-bootstrapping-with-scenedelegate)). Existing apps that subclass `RCTAppDelegate` should migrate to `RCTReactNativeFactory` directly.
+:::
 
-In the following example, we will create a ViewController that can load a React Native view as it's `view`.
+Once the class is initialized, you can either:
+
+- bootstrap a full app from your `SceneDelegate` (see [Bootstrapping with SceneDelegate](#6-bootstrapping-with-scenedelegate)), or
+- generate a `UIView` from the factory and load it in any `UIViewController` (the approach below).
+
+In the following example, we will create a ViewController that can load a React Native view as its `view`.
 
 #### Create the ReactViewController
 
@@ -268,7 +275,7 @@ Now open the `ReactViewController.m` file and apply the following changes
 +#import <React/RCTBundleURLProvider.h>
 +#import <RCTReactNativeFactory.h>
 +#import <RCTDefaultReactNativeFactoryDelegate.h>
-+#import <RCTAppDependencyProvider.h>
++#import <ReactAppDependencyProvider/RCTAppDependencyProvider.h>
 
 
 @interface ReactViewController ()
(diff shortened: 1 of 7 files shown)


docs/_integration-with-existing-apps-ios.md
```diff
@@ -461,7 +468,222 @@ Finally, make sure to add the `UIViewControllerBasedStatusBarAppearance` key int
 
 ![Disable UIViewControllerBasedStatusBarAppearance](/docs/assets/disable-UIViewControllerBasedStatusBarAppearance.png)
 
-## 6. Test your integration
+## 6. Bootstrapping with SceneDelegate
+
+If your app uses the UIScene lifecycle (the default for apps created with recent versions of Xcode), React Native bootstrap happens in your app-owned `SceneDelegate`. Your `AppDelegate` stays as the process entry point (`@main`) and handles `UIApplication`-level callbacks such as push notifications.
```
cipolleschi, line 473: Could we add a version callout here? These SceneDelegate APIs require React Native 0.88 or later. In particular, 0.87 does not provide the `connectionOptions:` factory overload or the SceneDelegate methods on `RCTLinkingManager`, so following these instructions on an earlier release would fail to compile or result in incomplete linking support.

docs/_integration-with-existing-apps-ios.md
```diff
@@ -461,7 +468,222 @@ Finally, make sure to add the `UIViewControllerBasedStatusBarAppearance` key int
...
+
+To bootstrap React Native with SceneDelegate:
+
+1. Declare `UIApplicationSceneManifest` in `Info.plist` and point it at your `SceneDelegate` class.
+2. Keep `UIApplicationSupportsMultipleScenes` set to `false`. React Native does not support multi-window / multi-instance apps yet; enabling this key causes React Native to fail during initialization due to an intentional assertion, unless you explicitly define `RN_ALLOW_MULTIPLE_SCENES` on your app target.
```
cipolleschi, line 478: I don’t think the landed implementation contains the assertion or the `RN_ALLOW_MULTIPLE_SCENES` escape hatch described here. Could we state only that multiple scenes are currently unsupported and remove the claims about failing during initialization and defining this macro? Otherwise users may rely on a guard that does not exist.

docs/_integration-with-existing-apps-ios.md
```diff
@@ -461,7 +468,222 @@ Finally, make sure to add the `UIViewControllerBasedStatusBarAppearance` key int
...
+
+</TabItem>
+</Tabs>
+
+:::note
+React Native currently supports a single React instance per process. If `UIApplicationSupportsMultipleScenes` is `true`, React Native fails during initialization by default (when using the New Architecture). Define `RN_ALLOW_MULTIPLE_SCENES=1` in your app target's `GCC_PREPROCESSOR_DEFINITIONS` (or pass `-DRN_ALLOW_MULTIPLE_SCENES=1` via `OTHER_CFLAGS`) only if you understand the risks (e.g. cross-talk of internal RN modules, or third-party libraries) and want to downgrade the failure to a warning.
+:::
+
+:::tip[`reactNativeFactory` field]
+It is important that you expose `reactNativeFactory` on your `SceneDelegate` if you need utilities such as `RCTGetActiveReactNativeFactory()` to resolve the active factory at runtime.
```
cipolleschi, line 683: `RCTGetActiveReactNativeFactory()` was removed before the SceneDelegate work landed and is not available on React Native `main`. Could we remove this reference? We can instead explain that the factory should be retained for the lifetime of the scene.

docs/linking.md
```diff
@@ -54,7 +54,53 @@ If you wish to receive the intent in an existing instance of MainActivity, you m
 <TabItem value="ios">
 
 :::note
-On iOS, you'll need to add the `LinkingIOS` folder into your header search paths as described in step 3 [here](linking-libraries-ios#step-3). If you also want to listen to incoming app links during your app's execution, you'll need to add the following lines to your `*AppDelegate.m`:
+On iOS, you'll need to add the `LinkingIOS` folder into your header search paths as described in step 3 [here](linking-libraries-ios#step-3). If you also want to listen to incoming app links during your app's execution, forward deep links from your `SceneDelegate`. If your app declares `UIApplicationSceneManifest` in `Info.plist`, `RCTLinkingManager` ignores the `AppDelegate` linking methods below - you must forward links from `SceneDelegate` instead.
```
cipolleschi, line 57: Could we also document the cold-start path here? These callbacks cover links received after the scene is connected, but `Linking.getInitialURL()` depends on passing `UIScene.ConnectionOptions` to `startReactNative`. Without that step, apps launched from a custom URL or Universal Link will receive `null`.

cipolleschi (changes requested):
Thanks for documenting this migration—it will be especially useful with the iOS 27 lifecycle requirements. One small housekeeping request: could we update the prerequisite link from the closed umbrella PR #53744 to the landed implementation PR #57700? There are also two trailing-whitespace lines in `docs/releases/release-levels.md` that cause `git diff --check` to fail.

cipolleschi:
also, there are conflicts that needs to be resolved.

PR #5187, opened 2026-07-25
Introduce secondary site nav, relocate nav items

huntie:
> [!Note]
> **Stacked PR**: Depends on #5186.

## Summary

Introduces a secondary navbar above every docs-style section of the website, and renames the top-level **"Development"** menu item to **"Docs"**.

**Motivation**

This, IMO, fixes a missing part of the site navigation story as reactnative.dev has evolved over time.

**Principally**, the navbar heading helps reinforce where you are on the site (motivated by the incoming new **"Releases"** section) — but also improves general context and navigation across the site.

| Before | After |
|--------|--------|
| **Docs > APIs** | |
| image |image |  |
| | Docs > APIs breadcrumb/selector at page level, section-specific version dropdown now in secondary nav, copy button migrated |
| **Contributing** | |
| image | image | |
| **Blog index** | |
| image | image | |
| | (No copy button on index page) |
| **Blog single** | |
| image | image |
| | Improved signalling via "All posts" back button, copy button now available |

(Home, Showcase, etc pages are unchanged.)

This also **significantly improves navigation UX on mobile**:

rn-website-mobile-nav-comparison

**What moved where**

- **Section dropdown:** Guides / Components / APIs / Architecture now live in a dropdown, shown both in the top navbar "Docs" entry and in the secondary nav; the active section follows the current doc's sidebar.
- **Version dropdown:** Relocated from the top navbar into the secondary nav (main docs only) as a pill — brand blue for "Next", otherwise the same outlined style as the section dropdown.
- **Copy page widget:** Moved out of the article body into the secondary nav, on docs, architecture, community and contributing pages and on individual blog posts (not the blog index).

**Other changes**

- Each section shows its name (Docs / Community / Contributing / Blog) in the secondary nav, above an inset bottom keyline.
- On the blog, an "All posts" back link sits beside the heading on every route except the listing.
- Dropped the now-redundant "Community" sidebar category and the "All Blog Posts" blog sidebar heading.

## 🚧 Open questions

- The top level "Docs" nav item retains its dropdown of 4 options ("Guides", "Components", ...) — technically a repeat of what the secondary nav offers. This could be dropped entirely, in favour of "single click to go to docs" (reduce overhead at this point).
    - **Decision**: For now, keep both (do not regress ability to navigate fast from Homepage → Docs subsection. Could revisit this in future.

## Test Plan

**Comparison tour — desktop**

https://github.com/user-attachments/assets/d8870a14-fae5-475e-be5f-6e9134fe3bfc


website/src/components/DocsSecondaryNav/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+] as const;
+
+// Guides, Components and APIs all share the /docs/ URL space, so the active
+// section is read from the current doc's sidebar rather than the pathname.
+const SECTION_BY_SIDEBAR: Record<string, string> = {
+  docs: 'Guides',
+  components: 'Components',
+  api: 'APIs',
+  architecture: 'Architecture',
+};
```
Simek: Looks like should be able to extract the section information directly from config via `useDocusaurusContext`:
* https://docusaurus.io/docs/docusaurus-core#useDocusaurusContext

huntie: Updated ✨

Simek:
New selects does not use the correct border color in dark theme. The appearance can be copied from "CopyPageButton" styling.

Screenshot 2026-07-26 121728

Also the border radiuses does not match, which looks a bit off.

website/src/components/DocsSecondaryNav/styles.module.css
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+.secondaryNav {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 1rem;
+  height: var(--rn-subnav-height);
+  padding: 0 1.5rem;
+  background-color: var(--ifm-background-color);
```
Simek:
```suggestion
  background-color: color-mix(in srgb, var(--ifm-background-color) 93%, transparent 7%);
  backdrop-filter: blur(12px);
```

Backdrop blur for the nav, to match primary nav styling.

Image

website/src/components/DocsSecondaryNav/styles.module.css
```diff
@@ -0,0 +1,181 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+.secondaryNav {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 1rem;
+  height: var(--rn-subnav-height);
+  padding: 0 1.5rem;
```
Simek, line 14: "Components" section can cause nav elements to overlap on mobile.

Screenshot 2026-07-26 123232

We can adjust padding to `1rem` on mobile to better match doc padding, which will slightly help here, but probably won't solve the issue on all devices.

website/sidebarsCommunity.ts
```diff
@@ -1,13 +1,5 @@
...
-  community: [
-    {
-      type: 'category',
-      label: 'Community',
-      collapsed: false,
-      collapsible: false,
-      items: ['overview', 'staying-updated', 'communities', 'support'],
-    },
-  ],
+  community: ['overview', 'staying-updated', 'communities', 'support'],
```
Simek, line 4: We probably also remove the label from the `Contributing` section (to avoid repetition), but this break styling a bit now, causing large links font and lack of styling for non-interactable sub sections.

Image

huntie:
> Also the border radiuses does not match, which looks a bit off.

@Simek Intentional design choice to differentiate these. Versions dropdown also gets a unique colour state for "Next", and logically is a separate component.

Understand the alternative take though! If we need let's launch with this then update later.

website/sidebarsCommunity.ts
huntie: Yeah this is already the state — maybe not pushed to this branch earlier.

image

website/src/components/DocsSecondaryNav/styles.module.css
huntie: 🤯

Even on my Studio Display, I didn't notice this before! Awesome 👍🏻

huntie: It's actually such a slight effect that it's distracting 😂 But yeah aligned, and could be revised later 👍🏻

website/src/components/DocsSecondaryNav/styles.module.css
huntie: I reckon we'll use Flex and have the copy button drop away as the lowest priority item (especially for mobile).

huntie:
@Simek Thanks, all changes addressed.

image

- ~~🚧 Maybe the subsections dropdown is a bit washed out now and could do with a stronger bg/border colour.~~ **Done**.

420px and under:

image

Simek (approved):
Thanks for working on this feature, and all the follow up updates! 🚀

PR #5205, opened 2026-08-09
Document Android 14+ non-linear font scaling on getFontScale and allowFontScaling

sbaiahmed1:
The `PixelRatio.getFontScale()` docs currently tell readers the value "is used to calculate the absolute font size", and the `allowFontScaling` docs imply a single scale factor. On Android 14 and newer that is no longer accurate: the OS applies [non-linear font scaling](https://developer.android.com/about/versions/14/features#non-linear-font-scaling) once the user's font scale is ≥ 1.05 — small fonts grow by roughly the full factor while large fonts grow progressively less, and a 30sp font does not grow at all until the setting exceeds 1.5 (per AOSP's [`FontScaleConverterFactory`](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/android15-release/core/java/android/content/res/FontScaleConverterFactory.java) tables). React Native inherits this via `TypedValue.applyDimension(COMPLEX_UNIT_SP, …)`, so `fontSize * getFontScale()` overestimates rendered sizes of large text.

This mismatch was mistaken for a rendering bug in react/react-native#56961 — I verified there that RN matches the platform tables exactly (measurements in [this comment](https://github.com/react/react-native/issues/56961#issuecomment-5233429897)).

Changes:
- `docs/pixelratio.md`: add a note under `getFontScale()` explaining non-linear scaling and the caveat about multiplying by the returned value.
- `docs/text.md`: add one sentence under `allowFontScaling`.
- Same two changes in `website/versioned_docs/version-0.86/`.


sbaiahmed1:
@Simek can you check please when you have time, Thank you :D

PR #5207, opened 2026-08-13
Correct deintegrate command syntax in documentation

retyui:
small typo


Simek (approved):
Hey @retyui, thanks for the fix! Confirmed that it's the typo locally, LGTM! 👍

PR #5218, opened 2026-08-30
Fix duplicate word in Fabric native commands doc

wook10:
## Summary

Removes an accidental duplicate "the" in `docs/the-new-architecture/fabric-component-native-commands.md`.

> Now you can use the command in the ~~the~~ app.

Text-only fix.


Simek (approved):
Nice catch, thanks for the typo fix! 👍

PR #5238, opened 2026-09-10
feat: Add missing `ReactNativeVersion` API

retyui:
From React Native `0.82.x` the `ReactNativeVersion` can be imported from `'react-native/index.js'` module 

Commit: https://github.com/react/react-native/commit/ec5638abd0e872be62b6ea5d8df9bed6335c2191

### DIFF
--- a/docs/reactnativeversion.md
+++ b/docs/reactnativeversion.md
@@ -1,0 +1,83 @@
+---
+id: reactnativeversion
+title: ReactNativeVersion
+---
+
+```tsx
+import {ReactNativeVersion} from 'react-native';
+```
+
+The `ReactNativeVersion` API exposes the version of the resolved `react-native` package on JS side. Apps and libraries can use it to check compatibility or enable version-specific behavior.
+
+:::note
+
+Also, we have [`Platform.constants.reactNativeVersion`](platform#constants) in our public API already. **However**, this is the per-platform _**native-reported**_ React Native version.
+
+:::
+
+## Example
+
+```tsx
+import {ReactNativeVersion} from 'react-native';
+
+const version = ReactNativeVersion.getVersionString(); // 0.88.1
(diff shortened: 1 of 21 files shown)


website/versioned_docs/version-0.82/reactnativeversion.md
```diff
@@ -0,0 +1,99 @@
...
+title: ReactNativeVersion
+---
+
+```tsx
+import {ReactNativeVersion} from 'react-native';
+```
+
+The `ReactNativeVersion` API exposes the version of the resolved `react-native` package on JS side. Apps and libraries can use it to check compatibility or enable version-specific behavior.
+
+> ℹ️Note: Also we have [`Platform.constants.reactNativeVersion`](platform#constants) in our public API already. **However**, this is the per-platform _**native-reported**_ React Native version.
```
Simek: Let's use Docusuaurs admonition here, instead of quote:

Image

docs/reactnativeversion.md
```diff
@@ -0,0 +1,99 @@
...
+```tsx
+static major: number;
+```
+
+The major version number of the resolved `react-native` package.
+
+| Type   |
+| ------ |
+| number |
+
```
Simek: Repeating type in tables seems redundant in this case. It makes more sense when the value has a more complex type like object, refers to external type, or types/values depends on the platform.

Screenshot 2026-09-10 at 16 25 33

Let's remove the tables for this API from all files.

website/versioned_docs/version-0.82/reactnativeversion.md
retyui: added a note admonition

docs/reactnativeversion.md
retyui: all were removed

Simek (approved):
Thanks for adding a missing doc page and all the updates! 🚀

